PR #2, opened 2019-04-09
Add support for managing multiple clusters

gabrieljackson:
This adds initial support for having multiple clusters running
concurrently. Most of the cluster state data still resides locally
in directories labelled with the cluster ID.

Addresses https://mattermost.atlassian.net/browse/MM-15036

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -7,7 +7,6 @@ import (
 )
 
 func init() {
-	clusterCreateCmd.Flags().String("cluster", "", "The id of the cluster to be created. A random id will be used if left empty.")
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
 	clusterCreateCmd.Flags().String("size", "", "The size constant describing the cluster.")
 	clusterCreateCmd.MarkFlagRequired("size")
@@ -28,13 +27,12 @@ var clusterCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: "Create a cluster.",
 	RunE: func(command *cobra.Command, args []string) error {
-		clusterId, _ := command.Flags().GetString("cluster")
 		provider, _ := command.Flags().GetString("provider")
 		size, _ := command.Flags().GetString("size")
 
 		command.SilenceUsage = true
 
-		return provisioner.CreateCluster(clusterId, provider, size, logger)
+		return provisioner.CreateCluster(provider, size, logger)
 	},
 }
 
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -7,6 +7,7 @@ import (
 )
 
 func init() {
+	clusterCreateCmd.Flags().String("cluster", "", "The id of the cluster to be created. A random id will be used if left empty.")
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
 	clusterCreateCmd.Flags().String("size", "", "The size constant describing the cluster.")
 	clusterCreateCmd.MarkFlagRequired("size")
@@ -27,12 +28,13 @@ var clusterCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: "Create a cluster.",
 	RunE: func(command *cobra.Command, args []string) error {
+		clusterId, _ := command.Flags().GetString("cluster")
 		provider, _ := command.Flags().GetString("provider")
 		size, _ := command.Flags().GetString("size")
 
 		command.SilenceUsage = true
 
-		return provisioner.CreateCluster(provider, size, logger)
+		return provisioner.CreateCluster(clusterId, provider, size, logger)
 	},
 }
 
--- a/internal/provisioner/cluster.go
+++ b/internal/provisioner/cluster.go
@@ -24,7 +24,7 @@ const SizeAlef500 = "SizeAlef500"
 const clusterRootDir = "clusters"
 
 // CreateCluster creates a cluster using kops and terraform.
-func CreateCluster(clusterId, provider, size string, logger log.FieldLogger) error {
+func CreateCluster(provider, size string, logger log.FieldLogger) error {
 	provider = strings.ToLower(provider)
 	if provider != ProviderAWS {
 		return fmt.Errorf("unsupported provider %s", provider)
@@ -34,9 +34,7 @@ func CreateCluster(clusterId, provider, size string, logger log.FieldLogger) err
 		return fmt.Errorf("unsupported size %s", size)
 	}
 
-	if clusterId == "" {
-		clusterId = model.NewId()
-	}
+	clusterId := model.NewId()
 
 	// Temporarily locate the kops output directory to a local folder based on the
 	// cluster name. This won't be necessary once we persist the output to S3 instead.
--- a/internal/provisioner/cluster.go
+++ b/internal/provisioner/cluster.go
@@ -20,7 +20,7 @@ const ProviderAWS = "aws"
 const SizeAlef500 = "SizeAlef500"
 
 // CreateCluster creates a cluster using kops and terraform.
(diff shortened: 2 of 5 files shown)


cmd/cloud/cluster.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 func init() {
+	clusterCreateCmd.Flags().String("cluster", "", "The id of the cluster to be created. A random id will be used if left empty.")
```
lieut-data: Contrary to my own scratch notes on the associated ticket, I'm not entirely convinced we should let this be specified by the user. The only practical use case I can imagine is to support idempotency. But even if we do want to support that, I'd propose not doing it at the primary key level (e.g. leave the application to pick those values), and instead allow some other unique column like a "friendly cluster name" (e.g. `nod-bygone-hyped`). Either way, maybe a future feature?

internal/provisioner/cluster.go
```diff
@@ -30,15 +30,20 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
 		return fmt.Errorf("unsupported size %s", size)
 	}
 
-	outputDir := "tmp"
-	_, err := os.Stat("tmp")
+	if clusterId == "" {
+		clusterId = model.NewId()
+	}
+
+	// Temporarily relocate the kops output directory to a local folder based on the
```
lieut-data: Nit: at this point, we're not "relocating" per se (made more sense in the immediate vicinity of `os.Rename`).

```suggestion
	// Temporarily locate the kops output directory to a local folder based on the
```

internal/provisioner/cluster.go
```diff
@@ -30,15 +30,20 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
+	// cluster name. This won't be necessary once we persist the output to S3 instead.
+	outputDir := fmt.Sprintf("cluster-%s", clusterId)
```
lieut-data: Should we locate this under `clusters/cluster-%s` just to tidy things up for the local developer, and then update `.gitignore` to remove `tmp` and add `clusters`?

internal/provisioner/cluster.go
```diff
@@ -30,15 +30,20 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
+	_, err := os.Stat(outputDir)
 	if err == nil {
-		return errors.New("tmp folder already exists: delete existing cluster first")
+		return fmt.Errorf("encountered cluster ID collision: directory %q already exists", outputDir)
 	} else if err != nil && !os.IsNotExist(err) {
-		return errors.Wrap(err, "failed to stat temporary output directory")
+		return fmt.Errorf("failed to stat cluster directory %q: %s", outputDir, err)
```
lieut-data: Use `errors.Wrapf` here (and below) instead of manually concatenating the `err`.

internal/provisioner/cluster.go
```diff
@@ -30,15 +30,20 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
+		clusterId = model.NewId()
+	}
+
+	// Temporarily relocate the kops output directory to a local folder based on the
+	// cluster name. This won't be necessary once we persist the output to S3 instead.
+	outputDir := fmt.Sprintf("cluster-%s", clusterId)
+	_, err := os.Stat(outputDir)
 	if err == nil {
-		return errors.New("tmp folder already exists: delete existing cluster first")
+		return fmt.Errorf("encountered cluster ID collision: directory %q already exists", outputDir)
```
lieut-data: `fmt.Errorf` is fine, but I've literally just discovered that one can also use `errors.Errorf`. That also gives us the extra stack information. Might be good to update everywhere.

internal/provisioner/cluster.go
```diff
@@ -84,30 +87,49 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
 	dns := fmt.Sprintf("%s-kops.k8s.local", clusterId)
 
-	// Temporarily look for the kops output directory as a local folder named `tmp`. See above.
-	outputDir := "tmp"
+	// Temporarily look for the kops output directory as a local folder named after
+	// the cluster ID. See above.
+	outputDir := fmt.Sprintf("cluster-%s", clusterId)
+
+	// Validate the provided cluster ID before we alter state in any way.
+	logger.Info("verifying cluster ID")
```
lieut-data: nit: This is going to basically fail immediately, so is the log that useful? To me, this falls under the same class of errors as the input checks in `CreateCluster` above, which don't need "starting logs".

internal/provisioner/cluster.go
```diff
@@ -84,30 +87,49 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
+	_, err := os.Stat(outputDir)
+	if err != nil {
+		return fmt.Errorf("failed to find cluster directory %q: %s", outputDir, err)
+	}
 
 	terraform := terraform.New(outputDir, logger)
 	defer terraform.Close()
-	err := terraform.Init()
+	out, err := terraform.Output("cluster_name")
```
lieut-data: Is it true that we'll always be able to `Output` before `Init`? In general, the `Init` is redundant anyway given the local store, but I'm wondering if we should pattern this after the future S3 code anyway, where we'll need to `Init` again after fetching the information locally. Maybe it would break then?

internal/provisioner/cluster.go
```diff
@@ -84,30 +87,49 @@ func CreateCluster(provider, size string, logger log.FieldLogger) error {
...
 	if err != nil {
 		return err
 	}
-	err = terraform.Destroy()
-	if err != nil {
-		return err
+	if out != dns {
```
lieut-data: Can you elaborate on the thinking here? It seems like we're just checking the `clusterId` in a roundabout way, seeing as we build the DNS up from the `clusterId` anyway. Presumably we'd do the same thing when we switch to S3. Is the concern that the folder might be renamed?

internal/tools/kops/cluster.go
```diff
@@ -62,3 +63,32 @@ func (c *Cmd) DeleteCluster(name string) error {
 
 	return nil
 }
+
+// GetCluster invokes kops get cluster, using the context of the created Cmd, and
+// returns the stdout.
+func (c *Cmd) GetCluster(name string) (string, error) {
```
lieut-data: :tada:

internal/tools/terraform/plan.go
```diff
@@ -40,3 +42,28 @@ func (c *Cmd) Destroy() error {
 
 	return nil
 }
+
+// Output invokes terraform output and returns the value.
+func (c *Cmd) Output(variable string) (string, error) {
+	stdout, _, err := c.run(
```
lieut-data: Was this a case where we could get terraform to output JSON and parse that for additional "safety"?

internal/provisioner/cluster.go
gabrieljackson: Yep! Good point.

internal/provisioner/cluster.go
gabrieljackson: Yep! That is the concern. Basically, I want both kops and terraform to tell me that the cluster ID is what we are targeting. This may not be necessary in the future, but with our current local-file setup, I don't want to rely on the directory name being correct.

cmd/cloud/cluster.go
gabrieljackson: I would like to keep this for now. My thinking is it could speed up testing if we are able to create clusters with friendly IDs or even to just speed up scripting by creating clusters with the same ID every time.

Also, perhaps we would want to leave this as a command line option, but not as an API call later. That way it could be used by developers as a tool in a more powerful way.

Thoughts?

gabrieljackson: We chatted about this and agree that we will move forward with no custom IDs for now. If anything, we will add naming metadata in the future.

PR #3, opened 2019-04-11
Expose kops configuration and sizing options

gabrieljackson:
This change provides additional support for defining the size and
configuration of the clusters created by kops.

A few notes on this:
1. The cluster sizing is very arbitrary right now and doesn't support all of the kops command flags for size info.
2. In the longterm, we may want to use the kops template files instead: https://github.com/kubernetes/kops/blob/master/docs/cluster_template.md
3. I moved some of the hard-coded values to the default values of the corresponding command flags. This is simply to save time, but we will probably want to remove them in the longterm.

https://mattermost.atlassian.net/browse/MM-14993

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -9,14 +9,14 @@ import (
 )
 
 func init() {
+	clusterCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
+
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
-	clusterCreateCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
 	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed. Use commas to separate multiple zones.")
 	clusterCreateCmd.MarkFlagRequired("size")
 
 	clusterDeleteCmd.Flags().String("cluster", "", "The id of the cluster to be deleted.")
-	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the existing cluster state is stored.")
 	clusterDeleteCmd.MarkFlagRequired("cluster")
 
 	clusterCmd.AddCommand(clusterCreateCmd)
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -12,11 +12,11 @@ func init() {
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
 	clusterCreateCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
-	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed.")
+	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed. Use commas to separate multiple zones.")
 	clusterCreateCmd.MarkFlagRequired("size")
 
 	clusterDeleteCmd.Flags().String("cluster", "", "The id of the cluster to be deleted.")
-	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the cluster state is stored.")
+	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the existing cluster state is stored.")
 	clusterDeleteCmd.MarkFlagRequired("cluster")
 
 	clusterCmd.AddCommand(clusterCreateCmd)
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"strings"
+
 	"github.com/spf13/cobra"
 
 	"github.com/mattermost/mattermost-cloud/internal/provisioner"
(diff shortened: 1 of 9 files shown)


internal/tools/kops/sizes.go
```diff
@@ -0,0 +1,41 @@
...
+	NodeSize:   "t2.medium",
+	MasterSize: "m3.medium",
+}
+
+// sizeAlef1000 is a cluster sized for 1000 users.
+var sizeAlef1000 = ClusterSize{
+	NodeCount:  "4",
+	NodeSize:   "t2.medium",
+	MasterSize: "m4.large",
+}
```
jwilander: I know these are /super/ ballpark-y but these sizes seem a little small to be running a whole cluster installation (MM, DB, minio, elasticsearch, metrics, nginx). Probably fine for now but my guess is we'll need to bump them up later

internal/tools/kops/sizes_test.go
```diff
@@ -0,0 +1,30 @@
...
+		t.Run(tt.size, func(t *testing.T) {
+			clusterSize, err := GetSize(tt.size)
+			if err != nil && !tt.expectError {
+				t.Errorf("got error when expecting none: %s", err)
+			}
+			if err == nil && tt.expectError {
+				t.Errorf("expecting error, got none")
+			}
+			if clusterSize != tt.clusterSize {
+				t.Errorf("got %+v, want %+v", clusterSize, tt.clusterSize)
```
jwilander: Another suggestion would be to use testify for these (like we do in the server repo) https://godoc.org/github.com/stretchr/testify/assert

internal/tools/kops/sizes.go
gabrieljackson: yeah, this is basically in line with the standard kops sizing if no overrides are given. These will need a bunch of tweaking so I figured it would be good to keep them smaller right now to combat AWS spend.

cmd/cloud/cluster.go
```diff
@@ -1,17 +1,22 @@
...
 func init() {
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
-	clusterCreateCmd.Flags().String("size", "", "The size constant describing the cluster.")
+	clusterCreateCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state")
+	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
+	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed")
 	clusterCreateCmd.MarkFlagRequired("size")
 
 	clusterDeleteCmd.Flags().String("cluster", "", "The id of the cluster to be deleted.")
+	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the cluster state is stored")
```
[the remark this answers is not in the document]
gabrieljackson: So, I actually chose slightly different wording to signify that you needed to provide a specific S3 location where kops could find out about the cluster as opposed to any bucket. I figured it may avoid some confusion in the long run. Thoughts?

cmd/cloud/cluster.go
```diff
@@ -1,17 +1,22 @@
...
 func init() {
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
-	clusterCreateCmd.Flags().String("size", "", "The size constant describing the cluster.")
+	clusterCreateCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
+	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
+	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed.")
 	clusterCreateCmd.MarkFlagRequired("size")
 
 	clusterDeleteCmd.Flags().String("cluster", "", "The id of the cluster to be deleted.")
+	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the cluster state is stored.")
```
lieut-data: Worth putting the `--state-store` flag on the `clusterCmd` to share among all cluster commands?

cmd/cloud/cluster.go
```diff
@@ -28,11 +33,15 @@ var clusterCreateCmd = &cobra.Command{
 	Short: "Create a cluster.",
 	RunE: func(command *cobra.Command, args []string) error {
 		provider, _ := command.Flags().GetString("provider")
+		s3StateStore, _ := command.Flags().GetString("state-store")
 		size, _ := command.Flags().GetString("size")
+		zones, _ := command.Flags().GetString("zones")
+
+		splitZones := strings.Split(zones, ",")
```
lieut-data: I like that this mimics the kops input-style of `--zones us-west-2a,us-west-2b,us-west-2c`. Might we also support `--zone us-west-2a --zone us-west-2b --zone us-west-2c`? If we keep `--zones` with its `comma-splitting` semantics, lets also document that in the corresponding flag help text.

cmd/cloud/cluster.go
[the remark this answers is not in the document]
jwilander: Ahh ok, I see what you were going for. How about adding "existing" then:

```suggestion
	clusterDeleteCmd.Flags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket where the existing cluster state is stored.")
```

internal/provisioner/providers.go
```diff
@@ -0,0 +1,14 @@
+package provisioner
+
+import "fmt"
+
+// providerAWS is the cloud provider AWS.
+const providerAWS = "aws"
+
+func checkProvider(provider string) error {
```
lieut-data: Might we consider a signature of `checkProvider(provider string) (string, error)`, returning the valid provider and allowing us to internalize the `strings.ToLower` in this method, and even support other future multi-mappings (e.g. `"amazon"` or whatever).

internal/tools/kops/sizes_test.go
lieut-data: Agreed -- let's converge on testify here.

gabrieljackson: Sounds good!

cmd/cloud/cluster.go
gabrieljackson: I did think about it, but chose not to based on my comment here: https://github.com/mattermost/mattermost-cloud/pull/3#discussion_r275033601

What are your thoughts?

cmd/cloud/cluster.go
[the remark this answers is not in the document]
lieut-data: I’m not convinced of the need to distinguish the concepts. kops itself [puts](https://github.com/kubernetes/kops/blob/b91db4f360701a7ea90655249ecb7cf68e24545b/cmd/kops/root.go#L130) the flag on the root cmd. It’s perhaps a bit weird if you’re coming to the code without kops context, but I think this is a first class concept for us. Plus, this won’t really scale as we add new cluster commands, e.g. to edit, upgrade.

PR #4, opened 2019-04-16
[MM-14991] Additional kops functionality

gabrieljackson:
This change provides additional kops functionality in the form
of wrapping the kops 'validate', 'upgrade', and 'rolling-update'
commands. This, along with the 'update' command, allows for cluster
alteration of previously-created kops clusters.

A new cloud CLI command 'upgrade' is provided which wraps these
individual commands to update a previously-created cluster to
the latest recommended production ready k8s version.

Also, some additional unit tests were added for both kops and
terraform.

Addresses https://mattermost.atlassian.net/browse/MM-14991

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -51,7 +51,7 @@ var clusterCreateCmd = &cobra.Command{
 
 var clusterUpgradeCmd = &cobra.Command{
 	Use:   "upgrade",
-	Short: "Upgrade a cluster.",
+	Short: "Upgrade k8s on a cluster.",
 	RunE: func(command *cobra.Command, args []string) error {
 		clusterId, _ := command.Flags().GetString("cluster")
 		s3StateStore, _ := command.Flags().GetString("state-store")
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -16,10 +16,14 @@ func init() {
 	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed. Use commas to separate multiple zones.")
 	clusterCreateCmd.MarkFlagRequired("size")
 
+	clusterUpgradeCmd.Flags().String("cluster", "", "The id of the cluster to be upgraded.")
+	clusterUpgradeCmd.MarkFlagRequired("cluster")
+
 	clusterDeleteCmd.Flags().String("cluster", "", "The id of the cluster to be deleted.")
 	clusterDeleteCmd.MarkFlagRequired("cluster")
 
 	clusterCmd.AddCommand(clusterCreateCmd)
+	clusterCmd.AddCommand(clusterUpgradeCmd)
 	clusterCmd.AddCommand(clusterDeleteCmd)
 }
 
@@ -42,6 +46,19 @@ var clusterCreateCmd = &cobra.Command{
 		command.SilenceUsage = true
 
 		return provisioner.CreateCluster(provider, s3StateStore, size, splitZones, logger)
+	},
+}
+
+var clusterUpgradeCmd = &cobra.Command{
+	Use:   "upgrade",
+	Short: "Upgrade a cluster.",
+	RunE: func(command *cobra.Command, args []string) error {
+		clusterId, _ := command.Flags().GetString("cluster")
+		s3StateStore, _ := command.Flags().GetString("state-store")
+
+		command.SilenceUsage = true
+
+		return provisioner.UpgradeCluster(clusterId, s3StateStore, logger)
 	},
 }
 
--- a/internal/provisioner/cluster.go
+++ b/internal/provisioner/cluster.go
@@ -71,14 +71,14 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, logger l
 		return fmt.Errorf("failed to rename kops output directory to %q", outputDir)
 	}
 
-	terraform := terraform.New(outputDir, logger)
-	defer terraform.Close()
-	err = terraform.Init()
+	terraformClient := terraform.New(outputDir, logger)
+	defer terraformClient.Close()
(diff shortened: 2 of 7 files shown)


gabrieljackson:
I wanted to note that I am seeing the following bug in my testing: https://github.com/kubernetes/kops/issues/2990

I have been getting around it by running `kops rolling-update cluster --cloudonly --force` for now, but this is obviously less than ideal. It would be nice to run `kops validate` after cluster creation, but this currently fails due to the bug. Just something to think about for the longterm.

jwilander:
Unrelated to the code: Forgot to mention this earlier but can you throw the ticket number in your PR titles? e.g. `MM-14991 Additional kops functionality`. It's a convention we have

internal/provisioner/cluster.go
```diff
@@ -88,6 +88,76 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, logger l
...
+	out, err := terraform.Output("cluster_name")
+	if err != nil {
+		return err
+	}
+	if out != dns {
+		return fmt.Errorf("terraform cluster_name (%s) does not match dns from provided ID (%s)", out, dns)
+	}
+
+	kops, err := kops.New(s3StateStore, logger)
+	defer kops.Close()
```
jwilander: Should this defer be moved below the error checking? I imagine if creating the wrapper fails the close is unnecessary and it could panic if `kops` is nil

internal/provisioner/cluster.go
```diff
@@ -88,6 +88,76 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, logger l
...
+	// the cluster ID. See above.
+	outputDir := path.Join(clusterRootDir, clusterId)
+
+	// Validate the provided cluster ID before we alter state in any way.
+	_, err := os.Stat(outputDir)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find cluster directory %q", outputDir)
+	}
+
+	terraform := terraform.New(outputDir, logger)
```
jwilander: nit: Not sure if we've already been doing this all over the place but I'm not really a fan of creating variables using the name of an imported package. To me it's very confusing to read e.g. is `terraform.Init()` calling a function in the package `terraform` or a method on the `terraform` variable

cmd/cloud/cluster.go
```diff
@@ -45,6 +49,19 @@ var clusterCreateCmd = &cobra.Command{
 	},
 }
 
+var clusterUpgradeCmd = &cobra.Command{
+	Use:   "upgrade",
+	Short: "Upgrade a cluster.",
```
lieut-data: Wonder if we should explicitly call out that this upgrades Kubernetes -- might not be clear what is actually being "upgraded" as such:
```suggestion
	Short: "Upgrade k8s on a cluster.",
```

internal/provisioner/cluster.go
lieut-data: I started that convention in this repository, but agree with @jwilander that it can be confusing. Perhaps `terraformClient`?

internal/tools/kops/cluster.go
```diff
@@ -51,6 +67,37 @@ func (c *Cmd) UpdateCluster(name string) error {
 	return nil
 }
 
+// UpgradeCluster invokes kops upgrade cluster, using the context of the created Cmd.
+func (c *Cmd) UpgradeCluster(name string) error {
```
lieut-data: Confusingly, the kops documentation [suggests](https://github.com/kubernetes/kops/blob/master/docs/upgrade.md#other-notes):
> In general, we recommend that you upgrade your cluster one minor release at a time (1.7 --> 1.8 --> 1.9). Although jumping minor versions may work if you have not enabled alpha features, you run a greater risk of running into problems due to version deprecation.

Instead of leveraging their automated update, should we instead consider wrapping the manual update steps and accept a target version instead?

internal/tools/kops/run.go
```diff
@@ -40,12 +40,12 @@ func outputLogger(line string, logger log.FieldLogger) {
 	matches := glogRe.FindStringSubmatch(line)
 	if matches == nil {
 		// Assume non-glog output is a warning.
-		logger.Warnf("kops: %s", line)
+		logger.Warnf("[kops] %s", line)
```
lieut-data: 👍

gabrieljackson: Glad you like that change. I find this make the logs a lot easier to understand at a glance.

cmd/cloud/cluster.go
gabrieljackson: For sure, I meant to do that before I pushed this up.

internal/provisioner/cluster.go
gabrieljackson: Yep, agreed!

internal/tools/kops/cluster.go
gabrieljackson: Yeah, so I read the same thing in the kops documentation and had similar thoughts.

I am approaching this uncertainty with the following plan:

We probably want to have the `kops upgrade` command wrapped as it is in my PR, even if we don't use it much in the longterm. I would be curious to use it as we test out a full Mattermost provisioning and upgrade cycle just to see what we experience. In the longterm, once we have some experience, we may want to move to the stepping process you outlined.

i.e. Let's use `kops upgrade` for now so we can make sure that it works as expected, even if we don't want to use it for cluster upgrades in the future. How does that sound?

internal/provisioner/cluster.go
gabrieljackson: Yeah, I think you are right about that. Thanks!

internal/tools/kops/cluster.go
lieut-data: Agreed re: incremental steps. Can you file a ticket to capture the need to step through the upgrades one-by-one?

PR #5, opened 2019-04-17
[MM-15127] Fix golint errors and add to testing pipeline

gabrieljackson:
Addresses https://mattermost.atlassian.net/browse/MM-15127

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,15 @@ jobs:
     steps:
       - checkout
       - *restore_cache
+      - run: go get -u golang.org/x/lint/golint
+      - run: golint -set_exit_status ./...
+      - *save_cache
+  vet:
+    executor:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -16,6 +16,15 @@ aliases:
 
 jobs:
   lint:
+    executor:
```
lieut-data: Is it worth distinguishing `lint` from `vet`? Perhaps make them the same job?

gabrieljackson: I did consider that, but I like having them as separate checks so that you can immediately tell what failed in GitHub. Also, CircleCi seems to support this kind of thing pretty well and this separation is in line with recommendations for building gitlab-ci pipelines (which have worked well for me in the past).

I propose that we try this and change in the future if we find a reason to do so.

PR #6, opened 2019-04-18
add coveralls

cpanato:
Adding coveralls to test the service

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,15 @@ jobs:
       name: default
     steps:
       - checkout
+      - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -37,9 +37,15 @@ jobs:
       name: default
     steps:
       - checkout
+      - run: |
```
lieut-data: `*restore_cache` and `*save_cache` should wrap any `go` commands -- this speeds up the builds by allowing CircleCI to leverage the go module store from previous invocation builds.

cpanato: @lieut-data replace the order

PR #8, opened 2019-04-18
[MM-15217] Add k8s readiness check

gabrieljackson:
This change adds a readiness check to allow for the provisioning
server to properly wait for the cluster to be prepared for running
workloads.

To accomplish this, the ability to run commands silently was also
added so that we can run kops validation checks on a regular
basis without spamming logs.

Addresses https://mattermost.atlassian.net/browse/MM-15217


cmd/cloud/cluster.go
```diff
@@ -14,9 +14,11 @@ func init() {
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
 	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
 	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed. Use commas to separate multiple zones.")
+	clusterCreateCmd.Flags().Bool("cluster-ready", true, "Controls if the command should wait for k8s to become fully ready before exiting")
```
lieut-data: Nit: missing trailing period

internal/provisioner/cluster.go
```diff
@@ -98,13 +104,24 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, logger l
 		return err
 	}
 
+	if waitForReady {
+		logger.WithField("dns", dns).Infof("waiting up to %d seconds for k8s cluster to become ready...", maxReadynessWaitSeconds)
+		err = waitForKubernetesReadyness(dns, kops, maxReadynessWaitSeconds)
+		if err != nil {
+			// Run non-silent validate one more time to log final cluster state
```
lieut-data: Should we internalize this within the wait helper?

internal/provisioner/cluster.go
```diff
@@ -234,6 +258,38 @@ func DeleteCluster(clusterID, s3StateStore string, logger log.FieldLogger) error
 	return nil
 }
 
+// waitForKubernetesReadyness will poll a given kubernetes cluster at a regular
```
lieut-data: Should we push this into the kops interface?

cmd/cloud/cluster.go
lieut-data: Also, would it be simpler if we just called this `--wait`? In the future, we'll likely be invoking the REST API (like mmctl), and waiting will have multiple meanings.

go.mod
```diff
@@ -8,13 +8,14 @@ require (
...
 	github.com/mattermost/mattermost-server v5.9.0+incompatible
+	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/nicksnyder/go-i18n v1.10.0 // indirect
 	github.com/pborman/uuid v1.2.0 // indirect
 	github.com/pelletier/go-toml v1.3.0 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/sirupsen/logrus v1.4.1
 	github.com/spf13/cobra v0.0.3
-	github.com/spf13/pflag v1.0.3 // indirect
+	github.com/spf13/viper v1.3.2 // indirect
```
lieut-data: Any idea what triggered a change here?

internal/provisioner/cluster.go
```diff
@@ -98,13 +104,24 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, logger l
 		return err
 	}
 
+	if waitForReady {
+		logger.WithField("dns", dns).Infof("waiting up to %d seconds for k8s cluster to become ready...", maxReadynessWaitSeconds)
```
lieut-data: Do we need `WithField("dns", dns)` here? If we do, let's just modify `logger` directly on line 63. But I'd argue we don't -- the cluster id embedded is already sufficient. The DNS added on line 63 was more of a "hey, this is the DNS entry!".

internal/provisioner/cluster.go
```diff
@@ -234,6 +258,38 @@ func DeleteCluster(clusterID, s3StateStore string, logger log.FieldLogger) error
 	return nil
 }
 
+// waitForKubernetesReadyness will poll a given kubernetes cluster at a regular
+// interval for it to become ready. If the cluster fails to become ready before
+// the provided timeout then an error will be returned.
+func waitForKubernetesReadyness(dns string, kops *kops.Cmd, timeout time.Duration) error {
```
lieut-data: nit: `s/readyness/readiness/`

internal/provisioner/cluster.go
```diff
@@ -234,6 +258,38 @@ func DeleteCluster(clusterID, s3StateStore string, logger log.FieldLogger) error
...
+	ready := make(chan bool, 1)
```
lieut-data: Since there's no actual message going across the wire here, you can just make this unbuffered and replace the send on line 277 with a `close(ready)` to signal. It makes the intention of the channel as a signal mechanism more clear, vs. the traditional messaging passing mechanism.

(I note a different simplification below, though!)

internal/provisioner/cluster.go
```diff
@@ -234,6 +258,38 @@ func DeleteCluster(clusterID, s3StateStore string, logger log.FieldLogger) error
...
+	go func() {
+		for {
+			err := kops.ValidateCluster(dns, true)
+			if err == nil {
+				break
+			}
+			time.Sleep(5 * time.Second)
+		}
+
+		ready <- true
```
lieut-data: suggest `defer close(ready)` at the top of the function, and changing `break` to `return`.

internal/provisioner/cluster.go
```diff
@@ -234,6 +258,38 @@ func DeleteCluster(clusterID, s3StateStore string, logger log.FieldLogger) error
...
 
+// waitForKubernetesReadyness will poll a given kubernetes cluster at a regular
+// interval for it to become ready. If the cluster fails to become ready before
+// the provided timeout then an error will be returned.
+func waitForKubernetesReadyness(dns string, kops *kops.Cmd, timeout time.Duration) error {
+	ready := make(chan bool, 1)
+	defer close(ready)
+
+	go func() {
+		for {
```
lieut-data: This goroutine will never terminate if the validation never succeeds -- in practice, the CLI will exit so we're probably fine in this implementation, but this could leak resources in a long-running server.

I'm guessing we aren't actually worried about `kops validate` hanging? (If so, the following is invalid.)

Might we just eliminate the goroutine altogether and collapse this function to something like:
```go
func waitForKubernetesReadiness(dns string, kops *kops.Cmd, timeout time.Duration) error {
    timer := time.NewTimer(timeout * time.Second)
    defer timer.Stop()

    for {
        select {
        case <-timer.C:
            return errors.New("timed out waiting for k8s cluster to become ready")
        default:
            err := kops.ValidateCluster(dns, true)
            if err == nil {
                return nil
            }
            time.Sleep(5 * time.Second)
        }
    }

    return nil
}
```

internal/provisioner/cluster.go
```diff
@@ -16,8 +17,13 @@ import (
 // clusterRootDir is the local directory that contains cluster configuration.
 const clusterRootDir = "clusters"
 
+// maxReadynessWaitSeconds controls how long we will wait for a kubernetes
+// cluster to become ready after it is created before we timeout and return
+// an error.
+const maxReadynessWaitSeconds = 600
```
lieut-data: Is there a circumstance under which the client will want to control this parameter? Should we make `--wait` into an integer value with a default of `600`?

go.mod
gabrieljackson: Unfortunately not. I changed somewhere between starting this change and finishing it.

internal/provisioner/cluster.go
gabrieljackson: Yeah, that's probably overly verbose. I will change it.

internal/provisioner/cluster.go
gabrieljackson: That's a great idea!

internal/provisioner/cluster.go
gabrieljackson: I was sort of worried about `kops validate` hanging, but that's probably not worth the extra complexity here. This solution seems like it may be the better first implementation and we can tackle strange kops-hanging issues should we run into them.

internal/provisioner/cluster.go
gabrieljackson: I was thinking it made more sense to not include it in the wait helper. My thinking was that it is more flexible this way in that maybe you wanted to use the helper to wait multiple times in smaller increments with some action in between.

By keeping the option to print a final validation status outside of the helper we can keep the logs cleaner.

Thoughts?

internal/provisioner/cluster.go
gabrieljackson: Yeah, we can do that for now. I had some thoughts as to other checks we may want to do that aren't kops-specific, but we can move this helper back out should we want to include some of that.

PR #9, opened 2019-04-22
MM-14995: initial schema and tooling mvp

lieut-data:
Leverage https://github.com/jmoiron/sqlx and https://github.com/Masterminds/squirrel behind some simple wrappers to implement a store for managing the known clusters.

Leverage this store in the provisioner, introducing new commands to get and list clusters.

Rebrand "dns" as the "kops name" for clarity.

Note that this PR supports both SQLite and Postgres, the former being useful for quick local testing, with the latter meant for a deployed scenario.


internal/store/store.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, errors.Errorf("unsupported dsn scheme %s", url.Scheme)
+	}
+
+	return &SQLStore{
+		db,
+		logger,
+	}, nil
+}
+
+// queryer is an interface describing a resource that can query.
```
lieut-data: The following interfaces and functions are the "thin layer" I spoke of in the PR's description. At first, I intended to fully adopt `sqlx.*` types throughout the store, but I ran into a few limitations:
* `Exec`, by default, doesn't rebind the `?` placeholder for the database in question. There's an easy `Rebind` that does this, but it's annoying to do at the caller.
* There's a handy `NamedExec` that both rebinds and accepts structs and maps, but there's no top-level `NamedGet` equivalent. You can prepare a `NamedStmt` to get the `NamedGet` equivalent, but once again this is annoying at the caller.
* I wanted to leverage squirrel for those times when the SQL needs to be dynamic, but it's slightly annoying to have to extract the sql and args only then to pass them back into sqlx.

All of this spoke to the need for a thin layer to simplify usage in the store. I realized quickly that I could constrain `sqlx` entirely to this file, and so added a few more helper functions accordingly.

lieut-data:
Apologies for any in-flight reviews, but I've pushed two new commits:
* merge master, addressing merge conflicts
* a rework of my misguided "KopsCluster" struct. I've commented in that commit, but hopefully the code is cleaner and more correct as a result.

internal/store/migrate.go
```diff
@@ -0,0 +1,69 @@
...
+		if !currentVersion.EQ(migration.fromVersion) {
+			break
+		}
+
+		sqlStore.logger.Infof("Migrating schema from %s to %s", currentVersion, migration.toVersion)
+		tx, err := sqlStore.db.Beginx()
+		if err != nil {
+			return errors.Wrapf(err, "failed to begin applying target version %s", migration.toVersion)
+		}
+		defer tx.Rollback()
```
jwilander: Does rollback not do anything if the commits are successful? Also what's the convention around defer in a loop? I feel like it would be better to wrap the contents of the loop in a function so that the defers can't pile up

internal/store/cluster.go
```diff
@@ -0,0 +1,111 @@
...
+	json.Unmarshal(data, &clone)
+
+	return &clone
+}
+
+// GetCluster fetches the given cluster by id.
+func (sqlStore *SQLStore) GetCluster(id string) (*Cluster, error) {
+	var cluster Cluster
+
+	err := sqlStore.get(sqlStore.db, &cluster, `SELECT * FROM Cluster WHERE ID = ?`, id)
```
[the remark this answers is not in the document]
gabrieljackson: Makes sense to me and I am on board with that. 0/5 as well though.

internal/provisioner/cluster.go
```diff
@@ -42,26 +68,24 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, wait int
...
-
-	logger.WithField("dns", dns).Info("creating cluster")
+	logger.WithField("name", kopsMetadata.Name).Info("creating cluster")
 
 	kops, err := kops.New(s3StateStore, logger)
 	if err != nil {
 		return err
 	}
 	defer kops.Close()
-	err = kops.CreateCluster(dns, provider, kopsClusterSize, zones)
```
gabrieljackson: Should we clean any new cluster objects in the database if we error before this command? I leaked a cluster because I forgot to have my AWS credentials exported on my first creation attempt.

internal/provisioner/cluster.go
```diff
@@ -100,32 +124,40 @@ func CreateCluster(provider, s3StateStore, size string, zones []string, wait int
...
-	logger.WithField("dns", dns).Info("successfully created cluster")
+	logger.WithField("name", kopsMetadata.Name).Info("successfully created cluster")
 
 	return nil
 }
 
 // UpgradeCluster upgrades a cluster to the latest recommended production ready k8s version.
-func UpgradeCluster(clusterID, s3StateStore string, wait int, logger log.FieldLogger) error {
-	logger = logger.WithField("cluster", clusterID)
+func UpgradeCluster(cs clusterStore, clusterID, s3StateStore string, wait int, logger log.FieldLogger) error {
```
gabrieljackson: Just checking: eventually we will want to store some extra data like k8s version in the DB, right?

internal/store/cluster.go
```diff
@@ -0,0 +1,164 @@
...
+	CreateAt            time.Time
+	DeleteAt            *time.Time
+	LockAcquiredBy      *string
+	LockAcquiredAt      *time.Time
+}
+
+// Clone returns a deep copy the cluster.
+func (c *Cluster) Clone() *Cluster {
+	var clone Cluster
+	data, _ := json.Marshal(c)
```
gabrieljackson: Any risks not checking/handling the error?

internal/store/cluster.go
```diff
@@ -0,0 +1,111 @@
...
+// Cluster represents a Kubernetes cluster as tracked by the store.
+type Cluster struct {
+	ID                  string
+	Provider            string
+	Provisioner         string
+	ProviderMetadata    []byte
+	ProvisionerMetadata []byte
+	AllowInstallations  bool
+	CreateAt            time.Time
+	DeleteAt            *time.Time
```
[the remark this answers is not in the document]
lieut-data: Yeah, this is probably unnecessary. I had really wanted to use `time.Time` instead of just an integer, but I've already had to jump through hoops on this one. Will converge back to just an integer.

internal/provisioner/cluster.go
lieut-data: In all likelihood -- yes. Unclear if we want this in the top-level Clusters schema, or if it would be part of the provisioner metadata.

internal/store/cluster.go
lieut-data: My gut says no -- any marshalling issue is likely to apply to all clusters and not a particular cluster. 0/5 on continuing to omit, though.

internal/provisioner/cluster.go
lieut-data: I ran into this myself. My gut says not to do anything right now -- since we're just as likely to fail partway through any other command as well, and we'll want some way to track the status of the cluster and "--resume" creating it. I think this is captured by future epics.

internal/store/migrate.go
lieut-data: Yeah, `tx.Rollback` is a no-op if the transaction has already committed (or rolled back). This makes for a nice clean-up paradigm.

But you're 100% right on the defer in a loop -- that's just horribly wrong on my part. Will fix.

internal/store/cluster.go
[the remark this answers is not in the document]
lieut-data: Heh, I had contemplated that, but thought the idea of adopting it universally would be "oppressive". No objections -- though I'm opting not to force it in awkward cases, like `EXISTS` with subqueries.

PR #10, opened 2019-04-24
MM-14996: REST API proof-of-concept

lieut-data:
Introduce the REST API, rewriting the existing cli to leverage same (and require a `--server` parameter). Implement the `server` command to start a provisioning server to expose the API. Introduce a Go client for the API to facilitate unit-testing (currently remaining `internal`).

Note that there are a few new TODOs, largely around the pending changes to make this API asynchronous. Rather than jumping through hoops with the current model, I've just noted where changes need to be made.

Fixes: https://mattermost.atlassian.net/browse/MM-14996


internal/api/cluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+// handleGetClusters responds to GET /api/clusters.
+func handleGetClusters(c *Context, w http.ResponseWriter, r *http.Request) {
+	pageStr := r.URL.Query().Get("page")
+	page, _ := strconv.Atoi(pageStr)
+
+	perPageStr := r.URL.Query().Get("per_page")
+	perPage := 100
+	if perPageStr != "" {
+		perPage, _ = strconv.Atoi(perPageStr)
```
gabrieljackson: Do we want to log the error? It may help us uncover strange behavior in the short term at least. 1/5

internal/api/cluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+	outputJSON(c, w, clusters)
+}
+
+// handleCreateCluster responds to POST /api/clusters
+func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
+	createClusterRequest, err := newCreateClusterRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithField("error", err).Error("failed to decode request")
+		w.WriteHeader(http.StatusInternalServerError)
```
gabrieljackson: Is there another code we want to send when we have a payload parsing error vs. and error with the actual task?

internal/api/cluster.go
lieut-data: Indeed -- and I'm going to error out properly with `http.StatusBadRequest` here too. Thanks!

internal/api/cluster.go
lieut-data: Yes, good call! Switching to `http.StatusBadRequest`.

internal/api/api.go
```diff
@@ -0,0 +1,17 @@
...
+// Register registers the API endpoints on the given router.
+func Register(rootRouter *mux.Router, context *Context) {
+	addContext := func(handler contextHandlerFunc) *contextHandler {
+		return newContextHandler(context, handler)
+	}
+
+	apiRouter := rootRouter.PathPrefix("/api").Subrouter()
+	apiRouter.Handle("/clusters", addContext(handleGetClusters)).Methods("GET")
+	apiRouter.Handle("/clusters", addContext(handleCreateCluster)).Methods("POST")
+	apiRouter.Handle("/cluster/{cluster}", addContext(handleGetCluster)).Methods("GET")
```
jwilander: Three questions here that I'm 0/5 on:

1. Why register the routes here instead of inside of the individual api files (e.g. cluster.go)?
2. Would it make sense to use subrouters here to make it a bit cleaner and easier to extend? Something like:
```go
clustersRouter := apiRouter.PathPrefix("/clusters").Subrouter()
clustersRouter.Handle("", addContext(handleGetClusters)).Methods("GET")
// other "/clusters" routes

clusterRouter := apiRouter.PathPrefix("/cluster/{cluster}").Subrouter()
clusterRouter.Handle("", addContext(handleGetCluster)).Methods("GET")
clusterRouter.Handle("/kubernetes/{version}", addContext(handleUpgradeCluster)).Methods("PUT")
// other "/cluster/{cluster}" routes
```
3. Did you purposefully not use regexes for the path parameters? (e.g. `/cluster/{cluster}:[A-Za-z0-9]+`)

internal/api/cluster.go
```diff
@@ -0,0 +1,150 @@
+package api
```
jwilander: I imagine this file doesn't register it's own handlers to avoid having the router globally available in the package?

internal/api/cluster.go
```diff
@@ -54,7 +64,7 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	createClusterRequest, err := newCreateClusterRequestFromReader(r.Body)
 	if err != nil {
 		c.Logger.WithField("error", err).Error("failed to decode request")
-		w.WriteHeader(http.StatusInternalServerError)
+		w.WriteHeader(http.StatusBadRequest)
```
gabrieljackson: I like it!

internal/api/cluster.go
lieut-data: At the time, I had left everything in api.go, thinking it would actually be easier to introspect the endpoints if they were in one place instead of scattered across multiple files. 0/5, but maybe we defer until the "next" API endpoint and see how that looks?

internal/api/api.go
lieut-data: 1. No objections!
2. Not purposefully :)

Fixed on both counts.

internal/api/cluster.go
jwilander: Good with me 👍

PR #11, opened 2019-04-30
MM-15206 async cluster creation

lieut-data:
Move to an asynchronous model for managing clusters, setting the desired state and running a background job to effect same. 

The original list of proposed states for a cluster proved to be too limiting, so this PR models various `*-requested`, `*-failed` and a `stable` state. As we iterate on the provisioner, it is likely we'll need to break up these states further to allow us to resume part of the requested operation. At the moment, re-trying a creation mid-kops+terraform execution doesn't work because kops+terraform start from the beginning and get confused. But, creating/deleting multiple clusters concurrently is now fully supported.

Fixes: https://mattermost.atlassian.net/browse/MM-15206
Fixes: https://mattermost.atlassian.net/browse/MM-15207


cmd/cloud/cluster.go
```diff
@@ -18,7 +18,6 @@ func init() {
 	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
 	clusterCreateCmd.Flags().String("zones", "us-east-1a", "The zones where the cluster will be deployed. Use commas to separate multiple zones.")
 	clusterCreateCmd.Flags().Int("wait", 600, "The amount of seconds to wait for k8s to become fully ready before exiting. Set to 0 to exit immediately.")
-	clusterCreateCmd.MarkFlagRequired("size")
```
lieut-data: @gabrieljackson, I know we talked about this -- let me know if you disagree with this thinking, but I opted to make `--size` not required since we already default other bits in a non-trivial way, and it really simplifies the CLI to get a cluster going from scratch.

cmd/cloud/cluster.go
```diff
@@ -75,8 +74,7 @@ var clusterCreateCmd = &cobra.Command{
 			return errors.Wrap(err, "failed to create cluster")
 		}
 
-		// TODO --wait for the cluster to be ready. For now, this is done synchronously
-		// in CreateCluster with a hard-coded interval.
+		// TODO --wait for the cluster to be ready.
```
lieut-data: Let's discuss if we actually want to do this in the CLI.

cmd/cloud/logger.go
```diff
@@ -1,13 +1,16 @@
...
+	"github.com/mattermost/mattermost-cloud/internal/model"
 	log "github.com/sirupsen/logrus"
 )
 
+var instanceID string
 var logger log.FieldLogger
 
 func init() {
-	logger = log.New()
+	instanceID = model.NewID()
```
lieut-data: The thinking with the `instanceID` is to be able to uniquely identify a particular instance among others. We use this when locking clusters so that we can correlate "who locked" the cluster with logs from that instance. It should also help when we aggregate the logs from all the servers to pull out the instance-specific ones.

cmd/cloud/server.go
```diff
@@ -97,8 +124,6 @@ var serverCmd = &cobra.Command{
 		defer cancel()
 		srv.Shutdown(ctx)
 
-		logger.Info("Shut down")
```
lieut-data: Removed, since redundant.

internal/api/client.go
```diff
@@ -78,14 +78,31 @@ func (c *Client) CreateCluster(request *CreateClusterRequest) (*model.Cluster, e
 	defer closeBody(resp)
 
 	switch resp.StatusCode {
-	case http.StatusOK:
+	case http.StatusAccepted:
```
lieut-data: All modification endpoints now return accepted instead of ok, since the actual changes are of course still pending.

internal/api/cluster.go
```diff
@@ -18,26 +19,66 @@ func outputJSON(c *Context, w io.Writer, data interface{}) {
 	encoder := json.NewEncoder(w)
 	err := encoder.Encode(data)
 	if err != nil {
-		c.Logger.WithField("error", err).Error("failed to encode result")
+		c.Logger.WithError(err).Error("failed to encode result")
```
lieut-data: `WithError` standardizes on how the `error` is included in the logs.

internal/api/cluster.go
```diff
@@ -18,26 +19,66 @@ func outputJSON(c *Context, w io.Writer, data interface{}) {
...
+// handleGetClusters responds to GET /api/clusters, returning the specified page of clusters.
 func handleGetClusters(c *Context, w http.ResponseWriter, r *http.Request) {
+	var err error
 	pageStr := r.URL.Query().Get("page")
-	page, err := strconv.Atoi(pageStr)
-	if err != nil {
-		c.Logger.WithField("error", err).Error("failed to parse page")
-		w.WriteHeader(http.StatusBadRequest)
-		return
+	page := 0
```
lieut-data: Fix an issue when invoking the API via HTTP (by hand) and forgetting to pass the `page` parameter.

internal/api/cluster.go
```diff
@@ -59,33 +100,93 @@ func handleGetClusters(c *Context, w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	cluster, err := c.Provisioner.CreateCluster(createClusterRequest)
+	cluster := model.Cluster{
+		Provider:    createClusterRequest.Provider,
+		Provisioner: "kops",
+		Size:        createClusterRequest.Size,
+		State:       model.ClusterStateCreationRequested,
+	}
+	cluster.SetProviderMetadata(model.AWSMetadata{
```
lieut-data: I'm not fully sold on this `*Metadata` modelling business. For now, I use this to pass `Zones` downstream, but it's a bit all over the place. Thinking we might want to converge on just a single `Metadata` field at least.

internal/api/cluster.go
```diff
@@ -97,54 +198,92 @@ func handleGetCluster(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	cluster, status, unlockOnce := lockCluster(c, clusterID)
+	if status != 0 {
+		w.WriteHeader(status)
 		return
 	}
-	if cluster == nil {
-		w.WriteHeader(http.StatusNotFound)
+	defer unlockOnce()
+
+	switch cluster.State {
```
lieut-data: We should support additional states, such as deleting from a failed-creation state (or if not locked, from the creation requested state), but our kops+terraform provisioner chokes on that right now.

internal/api/request.go
```diff
@@ -24,15 +25,23 @@ func newCreateClusterRequestFromReader(reader io.Reader) (*CreateClusterRequest,
...
 		createClusterRequest.Zones = []string{"us-east-1a"}
 	}
 
+	if createClusterRequest.Provider != model.ProviderAWS {
+		return nil, errors.Errorf("unsupported provider %s", createClusterRequest.Provider)
+	}
+	if !model.IsSupportedSize(createClusterRequest.Size) {
+		return nil, errors.Errorf("unsupported size %s", createClusterRequest.Size)
+	}
+	// TODO: check zones?
```
lieut-data: Part of the above discussion about figuring out how to model things.

internal/provisioner/kops_provisioner.go
```diff
@@ -8,146 +8,105 @@ import (
...
-		}
-	}
-
-	if terraformFactory == nil {
-		terraformFactory = func(outputDir string, logger log.FieldLogger) TerraformCmd {
-			return terraform.New(outputDir, logger)
-		}
-	}
-
+func NewKopsProvisioner(clusterRootDir string, s3StateStore string, logger log.FieldLogger) *KopsProvisioner {
```
lieut-data: I ripped out all the factory stuff that I had just added, since the new model doesn't require me to fake out the provisioner anymore, and we're not testing this code path right now anyway.

The provisioner now doesn't interact with the cluster store at all -- ideally, we can keep it focussed on just "do this" vs. "query database and decide to do this". I envision breaking this creation up into more states from which we can resume successfully after an interruption.

internal/provisioner/kops_provisioner.go
```diff
@@ -8,146 +8,105 @@ import (
...
 	}
 }
 
 // CreateCluster creates a cluster using kops and terraform.
-func (provisioner *KopsProvisioner) CreateCluster(request *api.CreateClusterRequest) (*model.Cluster, error) {
-	provider, err := checkProvider(request.Provider)
-	if err != nil {
-		return nil, err
-	}
+func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error {
```
lieut-data: No longer returning `*model.Cluster` -- just modify it directly as necessary.

internal/provisioner/supervisor.go
```diff
@@ -0,0 +1,177 @@
...
+	return &Supervisor{
+		clusterStore: clusterStore,
+		provisioner:  provisioner,
+		workers:      workers,
+		logger:       logger,
+	}
+}
+
+// Do looks for work to be done on any pending clusters and attempts to schedule the required work.
+func (s *Supervisor) Do() error {
```
lieut-data: I'm imagining writing an equivalent supervisor for managing installations.

internal/store/cluster.go
```diff
@@ -23,6 +23,99 @@ func (sqlStore *SQLStore) GetCluster(id string) (*model.Cluster, error) {
...
+	} else if err != nil {
+		return nil, errors.Wrap(err, "failed to get cluster pending work")
+	}
+
+	return &cluster, nil
+}
+
+// LockCluster marks the cluster as locked for exclusive use by the caller.
+func (sqlStore *SQLStore) LockCluster(clusterID string) (bool, error) {
+	result, err := sqlStore.execBuilder(sqlStore.db, sq.
```
lieut-data: Relying on database semantics here to guarantee two updates can't both write to the same row like this.

internal/store/migrate.go
```diff
@@ -31,7 +31,7 @@ func (sqlStore *SQLStore) Migrate() error {
 	applied := 0
 	for _, migration := range migrations {
 		if !currentVersion.EQ(migration.fromVersion) {
-			break
+			continue
```
lieut-data: Fix an issue with migrations not actually doing more than the first one ;P

internal/store/migrations.go
```diff
@@ -45,6 +45,90 @@ var migrations = []migration{
 			return err
 		}
 
+		return nil
+	}},
+	{semver.MustParse("0.1.0"), semver.MustParse("0.2.0"), func(e execer) error {
```
lieut-data: Strictly speaking, I could force you all to just re-create your databases while we're <1.0.0, but I figure I'd exercise the migration system for now.

lieut-data:
@gabrieljackson & @bradjcoughlin, heads up that I've merged `origin/master` to pick up @gabrieljackson's changes to install the Mattermost operator.

gabrieljackson:
Awesome! I will review this as soon as possible.

cmd/cloud/server.go
```diff
@@ -50,20 +53,43 @@ var serverCmd = &cobra.Command{
...
+			s3StateStore,
+			logger,
+		)
+
+		// Setup the supervisor to effect any requested changes. It is wrapped in a
+		// scheduler to trigger it periodically in addition to being poked by the API
+		// layer.
+		poll, _ := command.Flags().GetInt("poll")
+		if poll == 0 {
+			logger.Info("Switching scheduler to manual mode only")
```
gabrieljackson: Hmm, this behavior is bit strange to me. With `poll` set to 0, the provisioning server will act on requests only if it has a non-busy job at the time the request is made. If the provisioning server is completely busy when another API request is made then the cluster objects will have their state updated, but no actions will be taken on them. Obviously, if another provisioning server is watching the cluster store objects that isn't set to `poll=0` then it will act on them, but I am wondering if we want any provisioning servers to behave like this when `poll=0`?

I could see this being useful if manual mode returned an error if it was completely busy and didn't update an cluster states at all so that the caller knew that another attempt was needed later.

Thoughts?

cmd/cloud/logger.go
lieut-data: Hmm, an excellent point -- we should indeed be locking per-worker instead of just per-instance.

cmd/cloud/logger.go
```diff
@@ -1,13 +1,16 @@
...
+	logger = log.New().WithField("instance_id", instanceID)
```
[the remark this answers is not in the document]
lieut-data: In practice, the CLI doesn't output many logs, but I nevertheless agree with your reasoning. It probably makes the most sense to modify the logger only in the `server` command.

cmd/cloud/server.go
lieut-data: You're right: the dissonance here is wrong. My primary use case was to force the server into not running "background" jobs, which can be really handy when pointing at a shared database for debugging but not wanting to start working on outstanding jobs.

However, the behaviour I have implemented that supports a "manual" poke makes that weird. Maybe if it had poked only the item in question, but I think I'll just rip the manual mode out and disable the scheduler altogether if `poll` is `0`.

cmd/cloud/cluster.go
lieut-data: Inclined to rip this one out altogether: any objections?

lieut-data:
@gabrieljackson: comments from first pass addressed. I'll look into the `--verbose` (or `--quiet`) flag as discussed separately.

internal/model/provider.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	// ProviderAWS is the cloud provider AWS.
+	ProviderAWS = "aws"
+	// ProviderAzure is the cloud provider Azure.
```
gabrieljackson: Just stubbing this out for later? :)

internal/provisioner/kops_provisioner.go
gabrieljackson: Awesome!

internal/store/migrate.go
gabrieljackson: Haha, good catch!

internal/store/migrations.go
gabrieljackson: Yeah, I am totally fine with this.

internal/api/cluster.go
```diff
@@ -97,54 +198,92 @@ func handleGetCluster(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	err = c.Provisioner.DeleteCluster(clusterID)
-	if err != nil {
-		c.Logger.WithField("error", err).Error("failed to delete cluster")
-		w.WriteHeader(http.StatusInternalServerError)
-		return
+	if cluster.State != model.ClusterStateDeletionRequested {
```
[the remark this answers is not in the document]
lieut-data: I took at look at this, but I think I can only save about ~two lines in the caller if we preserve ownership of setting the http status in the caller. As we flesh out the API, it might be good to revisit this to add some abstraction, but I'm inclined to keep the verbosity for now.

internal/model/provider.go
lieut-data: Meh, yeah I'll just remove this, especially since it's not meaningfully complete.

lieut-data:
@gabrieljackson, I've pushed up a few more changes based on our offline discussion. Part of these come from my refactoring to support the pending async installation API, and the rest comes from improved creation/deletion semantics that make it possible to (partially) recover a failed creations, and always delete a cluster from any state (if not locked).

internal/api/cluster.go
```diff
@@ -1,43 +1,86 @@
...
+
+	unlockOnce := sync.Once{}
+
+	return cluster, 0, func() {
+		unlockOnce.Do(func() {
+			unlocked, err := c.Store.UnlockCluster(cluster.ID, c.RequestID, false)
+			if err != nil {
+				c.Logger.WithError(err).Errorf("failed to unlock cluster")
+			} else if unlocked != true {
+				c.Logger.Warn("failed to release lock for cluster")
```
crspeller: This is an error that needs to be dealt with manually (cluster is stuck locked) Error seems appropriate.

internal/model/cluster.go
```diff
@@ -15,6 +34,8 @@ type Cluster struct {
 	ProviderMetadata    []byte `json:",omitempty"`
```
crspeller: As discussed offline this seems like overkill. (not blocking on this PR)

internal/model/aws_metadata.go
```diff
@@ -0,0 +1,21 @@
...
+
+// NewAWSMetadata creates an instance of AWSMetadata given the raw provider metadata.
+func NewAWSMetadata(providerMetadata []byte) *AWSMetadata {
+	awsMetadata := AWSMetadata{}
+
+	if providerMetadata == nil {
+		return &awsMetadata
+	}
+
+	_ = json.Unmarshal(providerMetadata, &awsMetadata)
```
crspeller: Do we really not care if this fails to unmarshal?

internal/provisioner/kops_provisioner.go
crspeller: Why?

internal/store/store.go
```diff
@@ -32,11 +33,14 @@ func New(dsn string, logger logrus.FieldLogger) (*SQLStore, error) {
 
 	switch strings.ToLower(url.Scheme) {
 	case "sqlite", "sqlite3":
```
crspeller: Why bother supporting sqlite?

internal/supervisor/cluster.go
```diff
@@ -0,0 +1,176 @@
...
+			s.workers.Release(1)
+			lock.Unlock()
+		}()
+
+		newState, err := s.transitionCluster(cluster, logger)
+		if err != nil {
+			logger.WithError(err).Error("transition cluster failed")
+		}
+
+		// Transition the state even if an error occurred.
```
crspeller: ... because failure is represented in the states.

internal/supervisor/cluster.go
```diff
@@ -0,0 +1,176 @@
...
+		if !more {
+			break
+		}
+	}
+
+	return nil
+}
+
+// DoOne looks for work to be done on a single cluster, and does it.
+func (s *ClusterSupervisor) DoOne() (bool, error) {
```
crspeller: Reminder comment about what we discussed offline. Moving this to having a queue of workers waiting for their locks.

internal/supervisor/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+
+// run is the main thread of the scheduler and responsible for triggering the doer as required.
+func (s *Scheduler) run() {
+	for {
+		// Enabling doing only if a non-zero interval was configured.
+		var poll <-chan time.Time
+		var notify <-chan bool
+		if s.period > 0 {
+			poll = time.After(s.period)
+			notify = s.notify
```
crspeller: Shouldn't notify be assigned anyway? So that manual Do functions if period is 0? Otherwise what's the point of a period of 0?

internal/provisioner/kops_provisioner.go
lieut-data: The code used to interact with the store to create the cluster, but it's now created much earlier in the pipeline. I /could/ return the updated `*model.Cluster`, but it's weird to have to write:
```
return cluster, err
```

instead of the usual:
```
return nil, err
```

and if I do the latter, the caller has to avoid clobbering its existing `cluster` to record the updated error state. So I figured it would be easier to just let the parameter be modified directly.

cmd/cloud/server.go
```diff
@@ -50,20 +62,46 @@ var serverCmd = &cobra.Command{
 		s3StateStore, _ := command.Flags().GetString("state-store")
 		logger.Infof("Using state store %s", s3StateStore)
 
+		// Limit the number of concurrent background jobs within this instance.
+		jobs, _ := command.Flags().GetInt("jobs")
+		workers := semaphore.NewWeighted(int64(jobs))
+		defer func() {
+			if err := workers.Acquire(context.TODO(), int64(jobs)); err != nil {
```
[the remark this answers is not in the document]
lieut-data: I agree: `context.Background` would be more appropriate here.

internal/model/aws_metadata.go
lieut-data: It's worth checking!

internal/store/store.go
lieut-data: Primarily for local development -- it's slightly easier to manage/switch between a local sqlite than a running Postgres instance.

internal/supervisor/scheduler.go
lieut-data: That was the original behaviour, but it proved to be somewhat useless to run the scheduler manually -- it would only happen on an API call, and doesn't "guarantee" that changes are made (e.g. if locks are acquired).

My primary use case here is to be able to run an instance against a shared database /without/ triggering background jobs unintentionally (e.g. by effectively adding a node to the cluster). Towards that end, setting a period of `0` disables the scheduler altogether, and is currently documented as such in the constructor function:
```
// If the period is zero, the scheduler is never run, even if manually run. Otherwise, the period
// specifies how long to wait after its last successful execution.
```

Maybe the manual `notify` is overkill, and I should just set a lower period so that the need to poke becomes irrelevant?

crspeller: Ok. That makes sense. I seen that it was documented but I thought it was strange enough to ask about.

Probably overkill, but I do like the concept of my commands poking the scheduler no matter the period.

lieut-data:
Thanks again, @crspeller :)

PR #14, opened 2019-05-08
[MM-15453] Refactor k8s client

gabrieljackson:
This introduces a few changes to the provisioning server:
 - k8s client now exposes clientsets to other packages
 - Added testing of k8s package
 - Moved mattermost-operator deployment image from `test` tag
   to `latest`
 - Added helper for getting pods in deployment

Fixes https://mattermost.atlassian.net/browse/MM-15453

Context for this PR:

After further experimentation, I have exposed the different kubernetes clientsets outside of the `k8s` package. The initial idea was to wrap the clientsets in functions that would only expose the necessary bits that we required for common tasks. This proved to be somewhat inflexible and also led to needless extra code in many places.

To expose the clientsets, they are now part of the `KubeClient` struct. This allows for us to mock them out for tests as well.


internal/tools/k8s/roles_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestClusterRolesBetaV1(t *testing.T) {
+	testClient := newTestKubeClient()
+	clusterRole := &rbacbetav1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-deployment"},
+	}
+
+	t.Run("create cluster role", func(t *testing.T) {
+		result, err := testClient.createClusterRoleBetaV1(clusterRole)
+		assert.NoError(t, err)
```
lieut-data: suggest `require.NoError` for this kind of thing, otherwise the `assert.Equal` will likely panic on trying to access a `nil` `result`.

internal/tools/k8s/helpers.go
```diff
@@ -2,27 +2,28 @@ package k8s
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 // WaitForPodRunning will poll a given kubernetes pod at a regular interval for
 // it to enter the 'Running' state. If the pod fails to become ready before
 // the provided timeout then an error will be returned.
-func (kc *KubeClient) WaitForPodRunning(name, namespace string, timeout int) (apiv1.Pod, error) {
+func (kc *KubeClient) WaitForPodRunning(namespace, podName string, timeout int) (*corev1.Pod, error) {
```
lieut-data: Might we accept a `context` instead of a `timeout` instead? This would allow the client to define the cancellation semantics (whether time-based, or an explicit `cancel`, etc.)

internal/tools/k8s/client.go
```diff
@@ -20,9 +26,27 @@ func New(configLocation string, logger log.FieldLogger) (*KubeClient, error) {
 		return &KubeClient{}, err
 	}
 
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return &KubeClient{}, err
```
lieut-data: Why return `&KubeClient{}` instead of `nil`?

internal/provisioner/kops_provisioner.go
```diff
@@ -187,14 +187,27 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
...
-	if err != nil {
-		return err
+	operators := []string{"mysql-operator", "mattermost-operator"}
+	for _, operator := range operators {
+		pods, err := k8sClient.GetPodsFromDeployment(operator, operator)
+		if err != nil {
+			return err
+		}
+		if len(pods.Items) == 0 {
+			return fmt.Errorf("no pods found from %q deployment", operator)
```
lieut-data: Are we guaranteed to find pods this early? Or might the pods themselves be slow to start?

operator-manifests/mattermost-operator/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:test
+          image: mattermost/mattermost-operator:latest
```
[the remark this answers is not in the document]
gabrieljackson: Yeah, it was always non-deterministic in that we weren't able to lock to a specific version. We should definitely track that requirement and start locking in versions once things are stable and we have most of the functionality that we need. For now, I think this is fine as we aren't updating the `test` tag for the time being.

internal/provisioner/kops_provisioner.go
gabrieljackson: This is actually something I am unsure of. I never ran into an issue, but it is definitely a concern that I will look into further.

Very shortly after the deployment is created (if not immediately) the pod objects will be created. This will happen far before they ever get scheduled on the cluster. That said, there may be a small period of time where the deployment exists, but no pods. I will investigate further, but propose that we move forward with this for now.

jwilander:
Real busy today but going to try to look at this tomorrow morning before our team meeting

gabrieljackson:
No rush. Let me know if you want me to swap you out for someone else on the review.

internal/provisioner/kops_provisioner.go
jwilander: I'm good with using this for now. Might be worth having a chat with @cpanato too about best ways to check the ready status of operators

internal/tools/k8s/helpers.go
```diff
@@ -12,14 +14,11 @@ import (
...
-	timer := time.NewTimer(time.Duration(timeout) * time.Second)
-	defer timer.Stop()
-
+func (kc *KubeClient) WaitForPodRunning(namespace, podName string, ctx context.Context) (*corev1.Pod, error) {
 	for {
 		select {
-		case <-timer.C:
-			return &corev1.Pod{}, errors.New("timed out waiting for pod to become ready")
+		case <-ctx.Done():
+			return &corev1.Pod{}, errors.Wrap(ctx.Err(), "timed out waiting for pod to become ready")
```
lieut-data: return `nil` here instead of an empty struct?

internal/tools/k8s/helpers_test.go
```diff
@@ -26,10 +28,18 @@ func TestWaitForPodRunning(t *testing.T) {
...
 	t.Run("wait for running", func(t *testing.T) {
-		pod, err := testClient.WaitForPodRunning(namespace, podName, 30)
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		defer cancel()
+		pod, err := testClient.WaitForPodRunning(namespace, podName, ctx)
 		require.NoError(t, err)
 		assert.Equal(t, corev1.PodRunning, pod.Status.Phase)
 	})
+	t.Run("don't wait for running", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 0*time.Second)
```
lieut-data: I wonder if constructing with `0` as the timeout guarantees the `context` will be immediately done or if a fast test might sometimes "pass". Perhaps use `context.WithCancel` and then immediately cancel it instead?

internal/tools/k8s/helpers.go
lieut-data: Can't comment on the `time.Sleep` below (I'm reviewing from a specific commit), but might we enable cancelling that timer (so we don't have to wait up to 5 seconds after a cancelled context) by structuring the code as follows instead?

```go
for {
    pod, err := ...

    select {
        case <-ctx.Done():
            return nil, errors.Wrap(...)
        case <-time.After(5 * time.Second):
    }
}
```

gabrieljackson: If we do that, we will then have to wait for 5 seconds even if the pod is immediately ready.

gabrieljackson: I suppose we could do a pod check out of the loop and then enter the loop if that doesn't succeed...

Hmm, not sure if that behavior is desirable or not.

lieut-data: Hmm, I’m not sure I follow — why wouldn’t we return after the `pod, err` line using the same semantics as the existing code?

gabrieljackson: Oh, I see what you are proposing now! I will push up something in a bit.

PR #15, opened 2019-05-11
MM-14999: Installation API

lieut-data:
This changelist introduces the CLI API, REST API, store updates and
supervisor additions to manage installations and cluster installations.
Created installations attempt to schedule a cluster installation on the
first available cluster. No attempt is currently made to wait for the cluster installation to actually stabilize, but I anticipate integrating that with pending changes from @gabrieljackson.

While the API remains async as previously changed, I've ripped out the
in-process parallelism in an attempt to simplify the data model
surrounding locking. This will need to be revisited when we scale the
provisioning server horizontally, but for now I haven't been able to come up with a straightforward strategy that avoids lock contention. Very open to suggestions!


internal/supervisor/installation.go
```diff
@@ -0,0 +1,340 @@
...
+
+	oldState := installation.State
+	installation.State = newState
+	err = s.store.UpdateInstallation(installation)
+	if err != nil {
+		logger.WithError(err).Warnf("Failed to set installation state to %s", newState)
+		return
+	}
+
+	logger.Debugf("Transitioned installation from %s to %s", oldState, newState)
```
jwilander: I wonder if this would be good to log at INFO level rather than DEBUG or would that be too spammy?

internal/provisioner/kops_provisioner.go
```diff
@@ -74,7 +77,7 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
 		return errors.Wrapf(err, "failed to stat cluster directory %q", outputDir)
 	}
 
-	logger.WithField("name", kopsMetadata.Name).Info("creating cluster")
```
gabrieljackson: Thanks for standardizing all of this!

internal/supervisor/installation.go
```diff
@@ -0,0 +1,340 @@
...
+		err = s.store.UpdateClusterInstallation(clusterInstallation)
+		if err != nil {
+			logger.WithError(err).Warnf("Failed to mark cluster installation %s for deletion", clusterInstallation.ID)
+			return installation.State
+		}
+
+		deletingClusterInstallations++
+	}
+
+	logger.Debugf(
```
gabrieljackson: Good place for `WithFields()` in my opinion.

internal/supervisor/installation.go
lieut-data: I personally found this a little spammy, and tried to make sure that the INFO-level logs capture all necessary detail for still being able to understand the meaningful changes here. Propose we run debug enabled in production anyway, but filter out downstream.

cmd/cloud/installation.go
```diff
@@ -1,15 +1,178 @@
...
 var installationCmd = &cobra.Command{
 	Use:   "installation",
 	Short: "Manipulate installations managed by the provisioning server.",
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("TODO: implement installation command")
+}
+
+var installationCreateCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Create an installation.",
```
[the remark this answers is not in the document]
lieut-data: I agree that installation as a concept is likely overkill given our current plans to make them 1:1 with a cluster installation. My hope was that by forcing the concept of a multi-cluster installation on the domain, we'd avoid painting ourselves into a naming corner that would be even more confusing in the future. Very happy to iterate on this, though!

cmd/cloud/installation.go
```diff
@@ -1,15 +1,178 @@
...
+			return err
+		}
+
+		return nil
+	},
+}
+
+var installationUpgradeCmd = &cobra.Command{
+	Use:   "upgrade",
+	Short: "Upgrade (or downgrade) the version of Mattermost.",
```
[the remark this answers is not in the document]
lieut-data: 5/5 that we should support downgrades: otherwise there is no way to rollback an installation gone awry. The current plan is to let the operator decide how far back to allow such a downgrade -- e.g. we may not support downgrading across major versions.

internal/supervisor/installation.go
lieut-data: 2/5, but I'm not sure fields are fungible for sprintf-statements in all cases. Certainly we'd want structured logging for pretty much all common verbiage (e.g. the `cluster` and `installation` we do today), but are we actually looking to "parse" `cluster_installation_deleted_count`, etc.?

lieut-data:
Thanks for the careful reviews and feedback, @jwilander & @gabrieljackson! A few comments above, plus some improvements based on the requested feedback :)

internal/supervisor/installation.go
gabrieljackson: In my head the fields would be no different that what you had: `deleting`, `failed`, etc. You would narrow the search down with a second criteria something else in the log message if you just wanted these messages.

We can always implement this later if we feel like it!

PR #18, opened 2019-05-20
MM-15205: Group Rest API

lieut-data:
This introduces the `/api/group` endpoints, as well as new `/api/installation` endpoints for joining and leaving a group.

The API supports changing the target version of the installation group, but defers the propagation of same to the corresponding installations to https://mattermost.atlassian.net/browse/MM-15369.

I also fixed a minor caching issue in `internal/tools/k8s` with the use of `TempDir` that precludes `go test ./...` from successfully caching these tests.

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -380,3 +380,132 @@ func (c *Client) SetClusterInstallationConfig(clusterInstallationID string, conf
 		return errors.Errorf("failed with status code %d", resp.StatusCode)
 	}
 }
+
+// CreateGroup requests the creation of a group from the configured provisioning server.
+func (c *Client) CreateGroup(request *CreateGroupRequest) (*model.Group, error) {
+	resp, err := c.doPost(c.buildURL("/api/groups"), request)
+	if err != nil {
+		return nil, err
+	}
+	defer closeBody(resp)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return model.GroupFromReader(resp.Body)
+
+	default:
+		return nil, errors.Errorf("failed with status code %d", resp.StatusCode)
+	}
+}
+
+// UpdateGroup updates the installation group.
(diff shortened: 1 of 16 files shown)


internal/api/group.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	c.Supervisor.Do()
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// handleDeleteGroup responds to DELETE /api/groups/{group}, marking the group as deleted.
+//
+// Installations will not automatically leave the group, but they will no longer consider the group version as an upgrade target.
```
gabrieljackson: nit: this comment line is a tad long. Perhaps we could wrap it.

internal/api/client.go
```diff
@@ -380,3 +380,132 @@ func (c *Client) SetClusterInstallationConfig(clusterInstallationID string, conf
...
+		return err
+	}
+	defer closeBody(resp)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return nil
+
+	default:
+		return errors.Errorf("failed with status code %d", resp.StatusCode)
```
gabrieljackson: super nit: this error message is used heavily throughout the code. Perhaps we may want a helper function to return it in a standard way to make updating it in the future a little easier. 0/5 on this.

lieut-data: I agree that this needs some attention. I'd like to also expose the actual error message to the client vs. requiring the server logs to explain. I'll add this as a TODO to revisit holistically.

lieut-data:
@gabrieljackson, your understanding of the `GroupID` being left set on an installation is correct. I had hoped to avoid cascading the necessary `UPDATE` to all associated installations (which might be in various states of being locked, etc.). This also has the nice advantage of technically being reversible (though the planned history tables would have enabled that anyway.)

PR #19, opened 2019-05-29
[MM-15368] Support installation version changes

gabrieljackson:
This change introduces support for spinning up installations with
specified versions. It also allows for changing the version of a
previously-running installation.

Fixes https://mattermost.atlassian.net/browse/MM-15368

### DIFF
--- a/internal/api/installation_test.go
+++ b/internal/api/installation_test.go
@@ -468,15 +468,6 @@ func TestUpgradeInstallation(t *testing.T) {
 		require.Equal(t, model.InstallationStateUpgradeRequested, installation1.State)
 	})
 
-	t.Run("while stable, to invalid version", func(t *testing.T) {
-		installation1.State = model.InstallationStateStable
-		err = sqlStore.UpdateInstallation(installation1)
-		require.NoError(t, err)
-
-		err = client.UpgradeInstallation(installation1.ID, "invalid")
-		require.EqualError(t, err, "failed with status code 400")
-	})
-
 	t.Run("after deletion failed", func(t *testing.T) {
 		installation1.State = model.InstallationStateDeletionFailed
 		err = sqlStore.UpdateInstallation(installation1)
--- a/internal/api/installation_test.go
+++ b/internal/api/installation_test.go
@@ -489,6 +489,19 @@ func TestUpgradeInstallation(t *testing.T) {
 		err = client.UpgradeInstallation(installation1.ID, "latest")
 		require.EqualError(t, err, "failed with status code 400")
 	})
+
+	t.Run("installation record updated", func(t *testing.T) {
+		installation1.State = model.InstallationStateStable
+		err = sqlStore.UpdateInstallation(installation1)
+		require.NoError(t, err)
+
+		err = client.UpgradeInstallation(installation1.ID, "5.9.0")
+		require.NoError(t, err)
+
+		installation1, err = client.GetInstallation(installation1.ID)
+		require.NoError(t, err)
+		require.Equal(t, "5.9.0", installation1.Version)
+	})
 }
 
 func TestJoinGroup(t *testing.T) {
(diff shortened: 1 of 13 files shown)


internal/api/installation_test.go
```diff
@@ -468,15 +468,6 @@ func TestUpgradeInstallation(t *testing.T) {
 		require.Equal(t, model.InstallationStateUpgradeRequested, installation1.State)
 	})
 
-	t.Run("while stable, to invalid version", func(t *testing.T) {
```
lieut-data: Can we replace this test with one that asserts the requested version is written to the installation record?

internal/provisioner/kops_provisioner.go
```diff
@@ -436,8 +436,13 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      makeClusterInstallationName(clusterInstallation),
 			Namespace: clusterInstallation.Namespace,
+			Labels: map[string]string{
```
lieut-data: 🎉

internal/model/installation.go
```diff
@@ -6,23 +6,25 @@ import (
 )
 
 const (
-	// InstallationStateCreationRequested is a installation in the process of being created.
```
lieut-data: 🤦‍♂

internal/provisioner/kops_provisioner.go
```diff
@@ -503,3 +508,119 @@ func (provisioner *KopsProvisioner) DeleteClusterInstallation(cluster *model.Clu
...
+	cr, err := k8sClient.MattermostClientset.MattermostV1alpha1().ClusterInstallations(clusterInstallation.Namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get cluster installation %s", clusterInstallation.ID)
+	}
+
+	logger.WithField("status", fmt.Sprintf("%+v", cr.Status)).Debug("Got cluster installation")
+
+	version := translateMattermostVersion(installation.Version)
+	if cr.Spec.Version == version {
+		return nil
```
lieut-data: Should we log something here that we decided not to take any action?

internal/supervisor/cluster_installation.go
```diff
@@ -159,13 +163,28 @@ func (s *ClusterInstallationSupervisor) transitionClusterInstallation(clusterIns
 
 		err = s.store.DeleteClusterInstallation(clusterInstallation.ID)
 		if err != nil {
-			logger.WithError(err).Error("Failed to record deleted cluster installation after creation")
+			logger.WithError(err).Error("Failed to record deleted cluster installation after deletion")
```
lieut-data: 👍

gabrieljackson:
Thanks for the review @lieut-data!

You are correct that any arbitrary version can be used with respect to the idea that at the end of the day operator is going to look for a docker image of `IMAGE_NAME:VERSION`. This presents a lot of flexibility, but also some challenges. The primary challenge I see is that we really can't do any sanitization of version number to see if it is correct or not unless we want to limit the docker image tags we support.

@cpanato explained to me recently that all PRs get a docker image built and pushed with the tag being the PR number. (apologies if I misunderstood that, but I believe that is what he said). This will work well for spinmints and other test clusters I believe, but also means we can't really limit the semantics of the version number at this time.

All of this may need to be altered in the future, but I believe the implementation I have in this PR will at least let us continue to develop and see how it goes before making further adjustments.

internal/model/installation.go
gabrieljackson: Not a big deal! I am sure it was just a minor copy+pasta error.

internal/provisioner/kops_provisioner.go
gabrieljackson: Sure!

internal/api/installation_test.go
gabrieljackson: Sure thing!

PR #20, opened 2019-06-04
[MM-15715] Add AWS client and support route 53 DNS

gabrieljackson:
This change introduces the following:
 - Client for interacting with AWS resources.
 - Support for configuring Mattermost DNS via route 53.
 - ClusterInstallations are now deployed with ELB services.
 - Newest Mattermost ClusterInstallation CRD file.

Fixes https://mattermost.atlassian.net/browse/MM-15715


internal/supervisor/installation.go
```diff
@@ -32,6 +33,13 @@ type installationProvisioner interface {
 	CreateClusterInstallation(cluster *model.Cluster, installation *model.Installation, clusterInstallation *model.ClusterInstallation) error
 	DeleteClusterInstallation(cluster *model.Cluster, installation *model.Installation, clusterInstallation *model.ClusterInstallation) error
 	UpdateClusterInstallation(cluster *model.Cluster, installation *model.Installation, clusterInstallation *model.ClusterInstallation) error
+	GetClusterInstallationResource(cluster *model.Cluster, installation *model.Installation, clusterInstallation *model.ClusterInstallation) (*mmv1alpha1.ClusterInstallation, error)
+}
+
+// aws abstracts the aws client operations required by the installation supervisor.
```
lieut-data: nit: it would nice to converge on a name that doesn't couple us to a particular cloud provider (for that mythical future ;P). Maybe `registrar` (meh) or `dnsManager`?

internal/supervisor/installation.go
```diff
@@ -237,6 +250,48 @@ func (s *InstallationSupervisor) createClusterInstallation(cluster *model.Cluste
...
+			logger.WithError(err).Error("Failed to get cluster installation resource")
+			return model.InstallationStateCreationDNS
+		}
+
+		endpoints = append(endpoints, cr.Status.Endpoint)
+	}
+
+	err = s.aws.CreateCNAME(installation.DNS, endpoints, logger)
+	if err != nil {
+		logger.WithError(err).Error("Failed to create DNS A record")
```
lieut-data: nit: `A record` vs. calling `CNAME`?

operator-manifests/mattermost/crds/mm_clusterinstallation_crd.yaml
```diff
@@ -101,4 +136,4 @@ spec:
   versions:
   - name: v1alpha1
     served: true
-    storage: true
+    storage: true
```
lieut-data: nit: trailing endline misconfiguration in editor?

internal/supervisor/installation.go
gabrieljackson: This is a good point. I am not entirely happy with how I mocked out the code for testing anyway, but haven't come up with anything better yet. I will think on it and may do a small follow-up PR to make some improvements.

PR #22, opened 2019-06-05
build docker image

cpanato:


cpanato:
Do we need other dependencies?

gabrieljackson:
Yeah, right now there are at least two other dependencies for running the provisioning server: kops and terraform. The code currently expects them to be in `/usr/local/bin`.

cpanato:
Will update
@gabrieljackson what are the versions required?

gabrieljackson:
Right now, I am running the latest versions that were out at the time I installed them:

```
gmbp ~/go/src/github.com/mattermost/mattermost-cloud (master) on$ kops version
Version 1.11.1
gmbp ~/go/src/github.com/mattermost/mattermost-cloud (master) on$ terraform version
Terraform v0.11.13
```

That said, I think we can probably use the newest versions that are out right now.

Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+## Runs lint against all packages.
+.PHONY: lint
+lint:
+	@echo Running lint
+	env GO111MODULE=off $(GO) get -u golang.org/x/lint/golint
+	golint -set_exit_status ./...
+	@echo lint success
+
+# Runs govet against all packages.
```
gabrieljackson: nit: single `#` where the other comments have two.

build/bin/entrypoint
```diff
@@ -0,0 +1,12 @@
...
+# This is documented here:
+# https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines
+
+if ! whoami &>/dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-mattermost-cloud}:x:$(id -u):$(id -g):${USER_NAME:-mattermost-cloud} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+
+exec ${CLOUD} server $@
```
gabrieljackson: We will need to change the provisioning server to accept env vars for server config parameters. Right now, there is a required flag that needs to be set here: `--route53-id`

.circleci/config.yml
```diff
@@ -2,82 +2,153 @@ version: 2.1
...
+    - *save_cache
+
+  push-docker:
+    executor:
+      name: default
+    steps:
+    - checkout
+    - setup_remote_docker
+    - run: make build-image
+    - run: |
```
lieut-data: Could this script go into the `Makefile` too, in the unlikely event we want to be able to tag/push a Docker image manually?

Makefile
```diff
@@ -0,0 +1,51 @@
+GO ?= $(shell command -v go 2> /dev/null)
+MATTERMOST_CLOUD_IMAGE ?= mattermost/mattermost-cloud:test
+MACHINE = $(shell uname -m)
+BUILD_IMAGE = golang:1.12
```
lieut-data: Could we call this something like `DOCKER_BUILD_IMAGE` and `DOCKER_BASE_IMAGE`? Their top-level nature suggests they apply to the project as a whole, so it would be good to clarify.

Makefile
```diff
@@ -0,0 +1,51 @@
...
+BASE_IMAGE = alpine:3.9
+GOPATH ?= $(shell go env GOPATH)
```
lieut-data: This is a go-module-enabled repository, so the `GOPATH` should be immaterial. Can you elaborate on why we need to define/trim it below?

Makefile
```diff
@@ -0,0 +1,51 @@
...
+BUILD_HASH := $(shell git rev-parse HEAD)
+
+export GO111MODULE=on
+
+## Checks the code style, tests, builds and bundles.
+all: check-style dist
+
+## Runs govet and gofmt against all packages.
+.PHONY: check-style
+check-style: lint govet
```
lieut-data: I think `govet` should be run first, since it will complain about compilation errors that `lint` might choke on too.

Makefile
```diff
@@ -0,0 +1,51 @@
...
+BASE_IMAGE = alpine:3.9
+GOPATH ?= $(shell go env GOPATH)
+GOFLAGS ?= $(GOFLAGS:)
+BUILD_TIME := $(shell date -u +%Y%m%d.%H%M%S)
+BUILD_HASH := $(shell git rev-parse HEAD)
+
+export GO111MODULE=on
+
+## Checks the code style, tests, builds and bundles.
+all: check-style dist
```
lieut-data: We don't actually have a `test` target here -- intentional?

build/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN make build
+
+# Final Image
+FROM ${BASE_IMAGE}
+LABEL name="Mattermost Cloud" \
+  maintainer="dev-ops@mattermost.com" \
+  vendor="Mattermost" \
+  distribution-scope="public" \
+  architecture="x86_64" \
+  url="https://mattermost.dev" \
```
lieut-data: Why mattermost.dev?

build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x
```
lieut-data: Can we `set -euo pipefail` here for maximum safety?

build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
...
+
+# ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
+mkdir -p ${HOME}
+chown ${USER_UID}:0 ${HOME}
+chmod ug+rwx ${HOME}
+
+# runtime user will need to be able to self-insert in /etc/passwd
+chmod g+rw /etc/passwd
+
+# no need for this script to remain in the image after running
```
lieut-data: This seems counter to how a script should be run: propose removing it in the caller instead of self-removing.

build/bin/entrypoint
cpanato: we can set that in the args
but will be better and easy if accept env vars

.circleci/config.yml
cpanato: we can do that later, this is specific for circleci since it uses the env vars from there

Makefile
cpanato: it is being used to trim the path, if you think it is not right let me know I can remove

Makefile
cpanato: will migrate that later

build/Dockerfile
cpanato: no reason, changed to .com

build/bin/user_setup
cpanato: this is from the operator, just re-using the same thing here

cpanato:
@gabrieljackson @lieut-data PTAL

do we need helm as well?

stylianosrigas:
@cpanato we do. The provisioner will run helm install for Prometheus and Fluentd.

cpanato:
@stylianosrigas helm added

cpanato:
will add kubectl

gabrieljackson:
We should get this merged into master now that we know it is working. Would it be possible to rebase and force push up one fresh commit to clean things up a bit?

cpanato:
we need to merge this first https://github.com/mattermost/mattermost-cloud/pull/31

gabrieljackson:
I wanted to note that https://github.com/mattermost/mattermost-cloud/pull/29 adds some new flags that are required to start the server.

cpanato:
this image is running on test account

@gabrieljackson @lieut-data PTAL

Makefile
lieut-data: Sounds good -- should we update the comment now, and add both later?

build/bin/user_setup
lieut-data: I confess I haven't stayed up to date with the operator -- can you elaborate on what it's doing that makes that required here?

Makefile
lieut-data: GOPATH as it exists today will be deprecated in an upcoming Go release, but it sounds like we're just trying to remove the current working directory -- could we trim that instead?

cpanato:
@lieut-data it is not required to have the `build/bin/user_setup` i thought this is ok to have, but i will confess that maybe is not need at all.

regarding the gopath I've removed that

PR #29, opened 2019-06-12
Mm 15951 Add internal-nginx and prometheus helm installation support, helm charts and Route53 DNS registration

stylianosrigas:
Mm 15951 Add internal-nginx and prometheus helm installation support, helm charts and Route53 DNS registration

Tickets: 
- https://mattermost.atlassian.net/browse/MM-15511
- https://mattermost.atlassian.net/browse/MM-15951

With this PR:

- Prometheus is installed by the provisioner in every new cluster
- Internal Nginx is installed by the provisioner in every new cluster
- Prometheus DNS registered with private hosted zone defined by --private-route53-id and --private-dns flags
- Prometheus Route53 registration removed when the cluster is deleted by the provisioner


internal/provisioner/kops_provisioner.go
```diff
@@ -233,12 +255,85 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
 		return err
 	}
 
+	// Old tiller pod is deleted because of the upgrade action and therefore wait is needed
+	logger.Infof("Waiting %d seconds for the old Tiller pod to be replaced...", wait)
+	time.Sleep(time.Duration(wait) * time.Second)
```
gabrieljackson: Hmm, if we just wait 60 seconds here, is it possible that everything may not be ready?

Also, if it completes sooner than 60 seconds, we will wait the full time regardless. As you can see above, we developed helper functions to wait for something to happen and then return asap. We should do that here as well.

internal/provisioner/kops_provisioner.go
```diff
@@ -233,12 +255,85 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
...
+			endpoint := service.Status.LoadBalancer.Ingress[0].Hostname
+			logger.Infof("Succesfully got LoadBalancer endpoint %s for Namespace %s", endpoint, namespace)
+			return endpoint, nil
+		}
+	}
+	return "", nil
+}
+
+// helmInstallation is used to install Helm charts.
+func helmInstallation(chart helmDeployment, logger log.FieldLogger, kops *kops.Cmd) error {
```
gabrieljackson: This could probably have a more descriptive. Maybe `installHelmChart()`?

gabrieljackson: Also, we are passing in a pointer to kops.Cmd here, but all we need it for is the config path, so we should just pass in the path string instead.

```suggestion
func helmInstallation(chart helmDeployment, logger log.FieldLogger, configPath string) error {
```

This should be done for the other functions for helm if they only need the config path string as well.

internal/provisioner/kops_provisioner.go
```diff
@@ -233,12 +255,85 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
...
+	}
+
 	logger.WithField("name", kopsMetadata.Name).Info("Successfully created cluster")
 
 	return nil
 }
 
-// HelmSetup is used for the initial setup of Helm in cluster
+// getEndpoint is used to get the endpoint of the internal ingress.
+func getEndpoint(namespace string, logger log.FieldLogger, kops *kops.Cmd) (string, error) {
```
gabrieljackson: This could probably have a more descriptive. Maybe `getLoadBalancerEndpoint()`?

The name and the comment above don't reflect that this will only return an endpoint if the service is of a loadbalancer type.

internal/provisioner/kops_provisioner.go
```diff
@@ -233,12 +255,85 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
...
+			return endpoint, nil
+		}
+	}
+	return "", nil
+}
+
+// helmInstallation is used to install Helm charts.
+func helmInstallation(chart helmDeployment, logger log.FieldLogger, kops *kops.Cmd) error {
+	logger.Infof("Installing helm chart %s", chart.chartName)
+	if chart.chartName == "stable/prometheus" {
```
gabrieljackson: Instead of checking the chart name, which could be brittle, couldn't we modify the `helmDeployment` struct to have a `deployCmd` value that is set when we create the object.

Then you could just have `cmd := exec.Command(chart.deployCmd)`.

internal/provisioner/kops_provisioner.go
```diff
@@ -26,15 +26,37 @@ type KopsProvisioner struct {
...
+	CreateCNAME(dnsName string, dnsEndpoints []string, logger log.FieldLogger) error
+	DeleteCNAME(dnsName string, logger log.FieldLogger) error
+}
+
+// Array of helm apps that need DNS registration
+var helmApps = []string{"prometheus"}
+
 // NewKopsProvisioner creates a new KopsProvisioner.
-func NewKopsProvisioner(clusterRootDir, s3StateStore, certificateSslARN string, logger log.FieldLogger) *KopsProvisioner {
+func NewKopsProvisioner(clusterRootDir string, s3StateStore string, certificateSslARN string, logger log.FieldLogger, aws aws, privateDNS string) *KopsProvisioner {
```
```suggestion
func NewKopsProvisioner(clusterRootDir, s3StateStore, certificateSslARN string, logger log.FieldLogger, aws aws, privateDNS string) *KopsProvisioner {
```

nit: you don't need to specify string for each variable when passing in multiple strings in a row

cmd/cloud/server.go
```diff
@@ -85,12 +89,16 @@ var serverCmd = &cobra.Command{
...
 
+		privateRoute53ZoneID, _ := command.Flags().GetString("private-route53-id")
+		privateDNS, _ := command.Flags().GetString("private-dns")
 		// Setup the provisioner for actually effecting changes to clusters.
 		kopsProvisioner := provisioner.NewKopsProvisioner(
 			clusterRootDir,
 			s3StateStore,
 			certificateSslARN,
 			logger,
+			aws.New(privateRoute53ZoneID),
```
gabrieljackson: This is my primary concern in this PR. I can see how storing an aws client in the provisioner would be useful, but this is a bit of an anti-pattern.

If you look here, you will see how we are passing in an aws client to the Installation supervisor for use with the installation changes. This should be mimicked for `NewClusterSupervisor` instead of this implementation.

Also, we should probably modify the aws client itself to hold a DNS value so we can store the private DNS in the client itself.

internal/provisioner/kops_provisioner.go
stylianosrigas: @gabrieljackson I used the helper function initially. The problem with this is that when Helm inits the first time it creates a pod and after some configuration the upgrade command forces the deletion of this pod and creation of a new one. If you use the helper to wait for a pod creation sometimes you end up with it waiting for the terminating pod to come up and it fails. That is why I added a 60 second sleep as this is enough for the tiller pod to come up.

helm-charts/private-nginx_values.yaml
```diff
@@ -0,0 +1,406 @@
+## nginx configuration
```
cpanato: maybe we can have here only the custom values the other we can remove

stylianosrigas: Left only custom values for both nginx and prometheus.

cmd/cloud/server.go
stylianosrigas: As discussed, I added a comment that aws is used only for Route53 registration and todo to move it to clustersupervisor in the future.

internal/provisioner/kops_provisioner.go
stylianosrigas: Automatically added by linter, fixed.

internal/provisioner/kops_provisioner.go
stylianosrigas: Fixed to check a chart.setArgument which is included only for charts that need extra arguments to be passed.

internal/provisioner/kops_provisioner.go
stylianosrigas: Fixed.

internal/provisioner/kops_provisioner.go
stylianosrigas: Fixed name plus to pass only the configpath needed.

internal/provisioner/kops_provisioner.go
stylianosrigas: I added a new check that checks the whole helm readiness. The reason is that even if we check the tiller pod and it moves to running state, tiller needs a couple of more seconds to be ready to accept commands from helm. By checking the whole helm status for up to a specific time we ensure that it moves forward only when really ready.

internal/provisioner/kops_provisioner.go
```diff
@@ -26,15 +26,37 @@ type KopsProvisioner struct {
 	s3StateStore      string
 	certificateSslARN string
 	logger            log.FieldLogger
+	aws               aws
+	privateDNS        string
 }
 
+// HelmDeployment deploys Helm charts.
```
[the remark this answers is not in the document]
stylianosrigas: fixed

internal/provisioner/kops_provisioner.go
```diff
@@ -26,15 +26,37 @@ type KopsProvisioner struct {
...
+type helmDeployment struct {
+	valuesPath          string
+	chartName           string
+	namespace           string
+	chartDeploymentName string
+	dns                 string
+}
+
+// aws abstracts the aws client operations required by the installation supervisor.
```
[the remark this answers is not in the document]
stylianosrigas: Using aws the same way we do in the installationsupervisor. Will be moved in the future as per comment.

internal/provisioner/kops_provisioner.go
```diff
@@ -422,6 +517,15 @@ func (provisioner *KopsProvisioner) DeleteCluster(cluster *model.Cluster) error
 		return errors.Wrap(err, "failed to clean up output directory")
 	}
 
+	for _, app := range helmApps {
+		logger.Infof("Deleting Route53 DNS Record for %s", app)
+		dns := fmt.Sprintf("%s.%s.%s", cluster.ID, app, provisioner.privateDNS)
+		err = provisioner.aws.DeleteCNAME(dns, logger)
+		if err != nil {
+			return errors.Wrap(err, "failed to delete Route53 DNS record for %s")
```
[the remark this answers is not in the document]
stylianosrigas: fixed

internal/tools/k8s/helpers.go
```diff
@@ -43,3 +43,12 @@ func (kc *KubeClient) GetPodsFromDeployment(namespace, deploymentName string) (*
 
 	return kc.Clientset.CoreV1().Pods(namespace).List(listOptions)
 }
+
+// GetServices gets the services that belong to a given namespace.
```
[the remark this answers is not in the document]
stylianosrigas: True, I don't think the services will be used anywhere else. Removed and k8s client is used directly.

helm-charts/prometheus_values.yaml
```diff
@@ -0,0 +1,6 @@
+server:
+  ingress:
+    enabled: true
+    annotations:
+      kubernetes.io/ingress.class: private-nginx
+
```
cpanato: new line

internal/provisioner/kops_provisioner.go
```diff
@@ -29,17 +30,31 @@ type KopsProvisioner struct {
...
+// helmDeployment deploys Helm charts.
+type helmDeployment struct {
+	valuesPath          string
+	chartName           string
+	namespace           string
+	chartDeploymentName string
+	setArgument         string
+}
+
+// Array of helm apps that need DNS registration
```
gabrieljackson: nit:
```suggestion
// Array of helm apps that need DNS registration.
```

internal/provisioner/kops_provisioner.go
```diff
@@ -251,12 +266,109 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		}
+	}
+
 	logger.WithField("name", kopsMetadata.Name).Info("Successfully created cluster")
 
 	return nil
 }
 
-// HelmSetup is used for the initial setup of Helm in cluster
+// waitForHelmRunning is used to check when Helm is ready to install charts
```
gabrieljackson: nit:
```suggestion
// waitForHelmRunning is used to check when Helm is ready to install charts.
```

internal/provisioner/kops_provisioner.go
```diff
@@ -251,12 +266,109 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		err := cmd.Run()
+		if err != nil {
+			return err
+		}
+	}
+	logger.Infof("Successfully installed helm chart %s", chart.chartName)
+	return nil
+}
+
+// helmSetup is used for the initial setup of Helm in cluster
```
gabrieljackson: nit:
```suggestion
// helmSetup is used for the initial setup of Helm in cluster.
```

internal/provisioner/kops_provisioner.go
```diff
@@ -251,12 +266,109 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		cmd := exec.Command("helm", "ls", "--kubeconfig", configPath)
+		var out bytes.Buffer
+		cmd.Stderr = &out
+		cmd.Run()
+		if out.String() == "" {
+			return nil
+		}
+		select {
+		case <-ctx.Done():
+			return (errors.Wrap(ctx.Err(), "timed out waiting for helm to become ready"))
```
gabrieljackson: nit:
```suggestion
			return errors.Wrap(ctx.Err(), "timed out waiting for helm to become ready")
```

internal/provisioner/kops_provisioner.go
```diff
@@ -251,12 +266,109 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		}
+	}
+	return "", nil
+}
+
+// installHelmChart is used to install Helm charts.
+func installHelmChart(chart helmDeployment, logger log.FieldLogger, configPath string) error {
+	logger.Infof("Installing helm chart %s", chart.chartName)
+	if chart.setArgument != "" {
+		cmd := exec.Command("helm", "install", "--kubeconfig", configPath, "--set", chart.setArgument, "-f", chart.valuesPath, chart.chartName, "--namespace", chart.namespace, "--name", chart.chartDeploymentName)
```
gabrieljackson: nit: you can clean up a line here with this instead.
```suggestion
		err := exec.Command("helm", "install", "--kubeconfig", configPath, "--set", chart.setArgument, "-f", chart.valuesPath, chart.chartName, "--namespace", chart.namespace, "--name", chart.chartDeploymentName).Run()
```

internal/provisioner/kops_provisioner.go
```diff
@@ -251,12 +266,109 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		err := cmd.Run()
+		if err != nil {
+			return err
+		}
+	} else {
+		cmd := exec.Command("helm", "install", "--kubeconfig", configPath, "-f", chart.valuesPath, chart.chartName, "--namespace", chart.namespace, "--name", chart.chartDeploymentName)
```
gabrieljackson: nit: same as above.

PR #31, opened 2019-06-12
Fallback to copyDirectory instead of os.rename when have invalid cross-device link error

cpanato:
When running the provisioner in docker we got this error:
```
time="2019-06-12T07:24:54Z" level=error msg="Failed to create cluster" cluster=zueqbod56fbj5xj3ka8fns4mfh error="failed to rename kops output directory to \"clusters/zueqbod56fbj5xj3ka8fns4mfh\": rename /tmp/kops-178583449/output clusters/zueqbod56fbj5xj3ka8fns4mfh: inval
id cross-device link" instance=uscqwmqjdb8xp8ok9cty34cz6c
```

this is caused because 
```
Renaming directories: Calling rename(2) for a directory is allowed only when both the source and the destination path are on the top layer. Otherwise, it returns EXDEV error (“cross-device link not permitted”). Your application needs to be designed to handle EXDEV and fall back to a “copy and unlink” strategy.
```
to fix we need to use the copy. Off course we can make the rename and if that fails we can fallback to copy, but since i was in a hurry did this dirty one.


internal/provisioner/kops_provisioner.go
```diff
@@ -105,7 +106,8 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
 		return err
 	}
 
-	err = os.Rename(kops.GetOutputDirectory(), outputDir)
+	// err = os.Rename(kops.GetOutputDirectory(), outputDir)
```
gabrieljackson: I would really like to attempt the `os.Rename` and then fall back to `copyFolder` if `err != nil` here.

Looking at all of the edge cases for copying files in go, it would be nice to try to do it cleanly first.

internal/provisioner/kops_provisioner.go
```diff
@@ -479,6 +481,7 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 			},
 		},
 		Spec: mmv1alpha1.ClusterInstallationSpec{
+			Replicas:               1,
```
gabrieljackson: I am guessing you didn't mean to change this?

internal/provisioner/kops_provisioner.go
```diff
@@ -663,3 +666,68 @@ func translateMattermostVersion(version string) string {
 
 	return version
 }
+
+func copyFolder(source string, dest string) (err error) {
+
```
gabrieljackson: nit: extra newline

internal/provisioner/kops_provisioner.go
```diff
@@ -663,3 +666,68 @@ func translateMattermostVersion(version string) string {
 
 	return version
 }
+
+func copyFolder(source string, dest string) (err error) {
```
```suggestion
func copyDirectory (source string, dest string) error {
```

internal/provisioner/kops_provisioner.go
```diff
@@ -663,3 +666,68 @@ func translateMattermostVersion(version string) string {
...
+			if err != nil {
+				fmt.Println(err)
+			}
+		}
+
+	}
+	return
+}
+
+func copyFile(source string, dest string) (err error) {
```
```suggestion
func copyFile(source string, dest string) error {
```

internal/provisioner/kops_provisioner.go
cpanato: yes, since we don't have a license, it should be one.
As soon we introduce the license we can make a check here if there is a license we make replicas = 2 otherwise keep in 1

internal/provisioner/kops_provisioner.go
cpanato: @gabrieljackson let me know if the change is good

cpanato:
@gabrieljackson PTAL

internal/provisioner/kops_provisioner.go
```diff
@@ -114,8 +116,17 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
 	}
 
 	err = os.Rename(kops.GetOutputDirectory(), outputDir)
-	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("failed to rename kops output directory to %q", outputDir))
+	if err != nil && err.(*os.LinkError).Err == syscall.EXDEV {
```
gabrieljackson: 👍

internal/provisioner/kops_provisioner.go
```diff
@@ -114,8 +116,17 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		err = utils.CopyDirectory(kops.GetOutputDirectory(), outputDir)
+		if err != nil {
+			return fmt.Errorf("failed to rename kops output directory to %q using utils.CopyFolder", outputDir)
+		}
+		err = os.RemoveAll(kops.GetOutputDirectory())
```
gabrieljackson: We actually already cleanup this dir with the `defer kops.Close()` line above.

This is what `kops.Close()` does:

```
// Close cleans up the temporary output directory used by kops.
func (c *Cmd) Close() error {
	return os.RemoveAll(c.tempDir)
}
```

cpanato: @gabrieljackson ah, so we dont need that?

gabrieljackson: Yes, exactly.

cpanato: will remove, thanks!

cpanato: @gabrieljackson done

PR #32, opened 2019-06-12
MM-16216 Add the ability to create clusters in existing subnets

jwilander:
##### Summary
The changes are fairly straight-forward once I figured out all the different pieces needed to use existing subnets.

Usage looks like this:

```
cloud server --state-store yourstatestore --route53-id yourroute53id --private-subnets subnetid1,subdnetid2 --public-subnets subnetid3,subnetid4 --certificate-aws-arn yourarn
```

##### Ticket Link
https://mattermost.atlassian.net/browse/MM-16216

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -73,6 +73,9 @@ var serverCmd = &cobra.Command{
 
 		s3StateStore, _ := command.Flags().GetString("state-store")
 		certificateSslARN, _ := command.Flags().GetString("certificate-aws-arn")
+		privateSubnetIds, _ := command.Flags().GetString("private-subnets")
+		publicSubnetIds, _ := command.Flags().GetString("public-subnets")
+		route53ZoneID, _ := command.Flags().GetString("route53-id")
 
 		wd, err := os.Getwd()
 		if err != nil {
@@ -85,14 +88,10 @@ var serverCmd = &cobra.Command{
 			"state-store":       s3StateStore,
 			"aws-arn":           certificateSslARN,
 			"working-directory": wd,
+			"private-subents":   privateSubnetIds,
+			"public-subnets":    publicSubnetIds,
+			"route53-id":        route53ZoneID,
 		}).Info("Starting Mattermost Provisioning Server")
-
-		privateSubnetIds, _ := command.Flags().GetString("private-subnets")
-		logger.Infof("Using private subnets %s", privateSubnetIds)
-		publicSubnetIds, _ := command.Flags().GetString("public-subnets")
-		logger.Infof("Using public subnets %s", publicSubnetIds)
-
-		route53ZoneID, _ := command.Flags().GetString("route53-id")
 
 		// Setup the provisioner for actually effecting changes to clusters.
 		kopsProvisioner := provisioner.NewKopsProvisioner(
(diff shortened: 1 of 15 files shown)


cmd/cloud/server.go
```diff
@@ -85,11 +87,21 @@ var serverCmd = &cobra.Command{
...
+		publicSubnetIds, _ := command.Flags().GetString("public-subnets")
+		logger.Infof("Using public subnets %s", publicSubnetIds)
+
+		route53ZoneID, _ := command.Flags().GetString("route53-id")
+
 		// Setup the provisioner for actually effecting changes to clusters.
 		kopsProvisioner := provisioner.NewKopsProvisioner(
 			clusterRootDir,
 			s3StateStore,
+			aws.New(route53ZoneID),
```
stylianosrigas: @jwilander In my PR https://github.com/mattermost/mattermost-cloud/pull/29 I am already passing an AWS client for private hosted zone ID, which we need for DNS registration. I guess you can use the same client for AWS related stuff.

internal/tools/kops/cluster.go
```diff
@@ -22,10 +22,24 @@ func (c *Cmd) CreateCluster(name, cloud string, clusterSize ClusterSize, zones [
...
 		arg("master-size", clusterSize.MasterSize),
+		arg("api-loadbalancer-type", "internal"),
+		arg("topology", "private"),
 		arg("target", "terraform"),
 		arg("out", c.GetOutputDirectory()),
 		arg("output", "json"),
-	)
+	}
+
+	if privateSubnetIds != "" {
```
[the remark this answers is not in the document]
jwilander: @gabrieljackson and I chatted about this. I'm pretty sure it should work fine if the subnets are not specified, with the added caveat that the clusters would be created with an internal load balancer and be private. I'm going to confirm and if it doesn't work then we'll only add the extra args if the subnets are provided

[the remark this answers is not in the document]
jwilander: https://community.mattermost.com/core/pl/qpcqjkwkcfd39caba6wxow58xe

jwilander:
Looks like I forgot to update the tests..

internal/provisioner/kops_provisioner.go
```diff
@@ -25,15 +27,19 @@ type KopsProvisioner struct {
...
 	s3StateStore      string
 	certificateSslARN string
+	privateSubnetIds  string
+	publicSubnetIds   string
 	logger            log.FieldLogger
 }
 
 // NewKopsProvisioner creates a new KopsProvisioner.
-func NewKopsProvisioner(clusterRootDir, s3StateStore, certificateSslARN string, logger log.FieldLogger) *KopsProvisioner {
+func NewKopsProvisioner(clusterRootDir, s3StateStore string, certificateSslARN, privateSubnetIds, publicSubnetIds string, logger log.FieldLogger) *KopsProvisioner {
```
gabrieljackson: nit:
```suggestion
func NewKopsProvisioner(clusterRootDir, s3StateStore, certificateSslARN, privateSubnetIds, publicSubnetIds string, logger log.FieldLogger) *KopsProvisioner {
```

internal/provisioner/kops_provisioner.go
```diff
@@ -151,6 +157,18 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster) error
 		return err
 	}
 
+	// Set the ELB tags for the public subnets
+	if provisioner.publicSubnetIds != "" {
```
gabrieljackson: nit: we normally do the `strings.Split` immediately after getting the flag value and then pass the `[]string` around, but this is totally fine!

PR #36, opened 2019-06-17
Mm 15513 Adding Fluentd installation in the cluster and fixing elb bug

stylianosrigas:
Ticket -> https://mattermost.atlassian.net/browse/MM-15513

- Adding Fluentd installation via the Provisioner server
- Fixing a bug when ELB gets more time to be created than expected and no endpoint can be retrieved


internal/provisioner/kops_provisioner.go
```diff
@@ -298,14 +300,21 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+	logger.Infof("Waiting up to %d seconds for internal ELB to be ready...", wait)
+	defer cancel()
+	endpoint, err := getLoadBalancerEndpoint(ctx, "internal-nginx", logger, kops.GetKubeConfigPath())
 	if err != nil {
 		return err
 	}
 
+	if endpoint == "" {
+		logger.Errorf("ELB endpoint is empty...")
+		return fmt.Errorf("ELB endpoint is empty")
```
```suggestion
		return errors.new("internal DNS ELB endpoint is empty")
```

internal/provisioner/kops_provisioner.go
```diff
@@ -298,14 +300,21 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
 	}
 
 	// Get the new ELB internal-nginx endpoint
-	endpoint, err := getLoadBalancerEndpoint("internal-nginx", logger, kops.GetKubeConfigPath())
+	logger.Infof("Waiting up to %d seconds for internal ELB to be ready...", wait)
+	defer cancel()
```
gabrieljackson: This would cancel a context created above which has already been counting down for some amount of time. A new context should be made here.

internal/provisioner/kops_provisioner.go
```diff
@@ -298,14 +300,21 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+	endpoint, err := getLoadBalancerEndpoint(ctx, "internal-nginx", logger, kops.GetKubeConfigPath())
 	if err != nil {
 		return err
 	}
 
+	if endpoint == "" {
+		logger.Errorf("ELB endpoint is empty...")
```
gabrieljackson: We don't really need to log the error and return it. It will be logged when it is returned.

internal/provisioner/kops_provisioner.go
```diff
@@ -298,14 +300,21 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+		return fmt.Errorf("ELB endpoint is empty")
+	}
+
 	for _, app := range helmApps {
-		logger.Infof("Registering DNS for %s", app)
 		dns := fmt.Sprintf("%s.%s.%s", cluster.ID, app, provisioner.privateDNS)
+		logger.Infof("Registering DNS for %s DNS:%s", app, dns)
```
```suggestion
		logger.Infof("Registering DNS %s for %s", dns, app)
```

internal/provisioner/kops_provisioner.go
stylianosrigas: New context created

stylianosrigas:
@gabrieljackson Thanks for the review, changes applied :)

PR #37, opened 2019-06-18
[MM-15596] add way to add a license to a mattermost installation

cpanato:
This PR add the ability to the provisioner server add a license to an mm installation in the creation part and as well in the upgrade

~Need to wait https://github.com/mattermost/mattermost-operator/pull/42~


internal/provisioner/kops_provisioner.go
```diff
@@ -715,6 +725,14 @@ func (provisioner *KopsProvisioner) DeleteClusterInstallation(cluster *model.Clu
 		return errors.Wrapf(err, "failed to delete cluster installation %s", clusterInstallation.ID)
 	}
 
+	secretName := fmt.Sprintf("%s-license", makeClusterInstallationName(clusterInstallation))
+	err = k8sClient.DeleteSecret(clusterInstallation.Namespace, secretName)
+	if k8sErrors.IsNotFound(err) {
+		logger.Infof("Secret %s/%s not found. Maybe the license was not set for this installation or was already deleted", clusterInstallation.Namespace, secretName)
```
gabrieljackson: We should be able to handle this better.

If the installation object has `""` for a license value, we should know that there shouldn't be a license file, so we can try to clean it up just in case and then log if we don't find anything like you are doing here.

That said, if the license value is not `""`, we may want to return an error here and not log.

internal/tools/k8s/secret.go
```diff
@@ -0,0 +1,63 @@
...
+		err := kc.DeleteSecret(nameSpace, secretName)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// DeleteSecret delete the given secret.
```
gabrieljackson: I don't think we need this function. We should just directly use the `kc.Clientset` instead of wrapping it like we are doing here.

One of my open PRs is doing that elsewhere to clean things up.

internal/tools/k8s/secret.go
```diff
@@ -0,0 +1,63 @@
...
+			Namespace: namespaceName,
+		},
+		StringData: map[string]string{
+			secretKey: secretValue,
+		},
+	}
+	return kc.Clientset.CoreV1().Secrets(namespaceName).Create(svcSpec)
+}
+
+// GetSecrets returns a list of kubernetes namespace objects if they exist.
```
gabrieljackson:
```suggestion
// GetSecrets returns a list of kubernetes secret objects if they exist.
```

internal/provisioner/kops_provisioner.go
cpanato: updated

internal/tools/k8s/secret.go
cpanato: It is similar to the namespace.go
not clear to me your comment.

cpanato:
@gabrieljackson 
1. Added in the upgrade because if there is an existing running cluster and you want to add the license, this, for my understanding, can be an upgrade. 
for the existing cluster, if you want to add the license we use the upgrade or otherwise, I think we need to add a new API to add the license. or am I wrong?

2. i'm not storing the license in the DB, but maybe this will be an issue when deleting the installation because the license field will be blank always. Maybe we can discuss to we decide what is the best option

internal/tools/k8s/secret.go
gabrieljackson: I am just saying we should delete this function entirely.

In another PR, I am doing the same to namespace.go and all the other functions that wrap a single Clientset command.

cpanato: done

cpanato:
@gabrieljackson @jwilander finish this and did the tests with a new installation with a license and a upgrade installation adding a license and works.

however to work the upgrade need a change in the operator side wich i will make a pr soon. stay tunned.

but you both can review now.

cpanato:
@gabrieljackson @jwilander PTAL

internal/provisioner/kops_provisioner.go
```diff
@@ -878,11 +900,21 @@ func (provisioner *KopsProvisioner) UpdateClusterInstallation(cluster *model.Clu
...
-		return nil
+		logger.Infof("Cluster installation already on version %s", version)
 	}
 	cr.Spec.Version = version
 
+	cr.Spec.MattermostLicenseSecret = ""
+	var secretName string
+	if installation.License != "" {
+		secretName = fmt.Sprintf("%s-license", name)
+		_, err = k8sClient.CreateSecret(clusterInstallation.Namespace, secretName, "license", installation.License)
```
gabrieljackson: Correct me if I am wrong, but this won't update the secret if there was a previously created secret right? The function will return with no error when it finds the previous secret and it won't be updated.

cpanato: good catch!
so we need to check if the secret has the same content if is different then update the secret.

the case can be: an existing installation have a license but that expires and we need to update the license.

I will refactor this part to make the appropriate checks

thanks for your review @gabrieljackson

gabrieljackson: Sure thing!

We may also want to make logic to remove a license if the value is set to `""` although that is less pressing than the update issue I noted above.

cpanato: done @gabrieljackson

PR #39, opened 2019-06-21
[MM-16413] Add separate cluster 'provisioning' state

gabrieljackson:
This new state contains the logic for creating the k8s operators
that are necessary to deploy Mattermost installations. This process
used to occur directly after cluster creation with no separation.
Now, clusters can be 'provisioned' again in the stable state, which
will update all operator resources to the newest version contained
in the provisioner manifests.

This change also does the following:
 - Refactors much of the k8s client library to support updates of
   previously-created k8s resources.
 - Fixes a unit test bug where we were not properly checking that
   the right cluster states were being returned for work by the
   supervisor.
 - Cleans up logging in various places.

Ticket: https://mattermost.atlassian.net/browse/MM-16413

### DIFF
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -20,6 +20,7 @@ func initCluster(apiRouter *mux.Router, context *Context) {
 	clusterRouter := apiRouter.PathPrefix("/cluster/{cluster:[A-Za-z0-9]{26}}").Subrouter()
 	clusterRouter.Handle("", addContext(handleGetCluster)).Methods("GET")
 	clusterRouter.Handle("", addContext(handleRetryCreateCluster)).Methods("POST")
+	clusterRouter.Handle("/provision", addContext(handleProvisionCluster)).Methods("POST")
 	clusterRouter.Handle("/kubernetes/{version}", addContext(handleUpgradeCluster)).Methods("PUT")
 	clusterRouter.Handle("", addContext(handleDeleteCluster)).Methods("DELETE")
 }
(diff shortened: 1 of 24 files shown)


internal/api/cluster.go
```diff
@@ -130,6 +131,49 @@ func handleRetryCreateCluster(c *Context, w http.ResponseWriter, r *http.Request
...
+	case model.ClusterStateStable:
+	case model.ClusterStateProvisioningFailed:
+	case model.ClusterStateProvisioningRequested:
+	default:
+		c.Logger.Warnf("unable to provision cluster while in state %s", cluster.State)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if cluster.State != model.ClusterStateProvisioningRequested {
```
crspeller: We already checked this above.

internal/provisioner/kops_provisioner.go
```diff
@@ -208,33 +300,53 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
 
-	_, err = k8sClient.CreateNamespaces(namespaces)
+	// Remove all previously-installed operator namespaces and resources.
+	for _, namespace := range namespaces {
+		logger.Infof("Cleaning up namespace %s", namespace)
+		err = k8sClient.Clientset.CoreV1().Namespaces().Delete(namespace, &metav1.DeleteOptions{})
+		if k8sErrors.IsNotFound(err) {
+			logger.Infof("Namespace %s not found; skipping...", namespace)
+		} else if err != nil {
+			return errors.Wrapf(err, "failed to delete namespace %s", namespace)
```
crspeller: Does it make sense to fail to provision if we couldn't delete something?

internal/provisioner/kops_provisioner.go
```diff
@@ -605,6 +570,128 @@ func makeClusterInstallationName(clusterInstallation *model.ClusterInstallation)
...
+
+			time.Sleep(5 * time.Second)
+		}
+	}
+}
+
+// waitForHelmRunning is used to check when Helm is ready to install charts.
+func waitForHelmRunning(ctx context.Context, configPath string) error {
+	for {
+		cmd := exec.Command("helm", "ls", "--kubeconfig", configPath)
```
crspeller: This applies in a bunch of places, but these exec functions do have versions that accept contexts in case this gets stuck.

internal/api/cluster.go
gabrieljackson: Yeah, although this check feels a bit redundant, the purpose of it is to only update the provisioner database if the state is anything other than `ClusterStateProvisioningRequested`. It saves some code space by not duplicating the code under each case statement.

internal/provisioner/kops_provisioner.go
gabrieljackson: In this case it does. We are trying to remove the namespace to entirely cleanup the old operator resources. If this fails, we will have a problem when we immediately try to recreate a namespace with the same name a few lines down.

internal/provisioner/kops_provisioner.go
gabrieljackson: Yeah, the exec functions are a less-then-ideal stopgap we put in to get helm up and running. We plan to remove helm and all of this code later. Also, the helm/exec code isn't actually new in this PR, I just moved some functions to a better spot.

We will definitely improve this in the future.

internal/provisioner/kops_provisioner.go
```diff
@@ -605,6 +570,128 @@ func makeClusterInstallationName(clusterInstallation *model.ClusterInstallation)
 	return fmt.Sprintf("mm-%s", clusterInstallation.Namespace[0:4])
 }
 
+// waitForHelmRunning is used to check when Helm is ready to install charts.
```
jwilander: Copy/paste error on this comment?

PR #40, opened 2019-06-21
Always pull operator images

gabrieljackson:
This ensures we don't get stale operator images on long-lived clusters.


operator-manifests/mysql/mysql-operator.yaml
```diff
@@ -231,8 +231,8 @@ spec:
       serviceAccountName: mysql-operator
       containers:
       - name: mysql-operator-controller
-        imagePullPolicy: IfNotPresent
         image: iad.ocir.io/oracle/mysql-operator:0.3.0
+        imagePullPolicy: Always
```
cpanato: if is set with a version tag we dont need to pull always, only apply for `latest` tags

PR #46, opened 2019-06-27
update minio spec

cpanato:


cpanato:
merging this because we cannot create new servers.

PR #49, opened 2019-07-02
fix comments for api endpoints #2

cpanato:
second part of the fix comments
and also added the missing json header for some endpoint that i forgot

### DIFF
--- a/internal/api/group.go
+++ b/internal/api/group.go
@@ -52,7 +52,7 @@ func handleGetGroups(c *Context, w http.ResponseWriter, r *http.Request) {
 	outputJSON(c, w, groups)
 }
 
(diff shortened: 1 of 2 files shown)


internal/api/group.go
```diff
@@ -52,7 +52,7 @@ func handleGetGroups(c *Context, w http.ResponseWriter, r *http.Request) {
 	outputJSON(c, w, groups)
 }
 
-// handleCreateGroup responds to POST /api/groups, beginning the process of creating a new group.
+// handleCreateGroup responds to POST /api/group, beginning the process of creating a new group.
```
gabrieljackson: The original comment was actually correct here.

cpanato: done

PR #53, opened 2019-07-11
Allow specifying size when creating installations

jwilander:
Tested locally with SQLite and PostgreSQL

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -13,6 +13,7 @@ func init() {
 	installationCreateCmd.Flags().String("owner", "", "An opaque identifier describing the owner of the installation.")
 	installationCreateCmd.Flags().String("version", "stable", "The Mattermost version to install.")
 	installationCreateCmd.Flags().String("dns", "", "The URL at which the Mattermost server will be available.")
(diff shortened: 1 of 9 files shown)


cmd/cloud/installation.go
```diff
@@ -13,6 +13,7 @@ func init() {
 	installationCreateCmd.Flags().String("owner", "", "An opaque identifier describing the owner of the installation.")
 	installationCreateCmd.Flags().String("version", "stable", "The Mattermost version to install.")
 	installationCreateCmd.Flags().String("dns", "", "The URL at which the Mattermost server will be available.")
+	installationCreateCmd.Flags().String("size", "", "The size of the installation. Accepts 100users, 1000users, 5000users, 10000users or 25000users. Defaults to 100users.")
```
gabrieljackson: nit: it may be a bit nicer to set the default value in the field for it in the flag.

0/5

jwilander: Oh yeah I should definitely do that

jwilander: Done. I left the default setting in the model as well to cover requests not coming from the CLI

PR #55, opened 2019-07-19
use the tagged version

cpanato:
lets use the tagged version for the provisioner since we have one now :)


cpanato:
after this we can tag the provisioner

PR #58, opened 2019-08-04
[MM-15371] Add Mattermost config get/set functionality

gabrieljackson:
This initial implementation is done by executing the Mattermost
CLI in one of the running containers of the cluster installation.

In the future, we should move to a direct API approach and should
use Kubernetes jobs. That said, this change lays the groundwork
for running arbitrary commands in the Mattermost pods.

https://mattermost.atlassian.net/browse/MM-15371


internal/api/cluster_installation.go
```diff
@@ -122,6 +141,68 @@ func handleSetClusterInstallationConfig(c *Context, w http.ResponseWriter, r *ht
...
+
+	var applyConfig func(parentKey string, value map[string]interface{}) error
+
+	// applyConfig takes the decomposed configuration, walks the resulting map, and invokes
+	// something akin to:
+	//	mattermost config set <ParentKey1.ParentKey2.LeafKey> <value>
+	//
+	// Ideally, this would be replaced by simply using the API and passing in the config struct
+	// directly, but at the moment that requires authentication.
+	applyConfig = func(parentKey string, parentValue map[string]interface{}) error {
```
jwilander: A little annoying that we have to do this, I thought we had a CLI for setting the entire config but it doesn't seem we do. Maybe a future improvement to the CLI.

PR #61, opened 2019-08-05
Update kops to 1.13.0

gabrieljackson:
This allows us to deploy kubernetes 1.13.X clusters!


cpanato:
i was about to do that 😂

cpanato:
@gabrieljackson can we use the cluster provision command to upgrade the k8s cluster?

gabrieljackson:
The provision command is for updating the operators and other kubernetes objects. This will require running `cloud cluster upgrade` to specifically handle upgrading k8s.

A note here: we haven't been able to test this much so far so we should perform some tests on throwaway clusters first.

cpanato:
will do after we merge this :)

PR #66, opened 2019-08-12
[MM-17701] Handle k8s nodes that are not schedulable

gabrieljackson:
In addition to ensuring that a cluster has enough resources, each
node should be checked for its ability to schedule pods. This
change adds these checks when performing cluster resource
analysis.

https://mattermost.atlassian.net/browse/MM-17701


jwilander:
@cpanato can you deploy?

cpanato:
@jwilander deployed on test

PR #76, opened 2019-08-19
mysql is a statefulSet now, adding a check here

cpanato:
the mysql operator is a statefullSet now and the cloud is checking for a deployment


internal/tools/k8s/helpers.go
```diff
@@ -49,6 +49,19 @@ func (kc *KubeClient) GetPodsFromDeployment(namespace, deploymentName string) (*
 	return kc.Clientset.CoreV1().Pods(namespace).List(listOptions)
 }
 
+// GetPodsFromStatefulset gets the pods that belong to a given deployment.
```
gabrieljackson:
```suggestion
// GetPodsFromStatefulset gets the pods that belong to a given stateful set.
```

PR #77, opened 2019-08-19
increase the time to wait for operators to start

cpanato:
sometimes the creation/start the pods can take much more time.
just saw for mysql took around 4-5 minutes. adding a bit more time


internal/provisioner/kops_provisioner.go
```diff
@@ -392,6 +395,7 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster) err
 		}
 
 		for _, pod := range pods.Items {
+
```
gabrieljackson: nit: extra space

PR #78, opened 2019-08-30
Update helm charts

cpanato:


helm-charts/prometheus_values.yaml
```diff
@@ -1,5 +1,14 @@
...
+  alertmanager:
+    create: false
+
 server:
   ingress:
     enabled: true
     annotations:
       kubernetes.io/ingress.class: private-nginx
+    hosts:
+      - kofw9hx1qbrq9dj8nuuwiyug6h.prometheus.internal.test.cloud.mattermost.com
```
gabrieljackson: Should this be hard-coded for every k8s cluster?

stylianosrigas: Nope this can be removed. Its automatically added by the Provisioner.

cpanato: forgot
will update :(
thanks for the catch

cpanato: done @gabrieljackson @stylianosrigas

PR #85, opened 2019-10-01
[MM-17574] Add multi-master kubernetes cluster options

gabrieljackson:
 - Adds multi-master sizing options when deploying a kubernetes
   cluster. All previous cluster sizes can now be made with multiple
   master nodes.
 - Added new cluster size option for 5000 users.
 - Changed node types for existing cluster sizes to be more
   consistent.
 - Increased cluster create and upgrade timeout default to 1000
   seconds.

https://mattermost.atlassian.net/browse/MM-17574


cmd/cloud/cluster.go
```diff
@@ -15,16 +15,16 @@ func init() {
 	clusterCmd.PersistentFlags().String("server", "http://localhost:8075", "The provisioning server whose API will be queried.")
 
 	clusterCreateCmd.Flags().String("provider", "aws", "Cloud provider hosting the cluster.")
-	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster.")
+	clusterCreateCmd.Flags().String("size", "SizeAlef500", "The size constant describing the cluster. Add '-HA2' or '-HA3' to the size for multiple master nodes.")
```
jwilander: Interesting choice to append it to the size

gabrieljackson: Do you think this is a bad method @jwilander? I don't see any obvious issues here, and it seems pretty flexible, but I am open to concerns.

jwilander: FYI Gabe and I chatted about it on a call, I'm OK with it

PR #86, opened 2019-10-02
Update kops to version 1.13.2

gabrieljackson:
Version 1.14.0 just came out, but we should add major k8s upgrades
to this server before upgrading kops to that version.


cpanato:
We need to pair when we merge this and run the upgrade in the test env. Sounds good?

gabrieljackson:
Hey @cpanato, just checking in on this. This change should be backwards and forwards compatible so we should be able to update our provisioning server deployments whenever we feel like it.

cpanato:
@gabrieljackson i will deploy this tomorrow first thing

PR #88, opened 2019-10-11
Add option for AWS S3 bucket installation filestores

gabrieljackson:
This change includes many enhancements to allow the cloud server
to provision and manage the AWS resources necessary for a secure
S3 bucket installation filestore. To do this, the AWS IAM, S3,
and Secrets Manager services are used.

Primary changes:
 - Change database schema to store filestore type for installations.
 - Use IAM service to create users, policies, and access keys for
   each installation requiring an S3 filestore.
 - Use S3 for the filestore bucket.
 - Use Secrets Manager to store IAM access keys in a secure way.
 - Provide a flag for server startup that will optionally allow
   S3 filestore teardown to preserve the S3 bucket data as an
   extra precaution.
 - All generic filestore operations have been moved to interfaces.

Other changes:
 - Change database schema to store database type for installations.
   This lays the foundation for AWS RDS installation databases.
 - New installation states for pre-provisioning and cleanup. This
   allows for better state handling when an error is encountered.
 - Cluster Installation creation failure will now retry in most
   cases when a failure is encountered.
 - Bug fix for setting the wrong deletion failure state in one
   case.

https://mattermost.atlassian.net/browse/MM-19081


cpanato:
will take some time to read :)

cmd/cloud/server.go
```diff
@@ -40,6 +40,7 @@ func init() {
 	serverCmd.PersistentFlags().String("public-subnets", "", "The public subnet IDs to use on AWS.")
 	serverCmd.PersistentFlags().Int("poll", 30, "The interval in seconds to poll for background work.")
 	serverCmd.PersistentFlags().Int("cluster-resource-threshold", 80, "The percent threshold where new installations won't be scheduled on a multi-tenant cluster")
+	serverCmd.PersistentFlags().Bool("keep-filestore-data", false, "Whether to preserve filestore data after installation deletion or not")
```
jwilander: Should we default to true for safety?

internal/tools/aws/aim.go
```diff
@@ -0,0 +1,248 @@
+package aws
```
jwilander: Typo in the filename? "aim.go" vs "iam.go"?

model/installation_filestore.go
```diff
@@ -0,0 +1,79 @@
...
+
+	mmv1alpha1 "github.com/mattermost/mattermost-operator/pkg/apis/mattermost/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// InstallationFilestoreMinioOperator is a filestore hosted in kubernetes via the operator.
+	InstallationFilestoreMinioOperator = "minio-operator"
+	// InstallationFilestoreAwsS3 is a filestore hosted via Amazon S3.
+	InstallationFilestoreAwsS3 = "aws-s3"
```
jwilander:
```suggestion
	InstallationFilestoreAWSS3 = "aws-s3"
```

Should we capitalize AWS?

gabrieljackson: I did initially, but found it a bit hard to read so wasn't sure. I am fine with either version.

internal/tools/aws/aim.go
gabrieljackson: Yep, thanks!

model/installation_filestore.go
jwilander: 0/5 just saw us capitalizing other acronyms in other places

cmd/cloud/server.go
gabrieljackson: Yeah, I wasn't sure if we wanted to err on the side of cleaning up by default or not and having to do it manually. Either one is fine with me.

jwilander: I'd lean towards not cleaning up by default

cpanato:
thanks, this is very cool

PR #90, opened 2019-10-18
Use installation ID as namespace

jwilander:
Turned out to be really simple.


internal/supervisor/installation.go
```diff
@@ -329,7 +329,7 @@ func (s *InstallationSupervisor) createClusterInstallation(cluster *model.Cluste
 	clusterInstallation := &model.ClusterInstallation{
 		ClusterID:      cluster.ID,
 		InstallationID: installation.ID,
-		Namespace:      model.NewID(),
+		Namespace:      installation.ID,
```
cpanato: :tada:

cpanato: this is just for new installations, right? if run an update in an existing installation it will still work?

jwilander: Yes just new installations, old ones will work fine

PR #92, opened 2019-10-25
Add option for AWS RDS installation databases

gabrieljackson:
This change includes many enhancements to allow the cloud server
to provision and manage the AWS resources necessary for a secure
RDS database. The changes include:
 - Add RDS functionality to the AWS package.
 - Use Secrets Manager to store RDS connection information in a
   secure way.
 - Provide a flag for server startup to optionally allow for RDS
   database teardown to preserve the RDS database data.
 - Provide a new method for AWS resource lookup. Previously, all
   manually-created AWS resources that are required when creating
   clusters and installations were passed in as server startup
   flags. New AWS logic is used for RDS management that instead
   uses structured AWS tags to find the correct resources.

Todo:
 - Add more tests

Note:
Without https://github.com/mattermost/mattermost-operator/pull/93, RDS
databases work, but the cluster installation reports `stable` before it actually
is.

https://mattermost.atlassian.net/browse/MM-19080

### DIFF
--- a/internal/tools/aws/rds.go
+++ b/internal/tools/aws/rds.go
@@ -1,0 +1,193 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+func (a *Client) rdsGetDBSecurityGroupIDs(logger log.FieldLogger) ([]string, error) {
+	svc := ec2.New(session.New(), &aws.Config{
+		Region: aws.String(DefaultAWSRegion),
+	})
+
+	input := &ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name: aws.String(DefaultDBSecurityGroupTagKey),
+				Values: []*string{
+					aws.String(DefaultDBSecurityGroupTagValue),
+				},
+			},
+		},
+	}
+
+	result, err := svc.DescribeSecurityGroups(input)
+	if err != nil {
+		return []string{}, err
+	}
+
+	var dbSecurityGroups []string
+	for _, sg := range result.SecurityGroups {
+		dbSecurityGroups = append(dbSecurityGroups, *sg.GroupId)
+	}
+
+	if len(dbSecurityGroups) == 0 {
+		return []string{}, fmt.Errorf("unable to find security groups tagged for Mattermost DB usage: %s=%s", DefaultDBSecurityGroupTagKey, DefaultDBSecurityGroupTagValue)
+	}
+
+	logger.WithField("security-group-ids", dbSecurityGroups).Debugf("Found %d DB tagged security groups", len(dbSecurityGroups))
+
+	return dbSecurityGroups, nil
+}
+
+func (a *Client) rdsEnsureDBClusterCreated(awsID, username, password string, logger log.FieldLogger) error {
+	svc := rds.New(session.New(), &aws.Config{
+		Region: aws.String(DefaultAWSRegion),
+	})
+
+	_, err := svc.DescribeDBClusters(&rds.DescribeDBClustersInput{
+		DBClusterIdentifier: aws.String(awsID),
+	})
+	if err == nil {
+		logger.WithField("db-cluster-name", awsID).Debug("AWS DB cluster already created")
+
+		return nil
+	}
+
+	dbSecurityGroupIDs, err := a.rdsGetDBSecurityGroupIDs(logger)
+	if err != nil {
+		return err
+	}
+
+	input := &rds.CreateDBClusterInput{
+		AvailabilityZones: []*string{
+			aws.String("us-east-1a"),
+			aws.String("us-east-1b"),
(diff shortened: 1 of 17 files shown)


model/installation_database.go
```diff
@@ -1,12 +1,77 @@
...
+	return nil, nil, nil
+}
+
+// GetDatabase returns the Database interface that matches the installation.
+func (i *Installation) GetDatabase() Database {
+	switch i.Database {
+	case InstallationDatabaseMysqlOperator:
+		return NewMysqlOperatorDatabase()
+	case InstallationDatabaseAwsRDS:
+		return aws.NewRDSDatabase(i.ID)
```
gigawhitlocks: Is it possible for `aws.NewRDSDatabase` to return `nil` here? Just keeping an eye out for potential panics.

gabrieljackson: Right now, it wouldn't be possible to git a `nil` value here:

```
// NewRDSDatabase returns a new RDSDatabase interface.
func NewRDSDatabase(installationID string) *RDSDatabase {
	return &RDSDatabase{
		installationID: installationID,
	}
}
```

That said, you are right that if we change the code in the future it could possibly be nil. I will think about this and we can do a mini-refactor later.

internal/tools/aws/rds.go
```diff
@@ -0,0 +1,193 @@
...
+
+	input := &rds.CreateDBClusterInput{
+		AvailabilityZones: []*string{
+			aws.String("us-east-1a"),
+			aws.String("us-east-1b"),
+			aws.String("us-east-1c"),
+		},
+		BackupRetentionPeriod: aws.Int64(7),
+		DBClusterIdentifier:   aws.String(awsID),
+		DatabaseName:          aws.String("mattermost"),
```
gigawhitlocks: Nit: is there any reason we might not want to hardcode these values?

internal/tools/aws/rds.go
```diff
@@ -0,0 +1,193 @@
...
+		EngineVersion:         aws.String("5.7"),
+		MasterUserPassword:    aws.String(password),
+		MasterUsername:        aws.String(username),
+		Port:                  aws.Int64(3306),
+		StorageEncrypted:      aws.Bool(false),
+		DBSubnetGroupName:     aws.String(DefaultDBSubnetGroupName),
+		VpcSecurityGroupIds:   aws.StringSlice(dbSecurityGroupIDs),
+	}
+
+	_, err = svc.CreateDBCluster(input)
```
gigawhitlocks: nit, this could be wrapped to one line e.g. `if _, err = svc.CreateDBCluster(input); err != nil {` since we are not using the returned value and are just checking the error here.

internal/tools/aws/rds.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+
+	logger.WithField("db-instance-name", instanceName).Debug("AWS DB instance created")
+
+	return nil
+}
+
+func rdsGetDBCluster(awsID string, logger log.FieldLogger) (*rds.DBCluster, error) {
+	svc := rds.New(session.New(), &aws.Config{
```
gigawhitlocks: I'm seeing this pattern in a number of places, maybe these three lines should be in a helper function?

internal/tools/aws/rds.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		return err
+	}
+
+	if len(result.DBClusters) != 1 {
+		return fmt.Errorf("expected 1 DB cluster, but got %d", len(result.DBClusters))
+	}
+
+	for _, instance := range result.DBClusters[0].DBClusterMembers {
+		_, err = svc.DeleteDBInstance(&rds.DeleteDBInstanceInput{
```
gigawhitlocks: Hmm, now I'm noticing that there are a number of places, like this, where we don't use the return value but also are not using the `if _, err = thing(); err != nil {` pattern -- is this a style choice that I'm ignorant of?

internal/tools/aws/rds.go
gabrieljackson: Yep, we are very-likely to want to make some of these values configurable in the future. Due to this already being a large change I opted to hardcode things that made sense. When we require more options in the future we will definitely add them.

internal/tools/aws/rds.go
gabrieljackson: Yeah, this is a style choice for the cloud repo. I personally prefer it, but other initial contributors to the repo also used the same style.

internal/tools/aws/rds.go
gabrieljackson: Noted below that this is a style choice, but you are right that we cloud do this.

internal/tools/aws/rds.go
gabrieljackson: So, the client created here is often slightly different in that it is specific to the AWS service being used. Note that it's `rds` here. That said, we could make a helper function for each AWS service. Do you think that warrants the extra code?

internal/tools/aws/rds.go
gigawhitlocks: Yeah sorry about that, at first I thought it was an oversight but then I kept seeing the pattern and realized it was probably intentional

internal/tools/aws/rds.go
gigawhitlocks: Sounds good, I'm more than happy to adhere to project style.

internal/tools/aws/rds.go
gigawhitlocks: Yup that's sort of what I suspected :+1:

gigawhitlocks:
LGTM

gabrieljackson:
Thanks for the review @gigawhitlocks!

PR #96, opened 2019-10-29
Refactor cluster and installation states

gabrieljackson:
 - Moves state definition and handling for clusters and installations
   into the model package.
 - Provides new tooling for exporting state data of the current model
   for clusters and installations.
 - Adds two new subcommands for viewing this state data.

Note: no state decision-making was altered in this PR, although changes
that could be made are now easier to understand via the state reports.

https://mattermost.atlassian.net/browse/MM-19789

Example Report:
```
cloud cluster state-report
[
    {
        "RequestedState": "creation-requested",
        "ValidStates": [
            "creation-requested",
            "creation-failed"
        ],
        "InvalidStates": [
            "stable",
            "provisioning-requested",
            "provisioning-failed",
            "upgrade-requested",
            "upgrade-failed",
            "deletion-requested",
            "deletion-failed",
            "deleted"
        ]
    },
    {
        "RequestedState": "provisioning-requested",
        "ValidStates": [
            "stable",
            "provisioning-requested",
            "provisioning-failed"
        ],
        "InvalidStates": [
            "creation-requested",
            "creation-failed",
            "upgrade-requested",
            "upgrade-failed",
            "deletion-requested",
            "deletion-failed",
            "deleted"
        ]
    },
    {
        "RequestedState": "upgrade-requested",
        "ValidStates": [
            "stable",
            "upgrade-requested",
            "upgrade-failed"
        ],
        "InvalidStates": [
            "creation-requested",
            "creation-failed",
            "provisioning-requested",
            "provisioning-failed",
            "deletion-requested",
            "deletion-failed",
            "deleted"
        ]
    },
    {
        "RequestedState": "deletion-requested",
        "ValidStates": [
            "stable",
            "creation-requested",
            "creation-failed",
            "provisioning-failed",
            "upgrade-requested",
            "upgrade-failed",
            "deletion-requested",
            "deletion-failed"
        ],
        "InvalidStates": [
            "provisioning-requested",
            "deleted"
        ]
    }
]
```

### DIFF
--- a/internal/api/installation.go
+++ b/internal/api/installation.go
@@ -126,24 +126,23 @@ func handleRetryCreateInstallation(c *Context, w http.ResponseWriter, r *http.Re
 	}
 	defer unlockOnce()
 
-	switch installation.State {
-	case model.InstallationStateCreationRequested:
-	case model.InstallationStateCreationFailed:
-	default:
+	newState := model.InstallationStateCreationRequested
+
+	if !installation.ValidTransitionState(newState) {
 		c.Logger.Warnf("unable to retry installation creation while in state %s", installation.State)
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	if installation.State != model.InstallationStateCreationRequested {
+	if installation.State != newState {
 		webhookPayload := &model.WebhookPayload{
 			Type:      model.TypeInstallation,
 			ID:        installation.ID,
-			NewState:  model.InstallationStateCreationRequested,
+			NewState:  newState,
 			OldState:  installation.State,
 			Timestamp: time.Now().UnixNano(),
 		}
-		installation.State = model.InstallationStateCreationRequested
+		installation.State = newState
 
 		err := c.Store.UpdateInstallation(installation)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


internal/api/installation.go
```diff
@@ -341,36 +337,23 @@ func handleDeleteInstallation(c *Context, w http.ResponseWriter, r *http.Request
...
-	case model.InstallationStateUpgradeInProgress:
-	case model.InstallationStateUpgradeFailed:
-	case model.InstallationStateDeletionRequested:
-	case model.InstallationStateDeletionInProgress:
-	case model.InstallationStateDeletionFinalCleanup:
-	case model.InstallationStateDeletionFailed:
-	default:
+	newState := model.InstallationStateDeletionRequested
+
+	if !installation.ValidTransitionState(newState) {
```
jwilander: Much cleaner 👍

model/cluster_states.go
```diff
@@ -0,0 +1,173 @@
...
+	ClusterStateDeletionFailed,
+	ClusterStateDeleted,
+}
+
+// AllCusterStatesPendingWork is a list of all cluster states that the supervisor
+// will attempt to transition towards stable on the next "tick".
+// Warning:
+// When creating a new cluster state, it must be added to this list if the cloud
+// cluster supervisor should perform some action on its next work cycle.
+var AllCusterStatesPendingWork = []string{
```
jwilander: s/AllCusterStatesPendingWork/AllClusterStatesPendingWork/

model/cluster_states.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// ClusterStateReport is a report of all cluster requests states.
+type ClusterStateReport []StateReportEntry
+
+// StateReportEntry is a report entry of a given request state.
+type StateReportEntry struct {
+	RequestedState string
+	ValidStates    StateList
+	InvalidStates  StateList
```
jwilander: What about going with `CurrentState`, `ValidNextStates`, `InvalidNextStates`? I think that might make it clearer what's being represented here

gabrieljackson: So, I do like this idea, but this could be a bit tricky in that the current state of an object could be something that the supervisor is currently working on and there could be many more states before the object is stable again. This pathway is something even the refactor doesn't capture and still is managed in the supervisor package.

Right now, the Report will tell you what states the object can/can't be in in regards to making a successful API request, which we already enforced. Perhaps I can think through the "supervisor pathway" question and do a follow-up PR sometime?

Thoughts?

jwilander: Ok to confirm, the report data is telling me if I were to request a state for an object, the `ValidStates` are the states the object must be in for me to be able to make that request. If the object is instead in a `InvalidStates` then my API request would fail.

An example:

I want to retry creation of a cluster through the API. I can only do that if my cluster is currently in the `creation-requested` or `creation-failed` states.

Does that sound right?

As for the supervisor pathway, I do think making that clearer in the future would be helpful but not entirely sure if moving that data/logic out of the supervisor makes sense. Maybe to expose it in the API like the state data here I guess

gabrieljackson: Yep, that is correct!

And I agree on the supervisor stuff.

gigawhitlocks:
This LGTM as well, obviously with the caveat that I'm still getting familiar with the codebase. I really like that you broke those state objects out into their own files, those are really, really clean looking :+1: Good stuff.

PR #99, opened 2019-11-01
Additional Helm logging output and fixes to README for getting started

gigawhitlocks:
I found that the README was missing instructions for setting up an AWS profile to specify a region, and that Helm version 2.15+ doesn't appear to work currently with the rest of our setup (fails to provision the nginx Ingress).

I have added instructions to the README that specify a known-working Helm version, adds a step to specify the correct region before starting, and in a second commit, I have added some additional log messages that output what Helm command is being executed before it is run.

The second commit in particular is a little verbose -- I did not make an effort to make this extra logging code any more succinct. Let me know if y'all think we should not include this additional logging code (it's all in one commit so I can easily back it out), or if y'all think I should spend some time to make it less verbose, I can do that as well.


internal/provisioner/kops_utils.go
```diff
@@ -113,14 +113,28 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+
+		err := cmd.Run()
 		if err != nil {
 			return err
 		}
 	} else {
-		err := exec.Command("helm", "install", "--kubeconfig", configPath, "-f", chart.valuesPath, chart.chartName, "--namespace", chart.namespace, "--name", chart.chartDeploymentName).Run()
+		cmd := exec.Command("helm", "--debug", "install", "--kubeconfig", configPath, "-f", chart.valuesPath, chart.chartName, "--namespace", chart.namespace, "--name", chart.chartDeploymentName)
+
+		logger.WithFields(log.Fields{
```
gabrieljackson: Instead of defining and executing the `cmd`s within the if loop, we could run the command outside of the loop to streamline the code a bit.

gigawhitlocks: Yeah, I see what you're saying. I've cleaned this up a bit.

PR #100, opened 2019-11-06
Improve cluster version handling

gabrieljackson:
 - Adds new version column for clusters that stores the kubernetes
   version running on the cluster. This version value is updated
   when a cluster is created, provisioned, or upgraded.
 - Allows new clusters to have a specified version when created and
   for cluster upgrades to target a specific version instead of
   latest stable.
 - Exposes the `kops set cluster` command for updating cluster
   configuration.
 - Exposes a provisioner method for kops to get the current running
   kubernetes version at any time.
 - Removed unused `wait` flag for create and upgrade commands.

https://mattermost.atlassian.net/browse/MM-19742

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -22,6 +22,10 @@ import (
 	"github.com/mattermost/mattermost-cloud/model"
 )
 
+// DefaultKubernetesVersion is the default value for a kubernetes cluster
+// version value.
+const DefaultKubernetesVersion = "0.0.0"
+
 // KopsProvisioner provisions clusters using kops+terraform.
 type KopsProvisioner struct {
 	clusterRootDir    string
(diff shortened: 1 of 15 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -22,6 +22,10 @@ import (
 	"github.com/mattermost/mattermost-cloud/model"
 )
 
+// DefaultKubernetesVersion is the default value for a kubernetes cluster
+// version value.
+const DefaultKubernetesVersion = "0.0.0"
```
cpanato: what means 0.0.0 which version will be that? sounds confusing to me

gabrieljackson: Whenever you see `0.0.0` here, it is just the version we store in the database for the kubernetes version until we know what it really is. For example, a newly created cluster will have this value until it is created, then the database will be updated to the actual running version.

cpanato: thanks!

PR #103, opened 2019-11-11
Update Go and Alpine

gabrieljackson:
Changes:
 - Update Go from 1.12 to 1.13
 - Update Alpine from 3.9 to 3.10

I ran into an issue with Go version 1.12 and one of the k8s modules so I decided to take that as our cue to upgrade:

```
verifying k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/go.mod: checksum mismatch
	downloaded: h1:nEP/6rwhzfljWYGVS6pfyES3ipZTR19vzMnSM+ur3ho=
	go.sum:     h1:3jediapYqJ2w1BFw7lAZPCx7scubsTfosqHkhXCWJKw=
```

https://github.com/kubernetes-sigs/kubebuilder/issues/1169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,25 @@
-GO ?= $(shell command -v go 2> /dev/null)
-MATTERMOST_CLOUD_IMAGE ?= mattermost/mattermost-cloud:test
-MACHINE = $(shell uname -m)
-DOCKER_BUILD_IMAGE = golang:1.12
-DOCKER_BASE_IMAGE = alpine:3.9
-GOFLAGS ?= $(GOFLAGS:)
-BUILD_TIME := $(shell date -u +%Y%m%d.%H%M%S)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,17 +1,26 @@
...
+################################################################################
+
+## Docker Build Versions
+DOCKER_BUILD_IMAGE = golang:1.13
+DOCKER_BASE_IMAGE = alpine:3.10
+
+## Tool Versions
+TERRAFORM_VERSION=0.11.14
+KOPS_VERSION=1.14.1
+HELM_VERSION=v2.14.2
```
cpanato: we can use helm 2.16.0

gabrieljackson: @gigawhitlocks actually ran into issues using newer helm versions. 2.14.2 was the latest version that worked without issues in his testing.

PR #105, opened 2019-11-13
MM-19561 Install Team Edition if no license is provided

gigawhitlocks:
Corollary to https://github.com/mattermost/mattermost-plugin-cloud/pull/24 , the provisioner needs to be updated to pull down the correct image (Team Edition) when no license is specified.

This change works, and I'd like to go ahead and propose this change, however it does occur to me that this overloads the `license` field a bit, and it might be better to include an edition field explicitly rather than to simply choose Team Edition if there is no license as I have done here. I'm more than open to feedback about that design decision.

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -94,6 +94,10 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 
 		mattermostInstallation.Spec.MattermostLicenseSecret = licenseSecretName
 		logger.Debug("Cluster installation configured with a Mattermost license")
+	} else {
+		mattermostInstallation.Spec.Image = "mattermost/mattermost-team-edition"
+		// TODO: is the mattermost team edition image name stored somewhere as a constant?
+		logger.Debug("No license provided; changing the image type to team edition")
 	}
 
 	databaseSpec, databaseSecret, err := installation.GetDatabase().GenerateDatabaseSpecAndSecret(logger)


jwilander:
If I later add a license will that work and change the image back to the enterprise one?

gigawhitlocks:
I don't believe so, because the change is only in `createClusterInstallation`. If we want that behavior, I believe it needs to be added as an action under the related `upgrade` action as well.

It occurs to me that I started to add support for this behavior in the plugin, so I'll either need to finish adding that behavior on the provisioned side or remove it from the other PR. Happy to do either.

jwilander:
Hmm right the provisioner doesn't support specifying the image in the upgrade request. I think it makes sense to update the provisioner to do that as a follow-up later on and we can take this as is. Maybe worth adding a note here to the help output letting people know that upgrading from team to e10 or e20 won't work

gabrieljackson:
@joram, am I remembering correctly that the enterprise edition container will behave exactly like team edition if no license is provided? I know spinwicks originally didn't have any licenses.

gigawhitlocks:
@gabrieljackson if the answer to the above is yes do you think it's better that I just close this PR and we actually run licenseless EE rather than TE?

gabrieljackson:
Yeah, that is exactly what I am thinking. It gets around the issue with converting from license to no license and visa versa.

gigawhitlocks:
OK sounds good. That simplifies the code a bit as well (and this was a good exercise for me seeing how the provisioner passes the spec off to the Operator for deployment anyway, so no time wasted) and I'll change some of the language in my other PR so that it's clearer that we're actually installing licenseless EE just in case for some reason down the line we want actual TE. Closing this PR as installing licenseless EE will require no change to the provisioner.

jwilander:
I'm good with licenseless EE 👍

PR #107, opened 2019-11-21
MM-20003 Fix Fluentd issues with new chart

stylianosrigas:
Add ability to:
- Add new Helm repos
- Update Helm repos everytime to get latest available charts
- Install new Fluentd chart.

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -213,6 +213,12 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
 		return err
 	}
 
+	logger.Info("Adding https://kiwigrid.github.io Helm repo.")
+	addHelmRepo("kiwigrid", "https://kiwigrid.github.io", logger)
+
+	logger.Info("Updating all Helm repos.")
+	helmRepoUpdate(logger)
+
 	prometheusDNS := fmt.Sprintf("%s.prometheus.%s", cluster.ID, provisioner.privateDNS)
 	elasticsearchDNS := fmt.Sprintf("elasticsearch.%s", provisioner.privateDNS)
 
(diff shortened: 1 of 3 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -213,6 +213,12 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
 		return err
 	}
 
+	logger.Info("Adding https://kiwigrid.github.io Helm repo.")
+	addHelmRepo("kiwigrid", "https://kiwigrid.github.io", logger)
```
gabrieljackson: We are ignoring the returned error here. We should check it.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -213,6 +213,12 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, aws aw
...
+
+	logger.Info("Updating all Helm repos.")
+	helmRepoUpdate(logger)
```
gabrieljackson: Same here.

internal/provisioner/kops_utils.go
```diff
@@ -112,6 +112,44 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+
+	cmd = exec.Command("helm repo add", arguments...)
+
+	logger.WithFields(log.Fields{
+		"cmd":  cmd.Path,
+		"args": cmd.Args,
+	}).Info("Invoking command")
+
+	err := cmd.Run()
+	if err != nil {
```
gabrieljackson: We could just return err here without checking it. It makes sense to keep the if check if you want to wrap the error with more info.

internal/provisioner/kops_utils.go
```diff
@@ -112,6 +112,44 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+func helmRepoUpdate(logger log.FieldLogger) error {
+	var cmd *exec.Cmd
+	cmd = exec.Command("helm repo update")
+
+	logger.WithFields(log.Fields{
+		"cmd": cmd.Path,
+	}).Info("Invoking command")
+
+	err := cmd.Run()
+	if err != nil {
```
gabrieljackson: Same here.

internal/provisioner/kops_provisioner_cluster.go
stylianosrigas: Thanks @gabrieljackson fixed :)

internal/provisioner/kops_utils.go
stylianosrigas: Fixed ;)

internal/provisioner/kops_utils.go
stylianosrigas: Fixed ;)

internal/provisioner/kops_provisioner_cluster.go
stylianosrigas: Fixed ;)

PR #108, opened 2019-11-21
[MM-20635] add custom kops aws ami image to the configs

cpanato:
This adds a new config setting to set when starting the server `--custom-kops-aws-ami-image`

and when we create new clusters it will use the custom image, if that is not set it will use the default from kops.

In another ticket we will implement the upgrade, but need to discuss that before the implementation.

NOTE:
My understanding is every time we would like to upgrade the cluster we need to update kops, so we also will need to update the image

jira: https://mattermost.atlassian.net/browse/MM-20635


gabrieljackson:
After thinking about this a bit more, I do think that we want set the AMI when we create each cluster as opposed to having it be server-wide. To do that we will need to store the AMI to be used in the database. We should probably store the AMI in the provisioner metadata for the cluster.

cpanato:
closing in favor of https://github.com/mattermost/mattermost-cloud/pull/113

PR #110, opened 2019-11-25
Add per-cluster VPC support

gabrieljackson:
This change adds functionality for finding AWS VPCs that are tagged
as available. These VPCs are used for new kops cluster creation and
are then tagged as in-use. When a cluster is deleted, the VPC is
released for future use.

https://mattermost.atlassian.net/browse/MM-19824

### DIFF
--- a/operator-manifests/mattermost/operator.yaml
+++ b/operator-manifests/mattermost/operator.yaml
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:v0.8.5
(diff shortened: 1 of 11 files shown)


operator-manifests/mattermost/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:v0.8.5
+          image: mattermost/mattermost-operator:0de3fa2
```
jwilander: On purpose?

gabrieljackson: Oh no, not on purpose. Thanks for catching that.

PR #111, opened 2019-11-27
MM-20734 Migrating from Fluentd to Fluent-bit solution.

stylianosrigas:
This PR:

- Migrates from using Fluentd to Fluentbit solution
- Removes the function for installing new Helm repos as all repos used at the moment are using stable repo.


helm-charts/fluent-bit_values.yaml
```diff
@@ -0,0 +1,272 @@
...
+trackOffsets: false
+
+## PriorityClassName
+## Ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
+priorityClassName: ""
+
+backend:
+  type: es
+  es:
+    host: "elasticsearch.example.com"
```
gigawhitlocks: Are these all dummy values or does this `host` field need to be updated?

stylianosrigas: @gigawhitlocks The host is the only thing that needs to be updated and its done by the provisioner code.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -248,11 +242,11 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
...
-			valuesPath:          "helm-charts/fluentd_values.yaml",
-			chartName:           "kiwigrid/fluentd-elasticsearch",
-			namespace:           "fluentd",
-			chartDeploymentName: "fluentd",
-			setArgument:         fmt.Sprintf("elasticsearch.host=%s", elasticsearchDNS),
+			valuesPath:          "helm-charts/fluent-bit_values.yaml",
+			chartName:           "stable/fluent-bit",
+			namespace:           "fluent-bit",
+			chartDeploymentName: "fluent-bit",
+			setArgument:         fmt.Sprintf("backend.es.host=%s", elasticsearchDNS),
```
stylianosrigas: @gigawhitlocks check here ;)

PR #112, opened 2019-11-27
MM-19772 Update cluster AWS instance types

gabrieljackson:
Clusters are now deployed with updated AWS instance types based
on resource metrics collected in initial deployments.

https://mattermost.atlassian.net/browse/MM-19772


jwilander:
Would it be worth going to `SizeBeta*` prefix? I know that's what Jesse originally intended. That way we can keep around and know about old cluster sizes while introducing new sizes

gabrieljackson:
I had figured this would be our last real chance to do a reset at an educated first guess on cluster sizes and then introduce new size names later. That said, I am 0/5 on doing them one way or another so I will change it if you think that is best.

PR #115, opened 2019-12-04
Upgrade kops to 1.15.0

gabrieljackson:
This change brings kops up to 1.15.0. A change was made in this
version which gives better upgrade validation handling, so the
manual upgrade timeout flags have been removed as they are no
longer needed.

https://mattermost.atlassian.net/browse/MM-20916


gabrieljackson:
We will need to build new 1.15.X AMIs for use with k8s 1.15.X.

gigawhitlocks:
LGTM

cpanato:
@gabrieljackson i will make the whole chain

PR #116, opened 2019-12-04
Better AWS EC2 tag logging

gabrieljackson:


internal/tools/aws/cluster_management.go
```diff
@@ -249,7 +249,6 @@ func (a *Client) releaseVpc(clusterID string, logger log.FieldLogger) error {
 	}
 
 	for _, subnet := range publicSubnets {
-		logger.Debugf("Untagging subnet %s", *subnet.SubnetId)
```
gigawhitlocks: Just curious if you meant to remove this line?

gabrieljackson: Yep! Redundant logs are redundant. ;)

gigawhitlocks: Perfect, I figured it was something like that

PR #117, opened 2019-12-06
Add server flag for default kops creation

gabrieljackson:
A new flag `use-existing-aws-resources` can be used to skip the
VPC find-and-claim behavior that is used for production cluster
creation. Instead, the default kops behavior of creating a new VPC
and resources will be used.

https://mattermost.atlassian.net/browse/MM-21053

### DIFF
--- a/internal/tools/aws/route53.go
+++ b/internal/tools/aws/route53.go
@@ -63,7 +63,10 @@ func (a *Client) CreateCNAME(dnsName string, dnsEndpoints []string, logger log.F
 		return err
 	}
 
-	logger.Debugf("AWS route53 response: %s", prettyRoute53Response(resp))
+	logger.WithFields(map[string]interface{}{
+		"route53-dns-value":     dnsName,
+		"route53-dns-endpoints": dnsEndpoints,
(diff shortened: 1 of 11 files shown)


internal/tools/aws/route53.go
```diff
@@ -123,7 +124,10 @@ func (a *Client) DeleteCNAME(dnsName string, logger log.FieldLogger) error {
 		return err
 	}
 
-	logger.Debugf("AWS route53 response: %s", prettyRoute53Response(resp))
+	logger.WithFields(map[string]interface{}{
```
gigawhitlocks: Just a thought cause I know you've typed `map[string]interace{}{` a few times in this PR and in previous ones.. maybe we should define a type in this package `loggerFields` e.g. `type loggerFields map[string]interface{}` just to save some typing in the future? It'd also look nice. 0/5 on this.

gabrieljackson: Yep, that is actually provided by the package so let me clean that up. I had actually meant to get around to fixing that for all the log lines.

PR #120, opened 2019-12-12
add new field - ExtraData

cpanato:
#### Summary
Adding a bit more info in the webhook to help when need to drill down

but the idea is to have the cluster ID when posting messages, like when we receive in the `cloud-test-installation-states` channel

Please let me know if is missing something else.

### DIFF
--- a/internal/supervisor/cluster_installation.go
+++ b/internal/supervisor/cluster_installation.go
@@ -111,6 +111,7 @@ func (s *ClusterInstallationSupervisor) Supervise(clusterInstallation *model.Clu
 	webhookPayload := &model.WebhookPayload{
 		Type:      model.TypeClusterInstallation,
 		ID:        clusterInstallation.ID,
+		ClusterID: clusterInstallation.ClusterID,
 		NewState:  newState,
 		OldState:  oldState,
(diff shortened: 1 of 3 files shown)


gabrieljackson:
I can see how having more information could be useful on these webhooks. That said, this would be the first webhook field that may or may not be in the webhook payload depending on what type it is. Because of this, I think it makes sense to add one extra field that can contain any additional metadata, but it will be assumed that there may not be any extra data at all on some webhooks. Here is what I am thinking:

```
type WebhookPayload struct {
	Timestamp int64  `json:"timestamp"`
	ID        string `json:"id"`
	Type      string `json:"type"`
	NewState  string `json:"new_state"`
	OldState  string `json:"old_state"`
	ExtraData map[string]string `json:"extra_data,omitempty"`
}
```

You would then be able to define as much extra data as you want and `ExtraData` would possibly be empty. This could be used by all types of webhooks for anything we feel like and would be much more flexible.

What do you think @cpanato?

cpanato:
@gabrieljackson like this idea, will change

cpanato:
@gabrieljackson PTAL, let me know if need to change anything else

PR #124, opened 2019-12-18
abort if dns name is too long

cpanato:
#### Summary
- abort the cluster installation creation if the dns is too long
- and skip the deletion of the route53 entry is too long, since this never should happen if we are blocking in the creating, but will fix some cases that we already have

### DIFF
--- a/internal/tools/aws/route53.go
+++ b/internal/tools/aws/route53.go
@@ -78,6 +78,11 @@ func (a *Client) DeleteCNAME(dnsName string, logger log.FieldLogger) error {
 		return err
 	}
 
+	if len(dnsName) >= 64 {
+		logger.Warnf("DNS name too long, skipping since it could never have been created record-name=%s", dnsName)
+		return nil
+	}
+
 	nextRecordName := dnsName
 	var recordsets []*route53.ResourceRecordSet
 	for {
--- a/internal/api/installation_test.go
+++ b/internal/api/installation_test.go
@@ -269,6 +269,16 @@ func TestCreateInstallation(t *testing.T) {
 		require.EqualError(t, err, "failed with status code 400")
 	})
 
+	t.Run("dns too long", func(t *testing.T) {
+		_, err := client.CreateInstallation(&model.CreateInstallationRequest{
+			OwnerID:  "owner",
+			Version:  "version",
+			DNS:      "xoxo-serverwithverylongnametoexposeissuesrelatedtolengthofkeystha",
+			Affinity: model.InstallationAffinityIsolated,
+		})
+		require.EqualError(t, err, "failed with status code 400")
+	})
+
 	t.Run("invalid dns", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


internal/tools/aws/route53.go
```diff
@@ -78,6 +78,11 @@ func (a *Client) DeleteCNAME(dnsName string, logger log.FieldLogger) error {
 		return err
 	}
 
+	if len(dnsName) >= 64 {
+		logger.Warnf("DNS name too long, skipping since it will never created record-name=%s", dnsName)
```
gigawhitlocks: Minor nitpick on the error wording, "skipping since it could never have been created"

cpanato: done thanks!

cmd/cloud/installation.go
```diff
@@ -68,6 +68,10 @@ var installationCreateCmd = &cobra.Command{
 		database, _ := command.Flags().GetString("database")
 		filestore, _ := command.Flags().GetString("filestore")
 
+		if len(dns) >= 64 {
```
gabrieljackson: We don't normally check stuff in the client and just let the API server validate, but this is fine! Just an FYI.

internal/api/installation.go
```diff
@@ -71,6 +71,12 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	if len(createInstallationRequest.DNS) >= 64 {
+		c.Logger.Warningf("DNS name too long DNS=%s", createInstallationRequest.DNS)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
```
gabrieljackson: Cloud we move this check to https://github.com/mattermost/mattermost-cloud/blob/master/model/installation_request.go#L46 ?

We already validate a bunch of stuff when the request is read in and this fits perfectly there.

Also, it would be nice if the error message was something like `DNS names must be less than 64 characters, but name was %d long. DNS=%s`

cmd/cloud/installation.go
cpanato: Made here, because if using the API, like me and other robots we can get right away and dont have a two step action.
i can remove if you want 0/5

internal/api/installation.go
cpanato: ahh cool, yes will move there, thanks for the review

cmd/cloud/installation.go
cpanato: nvm, thought was another file, but if you call using the cmd line it will validate?

gabrieljackson: Yeah, the CLI is just making the same API request which always validates.

cpanato: removed

internal/api/installation.go
cpanato: done

gabrieljackson:
Looks like tests are failing @cpanato

cpanato:
@gabrieljackson fixed

PR #125, opened 2019-12-20
MM-21091: look up hosted zone ids and certificate ARN in AWS when flags are not supplied

gsagula:
#### Summary
This PR makes the the following CLI flags optional:
--route53-id
--private-route53-id
--certificate-aws-arn

#### What happens:
When the flags above are not supplied, provisioner will attempt to fetch them from AWS using manually created tags. The provisioner will not continue if it fails to fetch the values for any of these flags.

Fixes: https://mattermost.atlassian.net/browse/MM-21091 


gabrieljackson:
This is a good start @gsagula! I think we may want to change a bit of the behavior here though.

Instead of only checking for the AWS resources for zones, certs, etc. once on server startup, I think it would be nice to have the tag lookup happen when those resources are needed. By doing so, we would be able to change our AWS accounts on the fly without having to restart every provisioner server to pick up the details. This is how the other tag lookup behavior works.

gsagula:
@gabrieljackson This PR is ready for another pass. Thanks!

internal/tools/aws/constants.go
```diff
@@ -56,6 +56,31 @@ const (
 	// Note: This needs to be manually created before RDS databases can be used.
 	DefaultDBSubnetGroupTagValue = "MYSQL/Aurora"
 
+	// DefaultInstallCertificatesTagKey is the default key used to find the server
+	// TLS certificate ARN.
+	// Note: This needs to be manually created before attempting to provision a cluster in VPC.
```
gabrieljackson: This `Note` should probably be removed. It will apply to all clusters.

I believe this applies to your other constants as well.

internal/tools/aws/route53.go
```diff
@@ -2,22 +2,51 @@ package aws
...
-// CreateCNAME creates an AWS route53 CNAME record.
-func (a *Client) CreateCNAME(dnsName string, dnsEndpoints []string, logger log.FieldLogger) error {
+// CreatePublicCNAME creates a record in Route53 for a public domain name.
+func (a *Client) CreatePublicCNAME(dnsName string, dnsEndpoints []string, logger log.FieldLogger) error {
+	id, err := a.getHostedZoneIDWithTag(Tag{
+		Key:   DefaultCloudDNSTagKey,
+		Value: DefaultPublicCloudDNSTagValue,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "creating a public CNAME")
```
gabrieljackson:
```suggestion
		return errors.Wrapf(err, "unable to create public CNAME")
```

internal/tools/aws/route53.go
```diff
@@ -2,22 +2,51 @@ package aws
...
+}
+
+// CreatePrivateCNAME creates a record in Route53 for a private domain name.
+func (a *Client) CreatePrivateCNAME(dnsName string, dnsEndpoints []string, logger log.FieldLogger) error {
+	id, err := a.getHostedZoneIDWithTag(Tag{
+		Key:   DefaultCloudDNSTagKey,
+		Value: DefaultPrivateCloudDNSTagValue,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "creating a private CNAME")
```
gabrieljackson:
```suggestion
		return errors.Wrapf(err, "unable to create private CNAME")
```

internal/tools/aws/route53.go
```diff
@@ -64,15 +93,41 @@ func (a *Client) CreateCNAME(dnsName string, dnsEndpoints []string, logger log.F
...
-// DeleteCNAME deletes an AWS route53 CNAME record.
-func (a *Client) DeleteCNAME(dnsName string, logger log.FieldLogger) error {
+// DeletePublicCNAME deletes a AWS route53 record for a public domain name.
+func (a *Client) DeletePublicCNAME(dnsName string, logger log.FieldLogger) error {
+	id, err := a.getHostedZoneIDWithTag(Tag{
+		Key:   DefaultCloudDNSTagKey,
+		Value: DefaultPublicCloudDNSTagValue,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "deleting a public CNAME: %s", dnsName)
```
gabrieljackson:
```suggestion
		return errors.Wrapf(err, "unable to delete public CNAME: %s", dnsName)
```

internal/tools/aws/route53.go
```diff
@@ -64,15 +93,41 @@ func (a *Client) CreateCNAME(dnsName string, dnsEndpoints []string, logger log.F
...
+}
+
+// DeletePrivateCNAME deletes a AWS route53 record for a private domain name.
+func (a *Client) DeletePrivateCNAME(dnsName string, logger log.FieldLogger) error {
+	id, err := a.getHostedZoneIDWithTag(Tag{
+		Key:   DefaultCloudDNSTagKey,
+		Value: DefaultPrivateCloudDNSTagValue,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "deleting a private CNAME: %s", dnsName)
```
gabrieljackson:
```suggestion
		return errors.Wrapf(err, "unable to delete private CNAME: %s", dnsName)
```

cmd/cloud/server.go
```diff
@@ -126,7 +117,7 @@ var serverCmd = &cobra.Command{
 			"debug":                           debug,
 		}).Info("Starting Mattermost Provisioning Server")
 
-		deprecationWarnings(logger)
+		deprecationWarnings(logger, command)
```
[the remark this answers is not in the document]
gsagula: Interesting, where are you seeing  this? A pointer to command needs to be passed to `deprecationWarnings` so that we can call `Flags()` and check for deprecated params.

gsagula:
@gabrieljackson Thanks for reviewing it 👍PTAL

cmd/cloud/server.go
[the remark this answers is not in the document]
gabrieljackson: Sorry, I read this signature too quickly. We are good to go.

jwilander:
@cpanato are all the tags set correctly in the test environment? Probably want to confirm that before we merge

gsagula:
@gabrieljackson @jwilander Thanks for the review. I forgot to update the readme.

gsagula:
Quick question: Is the default for `--state-store` still valid? I couldn't get it to work.
https://github.com/mattermost/mattermost-cloud/blob/f352111fc85f1ace85654b8734067f875ef49f09/cmd/cloud/server.go#L37

gabrieljackson:
There really isn't a good default for that as it can be any bucket name, so I think that default is still about as good as anything.

README.md
```diff
@@ -55,12 +55,10 @@ Before running the server the first time you must set up the DB with:
 $ cloud schema migrate
 ```
 
-To run the server you will need a certificate ARN, private and public Route53 IDs, and a private DNS from AWS. For staff developers you can get these in our dev AWS account.
-
 Run the server with:
 
 ```
-$ cloud server --state-store=<your-s3-bucket> --private-dns dev.cloud.internal.mattermost.com --private-route53-id=<private-route53-id> --route53-id=<public-route53-id> --certificate-aws-arn <certifcate-aws-arn>
+$ cloud server --state-store=<your-s3-bucket> --private-dns dev.cloud.internal.mattermost.com
```
gigawhitlocks: :heart_eyes:

cpanato:
will finish my review tomorrow and deploy as well.
please do not merge it

PR #130, opened 2020-01-06
Defining an interface for managing cluster utilities as a group

gigawhitlocks:
This is a large change, ~still a work in progress,~ but it's ready to receive feedback while I continue fleshing out the details. Sorry that it is so large! It got away from me, a bit.

What's changed:
1) Moves helm-related utilities to their own file, all in one place
2) Defines an interface for creating, deleting, and updating cluster utility services that is agnostic to how the utility is deployed. Today those are all created with Helm but it should be possible to change these services to use Operators without updating `kops_provisioner_cluster.go` now, and without affecting the services that use Helm. Additional services can also be added cleanly using this interface.
3) Breaks out each of the existing utility services (nginx, Prometheus, and Fluentbit) into their own files for organizational purposes, and redefines their actions so that they satisfy the new interface.
4) Calls `helm upgrade` on Prometheus when `cloud cluster provision` is called.

To do:
~I am still working on how to properly call `helm upgrade` during Provision so that calling `cloud cluster provision` will upgrade the services. Currently `UpgradeUtilityGroup` is defined and called but it doesn't do anything.~

To resolve [MM-20995](https://mattermost.atlassian.net/browse/MM-20995) it should be possible, after this change, to add just the API endpoint and call, from within it, `Upgrade` on the Prometheus singleton object created as a part of this change, once I have the `helm upgrade` call worked out.

Some other notes:
~Commits from the other PRs I opened before the holiday are still in the history. They are isolated to 2 files though and should easily be ignored (`Makefile` and `internal/supervisor/cluster_installation.go`). If we merge those PRs the changes should disappear from this diff after I rebase.~

Also, all of my work in progress history is in the commit history. I obviously intend to squash the history before merging :)

### DIFF
--- a/internal/provisioner/fluentbit.go
+++ b/internal/provisioner/fluentbit.go
@@ -1,0 +1,56 @@
+package provisioner
+
+import (
+	"fmt"
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	log "github.com/sirupsen/logrus"
+)
+
+type fluentbit struct {
+	provisioner *KopsProvisioner
+	kops        *kops.Cmd
+	logger      log.FieldLogger
+}
+
+func newFluentbitHandle(provisioner *KopsProvisioner, kops *kops.Cmd, logger log.FieldLogger) (*fluentbit, error) {
+	if logger == nil {
+		return nil, fmt.Errorf("cannot instantiate Fluentbit handle with nil logger")
+	}
+
+	if provisioner == nil {
+		return nil, fmt.Errorf("cannot create a connection to Fluentbit if the provisioner provided is nil")
+	}
+
+	if kops == nil {
+		return nil, fmt.Errorf("cannot create a connection to Fluentbit if the Kops command provided is nil")
+	}
+
+	return &fluentbit{
+		provisioner: provisioner,
+		kops:        kops,
+		logger:      logger.WithField("cluster-utility", "fluentbit"),
+	}, nil
+}
+
+func (f *fluentbit) Create() error {
+	elasticSearchDNS := fmt.Sprintf("elasticsearch.%s", f.provisioner.privateDNS)
+	return (&helmDeployment{
+		chartDeploymentName: "fluent-bit",
+		chartName:           "stable/fluent-bit",
+		namespace:           "fluent-bit",
+		setArgument:         fmt.Sprintf("backend.es.host=%s", elasticSearchDNS),
+		valuesPath:          "helm-charts/fluent-bit_values.yaml",
+		kopsProvisioner:     f.provisioner,
+		kops:                f.kops,
+		logger:              f.logger,
+	}).Create()
+}
+
+func (f *fluentbit) Destroy() error {
+	return nil
+}
+
+func (f *fluentbit) Upgrade() error {
+	return nil
+}
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -1,0 +1,234 @@
+package provisioner
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/k8s"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// helmDeployment deploys Helm charts.
+type helmDeployment struct {
+	chartDeploymentName string
+	chartName           string
+	namespace           string
+	setArgument         string
+	valuesPath          string
+
+	cluster         *model.Cluster
+	kopsProvisioner *KopsProvisioner
+	kops            *kops.Cmd
+	logger          log.FieldLogger
+}
+
+func installHelm(kops *kops.Cmd, logger log.FieldLogger) error {
+	logger.Info("Installing Helm")
+
+	err := helmSetup(logger, kops)
+	if err != nil {
+		return errors.Wrap(err, "couldn't install Helm")
+	}
+
+	wait := 120
+	logger.Infof("Waiting up to %d seconds for helm to become ready...", wait)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(wait)*time.Second)
+	defer cancel()
+	err = waitForHelmRunning(ctx, kops.GetKubeConfigPath())
+	if err != nil {
+		return errors.Wrap(err, "Helm didn't start as expected, or we couldn't detect it")
+	}
+
+	logger.Info("Updating all Helm repos.")
+	return helmRepoUpdate(logger)
+}
+
+func (d *helmDeployment) Update() error {
+	logger := d.logger.WithField("helm-update", d.chartName)
+
+	logger.Infof("Refreshing helm chart %s -- may trigger service upgrade", d.chartName)
+	err := upgradeHelmChart(*d, d.kops.GetKubeConfigPath(), logger)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("got an error trying to upgrade the helm chart %s", d.chartName))
+	}
+	return nil
+}
+
+// Create will install a given HelmDeployment in the cluster specified in that object.
+func (d *helmDeployment) Create() error {
+	logger := d.logger.WithField("helm-create", d.chartName)
+	logger.Infof("Installing helm chart %s", d.chartName)
+	return installHelmChart(*d, d.kops.GetKubeConfigPath(), logger)
+}
+
+// waitForHelmRunning is used to check when Helm is ready to install charts.
+func waitForHelmRunning(ctx context.Context, configPath string) error {
+	for {
+		cmd := exec.Command("helm", "ls", "--kubeconfig", configPath)
+		var out bytes.Buffer
+		cmd.Stderr = &out
+		cmd.Run()
+		if out.String() == "" {
+			return nil
+		}
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "timed out waiting for helm to become ready")
+		case <-time.After(5 * time.Second):
+		}
+	}
+}
+
+// helmRepoUpdate updates the helm repos to get latest available charts
+func helmRepoUpdate(logger log.FieldLogger) error {
+	var cmd *exec.Cmd
+	arguments := []string{
+		"repo",
+		"update",
+	}
+
+	cmd = exec.Command("helm", arguments...)
+
+	logger.WithFields(log.Fields{
+		"cmd": cmd.Path,
+	}).Info("Invoking command")
+
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrap(err, "Failed to execute Helm to update the Helm repos")
+	}
+	return nil
+}
+
+// installHelmChart is used to install Helm charts.
+func installHelmChart(chart helmDeployment, configPath string, logger log.FieldLogger) error {
+	var cmd *exec.Cmd
+	arguments := []string{
+		"--debug",
+		"install",
+		"--kubeconfig", configPath,
+		"-f", chart.valuesPath,
+		chart.chartName,
+		"--namespace", chart.namespace,
+		"--name", chart.chartDeploymentName}
+
+	if chart.setArgument != "" {
+		arguments = append(arguments, "--set", chart.setArgument)
+	}
+
+	cmd = exec.Command("helm", arguments...)
+
+	logger.WithFields(log.Fields{
+		"cmd":  cmd.Path,
+		"args": cmd.Args,
+	}).Info("Invoking command")
+
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Couldn't execute Helm when trying to install the chart %s", chart.chartName))
+	}
+	return nil
+}
+
+// upgradeHelmChart is used to upgrade Helm deployments.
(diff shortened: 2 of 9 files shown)


gsagula:
@gigawhitlocks Nice work 👍Looking at your PR. Give me some time to process it though :)

gigawhitlocks:
Yeah definitely @gsagula no rush! That's why I published the PR before I was 100% ready to merge it

internal/provisioner/utility_group.go
```diff
@@ -0,0 +1,69 @@
...
+func CreateUtilityGroup(group UtilityGroup) error {
+	err := installHelm(group.kops, group.provisioner.logger)
+	if err != nil {
+		return err
+	}
+
+	for _, utility := range clusterUtilities {
+		err := utility.Create(group)
+		if err != nil {
+			return nil
```
gabrieljackson: I like your optimistic approach to not letting errors get in the way of progress, but we should probably return the error here. :)

internal/provisioner/utility_group.go
```diff
@@ -0,0 +1,69 @@
...
+	Destroy(group UtilityGroup) error
+}
+
+// UtilityGroup holds the metadata needed to manage a specific UtilityGroup, and therefore uniquely identifies one, and can be thought of as a handle to the real group of utilities running inside of the cluster
+type UtilityGroup struct {
+	cluster     *model.Cluster
+	provisioner *KopsProvisioner
+	awsClient   aws.AWS
+	kops        *kops.Cmd
+}
```
gabrieljackson: A few things on this:

1. I like the move to an interface here! It really gives us a flexible way to work on things.
2. We should pass a logging object into the utilities and not use the provisioner logger directly. The reason for this is that we add new fields to our loggers in different processes so it would be nice to pass that specific logger in to keep those fields. See the kops CreateCluster flow to get an example of what I mean.
3. UtilityGroup is slightly confusing to me. I would think this would refer to a group of Utilities, but it's actually more like tools for the utilities to do their work. Maybe `UtilityTools` instead? Before you think too much about that though, see my next comment.
4. I think it would do well to change the logic here a bit. Instead of passing in a bunch of tools/things that may be needed by a utility, I think we may want to have a `New()` function for each utility type. The idea here is each utility will init with only the resources it needs and those functions will not be part of the utility interface. That way the `New()` function can also validate the passed in resources. i.e. it can check for things like the various pointers being `nil`. Then the `ClusterUtility` interface will change to something like:
```
type ClusterUtility interface {
	Create() error
	Upgrade() error
	Destroy() error
}
```

internal/provisioner/utility_group.go
```diff
@@ -0,0 +1,69 @@
+package provisioner
+
+import (
+	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// ClusterUtility is a third party service that runs one per cluster and must be managed at cluster creation/deletion time. It is responsible for management of its own lifetime, given a handle pointing to the correct cluster.
```
gabrieljackson: Please try to keep comment lines under 80 chars for better readability.

internal/provisioner/utility_group.go
gigawhitlocks: "optimistic" haha, surely I mistyped here..

internal/provisioner/utility_group.go
```diff
@@ -0,0 +1,69 @@
...
+	kops        *kops.Cmd
+}
+
+// NOTE: if you add a utility to this list, the ORDER MATTERS.
+// The services will be provisioned IN ORDER and if something is provisioned before its dependencies it WILL FAIL
+var clusterUtilities []ClusterUtility = []ClusterUtility{
+	&nginx{}, &prometheus{}, &fluentbit{},
+}
+
+// CreateUtilityGroup creates and starts all of the third party services needed to run a cluster in parallel.
```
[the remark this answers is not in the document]
gigawhitlocks: No, haha, that comment is based on an old revision, need to change it. Right now there are dependencies between the services so the order is required, hard-coded, and I don't believe we can parallelize these three services because of the dependencies between them

gigawhitlocks:
@gabrieljackson I believe I've addressed the bulk of your design concerns. Take a look at let me know if you agree or have more feedback, and I'll start addressing some of the more minor ones about things like comment line length & the logging object comment tomorrow.

internal/provisioner/fluentbit.go
```diff
@@ -0,0 +1,50 @@
...
+	if provisioner == nil {
+		return nil, fmt.Errorf("Cannot create a connection to Prometheus is the provisioner provided is nil")
+	}
+
+	if kops == nil {
+		return nil, fmt.Errorf("Cannot create a connection to Prometheus is the Kops command provided is nil")
+	}
+
+	return &fluentbit{
+
```
gigawhitlocks: I'll take care of these extraneous empty lines, sorry

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -394,6 +318,18 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster) err
 		}
 	}
 
+	// finally install some extra utilities that run the cluster but are
```
gigawhitlocks: comment is wrong, copied it from elsewhere, will reword/delete

internal/provisioner/fluentbit.go
```diff
@@ -0,0 +1,50 @@
...
+type fluentbit struct {
+	provisioner *KopsProvisioner
+	kops        *kops.Cmd
+}
+
+func NewFluentbitHandle(provisioner *KopsProvisioner, kops *kops.Cmd) (*fluentbit, error) {
+
+	if provisioner == nil {
+		return nil, fmt.Errorf("Cannot create a connection to Prometheus is the provisioner provided is nil")
```
gigawhitlocks: when I copy and paste error messages I need to remember to change the name of the service :P

internal/provisioner/helm_utils.go
```diff
@@ -0,0 +1,225 @@
...
+	err = waitForHelmRunning(ctx, kops.GetKubeConfigPath())
+	if err != nil {
+		return err
+	}
+
+	logger.Info("Updating all Helm repos.")
+	return helmRepoUpdate(logger)
+}
+
+func (d *helmDeployment) Provision() error {
```
gigawhitlocks: need to rename this to Upgrade

internal/provisioner/utility_group.go
gigawhitlocks: > We should pass a logging object into the utilities and not use the provisioner logger directly. The reason for this is that we add new fields to our loggers in different processes so it would be nice to pass that specific logger in to keep those fields. See the kops CreateCluster flow to get an example of what I mean.

I _think_ I've done what you're asking here -- used `WithField` to create a new logging object to pass around whenever needing a new logger instead of just reusing the provisioner object. Please have a look and let me know if I have or have not done the right thing.

gigawhitlocks:
This is getting close! Let me know what y'all think..

internal/provisioner/prometheus.go
```diff
@@ -0,0 +1,99 @@
...
+
+type prometheus struct {
+	cluster     *model.Cluster
+	provisioner *KopsProvisioner
+	awsClient   aws.AWS
+	kops        *kops.Cmd
+}
+
+func newPrometheusHandle(cluster *model.Cluster, provisioner *KopsProvisioner, awsClient aws.AWS, kops *kops.Cmd) (*prometheus, error) {
+
```
gabrieljackson: nit: extra newlines

internal/provisioner/utility_group.go
```diff
@@ -0,0 +1,94 @@
...
+		kops:        kops,
+		provisioner: provisioner,
+	}, nil
+
+}
+
+// CreateUtilityGroup  creates  and  starts  all of  the  third  party
+// services needed to run a cluster.
+func (group utilityGroup) CreateUtilityGroup() error {
+
```
gabrieljackson: nit: extra newlines

internal/provisioner/helm_utils.go
```diff
@@ -0,0 +1,225 @@
...
+	err := upgradeHelmChart(*d, d.kops.GetKubeConfigPath(), logger)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Create will install a given HelmDeployment in the cluster specified in that object.
+func (d *helmDeployment) Create() error {
+	logger := d.kopsProvisioner.logger.WithField("helm-create", d.chartName)
```
gabrieljackson: This is better, but isn't quite what I think we want.

We normally create loggers when the provisioner starts managing a cluster (create, upgrade, etc.)

When it does so, it slowly appends fields to the logger (IDs, states, etc) so we should pass those loggers in when we need to log something on one of the utilities instead of creating a new logging object.

Here is an example of a logger we would want pass in as it will contain the cluster ID and other info: https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/kops_provisioner_cluster.go#L85

I could see doing this one of two ways:
1. Change the interface to accept loggers for everything as an argument
2. Require a logger to be passed in when creating the utilities

Either should be fine for now, so go ahead and decide which one makes more sense to you.

gigawhitlocks: OK I have required a logger be passed in while creating the utilities. I think I understand the pattern here -- we accumulate fields on the log entries similar to the way that errors bubble up and accumulate increasingly broad error messages.

In places where there is no obvious metadata that should accompany the field name, when creating the new logger, I have provided a UUID as a value, so that we can trace unique operations in log messages. This was a bit of a judgement call, so feel free to suggest something else if you are not a fan of this pattern, either generally or for specific places where I am creating logger objects.

internal/provisioner/prometheus.go
gigawhitlocks: I don't know why this piece of the diff doesn't show as outdated but I have removed the extra newline

internal/provisioner/prometheus.go
```diff
@@ -0,0 +1,110 @@
...
+	logger := p.logger.WithField("prometheus-create", uuid.New())
+
+	app := "prometheus"
+	dns := fmt.Sprintf("%s.%s.%s", p.cluster.ID, app, p.provisioner.privateDNS)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 120)
+	defer cancel()
+
+	endpoint, err := getLoadBalancerEndpoint(ctx, "internal-nginx", logger.WithField("prometheus-create", uuid.New()), p.kops.GetKubeConfigPath())
```
gabrieljackson: My main concern now is that although it definitely makes sense to add new logger fields for extra info, I think this may be a bit too much. By this time line 67 is called here we have three additional UUIDs used when logging only prometheus create actions.

Considering we will probably have other fields including cluster ID, I would probably leave out the UUIDs and do something like this:

```
	return &prometheus{
		cluster:     cluster,
		provisioner: provisioner,
		awsClient:   awsClient,
		kops:        kops,
		logger:      logger.WithField("cluster-utility", "prometheus"),
	}, nil

...

func (p *prometheus) Create() error {
	err := p.NewHelmDeployment().Create()
	if err != nil {
		return err
	}

	logger := p.logger.WithField("prometheus-action", "create")
...
```

I am not totally against UUIDs here, but I would only add one if any. Keep in mind that when doing any main provisioner action we already have a `run` UUID that would be passed in here.

```
time="2020-01-07T14:54:13Z" level=warning msg="[kops] terraform plan" cluster=og4nzpbstbrszbdwynf9zxza4o instance=6mt635wzqpbnunhsu574iragqc run=jee3iyhbz7n4t8bae1yyhbd66a
time="2020-01-07T14:54:13Z" level=warning msg="[kops] terraform apply" cluster=og4nzpbstbrszbdwynf9zxza4o instance=6mt635wzqpbnunhsu574iragqc run=jee3iyhbz7n4t8bae1yyhbd66a
time="2020-01-07T14:54:13Z" level=warning msg="[kops] Suggestions:" cluster=og4nzpbstbrszbdwynf9zxza4o instance=6mt635wzqpbnunhsu574iragqc run=jee3iyhbz7n4t8bae1yyhbd66a
```

gigawhitlocks: Yeah, I was torn on this. I thought maybe it would be nice to be able to trace these calls,  but it's definitely overkill for the current state of the design and there is likely a runtime cost for generating the UUIDs. But I thought just providing a static string would be redundant. I'll switch to a static string since I feel mixed about both options and you seem to prefer that.

internal/provisioner/helm_utils.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	cmd = exec.Command("helm", arguments...)
+
+	logger.WithFields(log.Fields{
+		"cmd": cmd.Path,
+	}).Info("Invoking command")
+
+	err := cmd.Run()
+	return err
```
gabrieljackson: nit: you have a few different ways of returning errors across your functions mismatch across your functions (you sometimes check the error and sometimes not). It would be nice if you standardized on one.

gigawhitlocks: Ah, yeah this is a style thing, what you're seeing is that it's redundant to do 
```go
if err != nil {
   return err
}
return nil
```
if the function producing `err` is the last operation in the routine and the routine only returns `error` type. I can definitely change these to be more redundant but also more explicit if that's the project style!

gigawhitlocks: Though the quoted one is silly, I assign `err` and then return, lol, so I'll go through and make sure I'm consistent.

gigawhitlocks: Yeah I see what you're saying, I will standardize these error returns (and also not do something silly like assigning `err` and then just returning it lol)

I'm also just going to leave this note to myself to go back and add a little bit of debugging info instead of just returning `err` as-is in most of these cases (using `errors.Wrap()`) now that the functionality is nailed down.

jwilander:
@gigawhitlocks if you can fix the conflict then feel free to merge

gigawhitlocks:
Roger that!

gigawhitlocks:
DO NOT MERGE

I am still testing the rebase & squash

PR #134, opened 2020-01-11
MM-21259 Tag VPCs with Human Owner's Name

gigawhitlocks:
Changes the behavior of the provisioner so that it will tag the VPC with git's `user.email` from the environment where the provisioner is run:
```
~/g/s/g/m/mattermost-cloud » git config --get user.email
ian.whitlock@mattermost.com
```
Screenshot from 2020-01-13 19-02-41

If `git` isn't installed or configured, the tag will just be blank.

Resolves [MM-21259](https://mattermost.atlassian.net/browse/MM-21259).

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -129,7 +129,7 @@ var serverCmd = &cobra.Command{
 		}
 
 		// best-effort attempt to tag the VPC with a human's identity for dev purposes
-		owner := getGitEmail()
+		owner := getHumanReadableID()
 
 		// Setup the provisioner for actually effecting changes to clusters.
 		kopsProvisioner := provisioner.NewKopsProvisioner(
@@ -233,13 +233,18 @@ func deprecationWarnings(logger logrus.FieldLogger, cmd *cobra.Command) {
 	}
 }
 
-// getGitEmail  represents  a  best  effort  attempt  to  retrieve  an
+// getHumanReadableID  represents  a  best  effort  attempt  to  retrieve  an
 // identifiable  human to  associate with  a given  provisioner. Since
-// this is for dev workflows, any  error causes it to merely return an
-// empty string.
-func getGitEmail() string {
+// this is for dev workflows, any  error causes it to merely return a
+// generic string.
+func getHumanReadableID() string {
 	cmd := exec.Command("git", "config",
 		"--get", "user.email")
+
+	envVar := os.Getenv("CLOUD_SERVER_OWNER")
+	if envVar != "" {
+		return envVar
+	}
 
 	output, err := cmd.Output()
 	if err != nil {
@@ -247,7 +252,7 @@ func getGitEmail() string {
 		if len(output) != 0 {
 			logger.Debugf("Command output was: %s", string(output))
 		}
-		return ""
+		return "SRETeam"
 	}
 
 	return string(output)
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
 	"os/signal"
 	"time"
 
@@ -127,10 +128,14 @@ var serverCmd = &cobra.Command{
 			logger.Warn("[DEV] Server is configured to not use cluster VPC claim functionality")
 		}
 
+		// best-effort attempt to tag the VPC with a human's identity for dev purposes
+		owner := getGitEmail()
+
 		// Setup the provisioner for actually effecting changes to clusters.
 		kopsProvisioner := provisioner.NewKopsProvisioner(
 			s3StateStore,
 			privateDNS,
+			owner,
 			useExistingResources,
 			logger,
 		)
@@ -226,4 +231,24 @@ func deprecationWarnings(logger logrus.FieldLogger, cmd *cobra.Command) {
 	if certificateARNFlag != "" {
 		logger.Warn("[Deprecation] Flag certificate-aws-arn is deprecated. Provisioner will use a resource tag to find certificate-aws-arn in AWS.")
 	}
+}
+
+// getGitEmail  represents  a  best  effort  attempt  to  retrieve  an
+// identifiable  human to  associate with  a given  provisioner. Since
+// this is for dev workflows, any  error causes it to merely return an
+// empty string.
+func getGitEmail() string {
+	cmd := exec.Command("git", "config",
+		"--get", "user.email")
+
+	output, err := cmd.Output()
+	if err != nil {
+		logger.Debugf("Couldn't determine username of developer with which to tag infrastructure due to error: %s", err.Error())
+		if len(output) != 0 {
+			logger.Debugf("Command output was: %s", string(output))
+		}
+		return ""
+	}
+
+	return string(output)
 }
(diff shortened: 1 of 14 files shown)


gigawhitlocks:
Ah, yeah that's slightly different behavior.  We can chat about it for sure, this PR still needs a bit more work anyway.

gigawhitlocks:
OK, I see the misunderstanding, certainly that would've been a little easier to implement. Let me share some of my reasoning (and why it did not even occur to me to just have a tag applied that simply identifies the provisioner that created the cluster, as opposed to the end user that issued the request).

In my understanding, in production there can/will be multiple end-users using one provisioning server through its API. If we tag the VPC with a flag from the server's startup, then we can only see which server provisioned the cluster, whereas with what I have implemented, the end-user has the ability (although it's currently only enforced through the command line client) to tag each cluster with an owner. Personally that seems more useful to me, because then we can extend that functionality to consumers of the API to allow tagging of VPCs created via, e.g., the cloud plugin.

Anyway I'm happy to re-implement however we feel is better.

gabrieljackson:
Yeah, I can see where you are coming from. The two main things I am considering are as follows:

1. The original intention of this tagging behavior was as a dev-focused option. i.e. We weren't going to tag stuff in our test, staging, etc. accounts as those clusters should all generally be uniform and/or not really owned by anyone. In those accounts, the owner is not really something I think we want due to the following.
2. Clusters could change purpose and/or "owners" over time. The way the scheduling algorithm currently works is any cluster that fits a given installation's criteria can be used. As such, we don't know where installations will end up necessarily and if there will be one or many installations on them. As a result, the tags for any of these clusters would have to be generic as we can't know what their ultimate purpose will be. This leads me around to the idea of only tagging clusters/VPCs when devs are doing testing work as none of those clusters should live forever and seeing the tag will immediately make the viewer know that the cluster was created for some dev purpose.

Let me know what you think about this. If we didn't have to update the DB schema I would think it would be less of an issue, but I am hesitant to add new data that will only be useful in dev-focused work.

gigawhitlocks:
My understanding was that the purpose of this tag is essentially so that an administrator can track down the owner of a given cluster, and I think that has utility beyond the obvious dev use-case. I was thinking that for (2) it could still be helpful to have a team name or something like that assigned to the cluster, in case we're not the only ones testing.

Along those lines of thinking, the input is only enforced in the CLI tool, so that an empty owner tag is acceptable in non-dev workflows, but can still be utilized if desired.

The DB change is a lot of additional complexity for a small feature, though, so I definitely see where you're coming from. If nobody else chimes in here with an opinion I'll change this behavior to be an ID per-provisioner instead. 

If this is for dev workflow only, though, how about instead of adding a flag to the provisioner, I'll just automatically add the user's configured git email to the tag if it is available and skip tagging if it's unavailable, so that we can have this info without any user intervention?

gabrieljackson:
"If this is for dev workflow only, though, how about instead of adding a flag to the provisioner, I'll just automatically add the user's configured git email to the tag if it is available and skip tagging if it's unavailable, so that we can have this info without any user intervention?"

I like this idea, but I am not entirely sure how our production provisioner containers are set up. They shouldn't have any git configuration, but I am not entirely sure about that. That said, if this would work and not add any additional tags then I think it is the best idea yet!

gigawhitlocks:
OK I have reworked this so that it attempts to pull the information from `git`, updated the PR description to reflect the new behavior, and added @cpanato to the review so he can comment on if this is ok in production.

cpanato:
this will work only in the DEV env correct?

gigawhitlocks:
No, but if git is not configured on the system, it'll skip tagging.

gigawhitlocks:
Sure! I just added this change for you @cpanato, in the last commit.

internal/tools/aws/cluster_management.go
```diff
@@ -203,11 +204,21 @@ func (a *Client) claimVpc(clusterResources ClusterResources, clusterID string, l
...
 		return errors.Wrapf(err, "unable to update %s", VpcAvailableTagKey)
 	}
+
 	err = a.TagResource(clusterResources.VpcID, trimTagPrefix(VpcClusterIDTagKey), clusterID, logger)
 	if err != nil {
 		return errors.Wrapf(err, "unable to update %s", VpcClusterIDTagKey)
 	}
 
+	if owner == "" {
+		owner = "SRETeam"
```
cpanato: why not return here: https://github.com/mattermost/mattermost-cloud/pull/134/files#diff-e240f1301da8fcd8a09bc6dd63baf09bR242

gabrieljackson: Agreed, that seems better.

Also, 0/5 on this, but we could lookup an env var instead of hardcoding SRETeam. That way we could change it between our AWS accounts if we want.

gigawhitlocks: Oh, haha, yeah it makes more sense to put the logic there doesn't it?

gigawhitlocks: I do like the env var suggestion @gabrieljackson but do you have a suggestion for the default value if the env var is not set?

gigawhitlocks: I combined these suggestions: the routine now first looks for an environment variable and uses that, if it is set. If not, it looks for git. If that still fails, it sets the string to the generic "SRETeam"

gigawhitlocks:
Just realized this has a defect where it leaves the tag after the cluster is destroyed. Need to add a call to remove the tag to delete, let me do that before you review (to whoever happens to read this before I post another commit).

cmd/cloud/server.go
```diff
@@ -227,3 +233,28 @@ func deprecationWarnings(logger logrus.FieldLogger, cmd *cobra.Command) {
 		logger.Warn("[Deprecation] Flag certificate-aws-arn is deprecated. Provisioner will use a resource tag to find certificate-aws-arn in AWS.")
 	}
 }
+
+// getHumanReadableID  represents  a  best  effort  attempt  to  retrieve  an
+// identifiable  human to  associate with  a given  provisioner. Since
+// this is for dev workflows, any  error causes it to merely return a
+// generic string.
+func getHumanReadableID() string {
+	cmd := exec.Command("git", "config",
```
gabrieljackson: nit: setting command before possibly just returning an env var.

cmd/cloud/server.go
```diff
@@ -227,3 +233,28 @@ func deprecationWarnings(logger logrus.FieldLogger, cmd *cobra.Command) {
...
+		"--get", "user.email")
+
+	envVar := os.Getenv("CLOUD_SERVER_OWNER")
+	if envVar != "" {
+		return envVar
+	}
+
+	output, err := cmd.Output()
+	if err != nil {
+		logger.Debugf("Couldn't determine username of developer with which to tag infrastructure due to error: %s", err.Error())
```
gabrieljackson: nit: hmm, which logger is this using? One isn't passed in. As long as it works though it's not a big deal.

gigawhitlocks: Here we're in package `main` so this is the global logger from `logger.go`. Since this is just a helper function I think that makes sense, is the correct context, and I don't think initializing one with new fields will buy us anything.

cmd/cloud/server.go
gigawhitlocks: good catch

PR #136, opened 2020-01-13
check if the AWS AMI exists before create the k8s cluster

cpanato:
#### Summary
check if the AWS AMI exists

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-21704

### DIFF
--- a/internal/tools/aws/ec2.go
+++ b/internal/tools/aws/ec2.go
@@ -169,11 +169,7 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup, 
 }
 
 // IsValidAMI check if the provided AMI exists
-func (a *Client) IsValidAMI(AMIImage string) bool {
-	if AMIImage == "" {
-		return false
-	}
-
+func (a *Client) IsValidAMI(AMIImage string) (bool, error) {
 	svc := ec2.New(session.New(), &aws.Config{
 		Region: aws.String(DefaultAWSRegion),
 	})
@@ -189,13 +185,11 @@ func (a *Client) IsValidAMI(AMIImage string) bool {
 
 	out, err := a.api.describeImages(svc, describeImageInput)
 	if err != nil {
-		log.Errorf("failed to find ami %s. err=%s", AMIImage, err.Error())
-		return false
+		return false, err
 	}
 	if len(out.Images) == 0 {
-		log.Errorf("found no AMIs with the ID %s. err=%s", AMIImage, err.Error())
-		return false
+		return false, err
 	}
 
-	return true
+	return true, nil
 }
--- a/internal/tools/aws/ec2.go
+++ b/internal/tools/aws/ec2.go
@@ -170,6 +170,10 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup, 
 
 // IsValidAMI check if the provided AMI exists
 func (a *Client) IsValidAMI(AMIImage string) (bool, error) {
+	// if AMI image is blank it will use the default KOPS image
+	if AMIImage == "" {
+		return true, nil
+	}
 	svc := ec2.New(session.New(), &aws.Config{
 		Region: aws.String(DefaultAWSRegion),
 	})
@@ -188,7 +192,7 @@ func (a *Client) IsValidAMI(AMIImage string) (bool, error) {
 		return false, err
 	}
 	if len(out.Images) == 0 {
-		return false, err
+		return false, nil
 	}
 
 	return true, nil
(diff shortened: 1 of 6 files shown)


internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,35 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
 
 	return sgOutput.SecurityGroups, nil
 }
+
+// IsValidAMI check if the provided AMI exists
+func (a *Client) IsValidAMI(AMIImage string) bool {
+	if AMIImage == "" {
```
gigawhitlocks: Do we want to log something here like the other error cases? Actual question, not even really a suggestion, so do with this comment what you please :)

internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,35 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
...
+			{
+				Name:   aws.String("image-id"),
+				Values: []*string{aws.String(AMIImage)},
+			},
+		},
+	}
+
+	out, err := a.api.describeImages(svc, describeImageInput)
+	if err != nil {
+		log.Errorf("failed to find ami %s. err=%s", AMIImage, err.Error())
```
gabrieljackson:
```suggestion
		log.WithError(err).Errorf("Failed to find AMI %s", AMIImage)
```

internal/tools/aws/ec2.go
gabrieljackson: Yeah, unless I am wrong in how I see this being called it is possible to pass in `""` which is valid for kops as it will use the default, but here we are going to return false which doesn't seem right. Am I missing something?

internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,35 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
 
 	return sgOutput.SecurityGroups, nil
 }
+
+// IsValidAMI check if the provided AMI exists
+func (a *Client) IsValidAMI(AMIImage string) bool {
```
gabrieljackson: I think it may make more sense to return `(bool, error)` here instead of logging within the function. No logger is passed in so we wouldn't have all the extra logging fields that get created before this is called. If we return errors we can log them a bit better.

internal/tools/aws/ec2.go
gabrieljackson: Note: this doesn't apply if you change the function to return errors instead.

internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,35 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
...
+		},
+	}
+
+	out, err := a.api.describeImages(svc, describeImageInput)
+	if err != nil {
+		log.Errorf("failed to find ami %s. err=%s", AMIImage, err.Error())
+		return false
+	}
+	if len(out.Images) == 0 {
+		log.Errorf("found no AMIs with the ID %s. err=%s", AMIImage, err.Error())
```
gabrieljackson: Same thing here

internal/tools/aws/ec2.go
cpanato: makes sense, thanks for pointing this out

internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,29 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
...
+			},
+		},
+	}
+
+	out, err := a.api.describeImages(svc, describeImageInput)
+	if err != nil {
+		return false, err
+	}
+	if len(out.Images) == 0 {
+		return false, err
```
gabrieljackson: `err` would be nil here. We should probably return something else or `nil` directly.

internal/tools/aws/ec2.go
```diff
@@ -163,3 +167,29 @@ func GetSecurityGroupsWithFilters(filters []*ec2.Filter) ([]*ec2.SecurityGroup,
 
 	return sgOutput.SecurityGroups, nil
 }
+
+// IsValidAMI check if the provided AMI exists
+func (a *Client) IsValidAMI(AMIImage string) (bool, error) {
```
gabrieljackson: What happens if the AMI string passed in is `""` which is the default if no AMI is provided? Does this work okay?

cpanato: humm, so we have "" if is default kops image? i thought it fulfill with the default kops image and never will be blank

so we need to check if is blank we bypass, sounds good? @gabrieljackson

gabrieljackson: Yeah, exactly!

Here is an example of what the default kops metadata looks like:

`{"Name":"oifityg3r7b4z8cyfc9uo66gea-kops.k8s.local","Version":"latest","AMI":""}`

cpanato: got it

PR #143, opened 2020-01-22
MM-20995 Allow cluster utility versions to be managed via the API

gigawhitlocks:
I have added an optional HTTP Request body that may be provided to ProvisionCluster, and a new field to the same Request body for CreateCluster, that may be used to specify the versions of Prometheus, Nginx, and Fluentbit.

Usage: you can create a new cluster and specify a version (HELM CHART VERSION not application version) on the command line like so:

```
~ » cloud cluster create --zones us-east-1c --size SizeAlef500 --prometheus-version 10.3.0
{
    "ID": "d7ijtqc13iftuetakg7hgezeww",
    "Provider": "aws",
    "Provisioner": "kops",
    "ProviderMetadata": "eyJab25lcyI6WyJ1cy1lYXN0LTFjIl19",
    "ProvisionerMetadata": "eyJOYW1lIjoiIiwiVmVyc2lvbiI6ImxhdGVzdCIsIkFNSSI6IiJ9",
    "AllowInstallations": true,
    "Version": "0.0.0",
    "Size": "SizeAlef500",
    "State": "creation-requested",
    "CreateAt": 1580505171241,
    "DeleteAt": 0,
    "LockAcquiredBy": null,
    "LockAcquiredAt": 0,
    "UtilityMetadata": "eyJkZXNpcmVkVmVyc2lvbnMiOnsiUHJvbWV0aGV1cyI6IjEwLjMuMCIsIk5naW54IjoiIiwiRmx1ZW50Yml0IjoiIn0sImFjdHVhbFZlcnNpb25zIjp7IlByb21ldGhldXMiOiIiLCJOZ2lueCI6IiIsIkZsdWVudGJpdCI6IiJ9fQ=="
}
```
Then we can check the chart version. During cluster create:
```
~ » cloud cluster utilities --cluster d7ijtqc13iftuetakg7hgezeww
{
    "desiredVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "",
        "Nginx": "",
        "Fluentbit": ""
    }
}
```
Once the action is finished running, the actual version will be updated in the above output:
```
cloud cluster utilities --cluster d7ijtqc13iftuetakg7hgezeww
{
    "desiredVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "1.29.6",
        "Fluentbit": "2.8.7"
    }
}
```

After the fact, you can use `cloud cluster provision` to pin a new version,
```shell
~/g/s/g/m/mattermost-cloud » cloud cluster provision --cluster 1yj86sia3b8bzx1a8t611j46ca --prometheus-version 10.3.0
# ... wait for provision to complete, then
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 1yj86sia3b8bzx1a8t611j46ca 
{
    "desiredVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "1.29.6",
        "Fluentbit": "2.8.7"
    }
}

```
 or use the special keyword `"stable"` to move the utility to track the latest stable release:
```shell
~/g/s/g/m/mattermost-cloud » cloud cluster provision --cluster 1yj86sia3b8bzx1a8t611j46ca --prometheus-version stable --nginx-version 1.29.4
# during provisioning, actualVersions and desiredVersions show different values:
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 1yj86sia3b8bzx1a8t611j46ca
{
    "desiredVersions": {
        "Prometheus": "",
        "Nginx": "1.29.4",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "10.3.0",
        "Nginx": "1.29.6",
        "Fluentbit": "2.8.7"
    }
}
# after provisioning, everything is the same:
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 1yj86sia3b8bzx1a8t611j46ca
{
    "desiredVersions": {
        "Prometheus": "",
        "Nginx": "1.29.4",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "10.4.0",
        "Nginx": "1.29.4",
        "Fluentbit": "2.8.7"
    }
}
```
If the new flags are omitted from `create` the provisioner will default to the old behavior (latest stable version) and if they are omitted from `provision` the version will not be changed -- a pinned version will remained pinned and an unspecified version will update if a new stable release is available.


internal/store/migrations.go
```diff
@@ -578,6 +578,68 @@ var migrations = []migration{
...
+					Size                 TEXT NOT NULL,
+					State                TEXT NOT NULL,
+					AllowInstallations   BOOLEAN NOT NULL,
+					CreateAt             BIGINT NOT NULL,
+					DeleteAt             BIGINT NOT NULL,
+					LockAcquiredBy       TEXT NULL,
+					LockAcquiredAt       BIGINT NOT NULL,
+					PrometheusVersion    TEXT NOT NULL,
+					FluentbitVersion     TEXT NOT NULL,
+					NginxVersion         TEXT NOT NULL
```
gabrieljackson: I am not entirely sure we want separate database entries for these. Something like the Metadata entries we do above may make more sense as they are flexible for multiple types of data about a variety of utilities.

In the future, we may have more/less utilities. Having each one have a hardcoded "presence" in the database and various models will probably be messy and/or a hassle in the future.

gigawhitlocks: I'm fine with following the pattern laid out by the Metadata entries at least for now, but after a lot of thinking and going back and forth with myself, I think we might want to discuss our overall strategy for storing "schemaless" data (my guess is that this will almost always be JSON). 

The current pattern we follow for mapping our database to an in-memory object is a little bit of a leaky abstraction, and there are a number of ways to address that but I want to avoid derailing myself. But that leaky abstraction will have runtime performance related effects in this PR: once I change the type in Cluster from 3 `string` fields to one `[]byte`, then anything that needs to operate on any one field in the JSON hidden in that `[]byte` will have to first unmarshal the blob, and then that unmarshalled struct also cannot be cached anywhere standard, since the Cluster field types need to map to the SQL column types (without a large change to the `store` package).

There are a number of approaches to overcome these limitations but they are all tradeoffs so we'll have to discuss. 

For now I'll just use `[]byte` in the struct and BYTEA in the database as types for the data and we will incur the (un)marshalling runtime costs as a tradeoff for a simplified database schema.

gigawhitlocks:
Comment about the schema is addressed in the commit above @gabrieljackson and I'll be tackling the other portion that we discussed at length in DM tomorrow.

model/cluster.go
```diff
@@ -105,3 +112,63 @@ var clusterVersionMatcher = regexp.MustCompile(`^(([0-9]{1,3}.[0-9]{1,3}.[0-9]{1
...
+type utilityVersions struct {
+	Prometheus string
+	Nginx      string
+	Fluentbit  string
+}
+
+// GetUtilityVersion fetches the desired version of a utility from the
+// Cluster object
+func (c *Cluster) GetUtilityVersion(utility string) (string, error) {
+
```
gabrieljackson: nit: extra newline

internal/api/cluster.go
```diff
@@ -88,6 +88,12 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	err = cluster.SetProviderMetadata(model.AWSMetadata{
 		Zones: createClusterRequest.Zones,
 	})
```
gabrieljackson: It looks like you disconnected the error handling for this with the new code

internal/api/cluster.go
```diff
@@ -88,6 +88,12 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
...
+
+	err = cluster.UpdateUtilityMetadata(createClusterRequest.UtilityMetadata)
+	if err != nil {
+		c.Logger.WithError(err).Errorf("provided utility metadata could not be applied without error")
```
gabrieljackson:
```suggestion
		c.Logger.WithError(err).Error("unable to set utility metadata")
```

cmd/cloud/cluster.go
```diff
@@ -264,3 +279,46 @@ var clusterShowStateReport = &cobra.Command{
 		return nil
 	},
 }
+
+func getRequestedUtilityMetadata(command *cobra.Command) map[string]string {
```
gabrieljackson: This should be be handled server side. i.e. we should do any default-setting when the ProvisionClusterRequest is constructed from the reader. This allows non-cloud-cli behavior to be more similar to simply sending raw JSON payloads and also allows us to do everything in one place.

Look at the CreateClusterRequest from reader for an example on default and validation.
https://github.com/mattermost/mattermost-cloud/blob/master/model/cluster_request.go#L23

internal/provisioner/fluentbit.go
```diff
@@ -33,24 +35,41 @@ func newFluentbitHandle(provisioner *KopsProvisioner, awsClient aws.AWS, kops *k
 		return nil, errors.New("cannot create a connection to Fluentbit if the Kops command provided is nil")
 	}
 
+	version, err := cluster.GetUtilityVersion("fluentbit")
```
gabrieljackson: Does it make sense to just pass in the version for the utility instead of the whole cluster object? 0/5 on this.

internal/provisioner/fluentbit.go
```diff
@@ -33,24 +35,41 @@ func newFluentbitHandle(provisioner *KopsProvisioner, awsClient aws.AWS, kops *k
...
+	if err != nil {
+		return nil, errors.Wrap(err, "something went wrong while getting chart version for Prometheus")
```
gabrieljackson:
```suggestion
		return nil, errors.Wrap(err, "something went wrong while getting chart version for Fluentbit")
```

internal/api/cluster.go
```diff
@@ -88,6 +88,12 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	err = cluster.SetProviderMetadata(model.AWSMetadata{
 		Zones: createClusterRequest.Zones,
 	})
+
+	err = cluster.UpdateUtilityMetadata(createClusterRequest.UtilityMetadata)
```
gabrieljackson: nit: `SetUtilityMetadata` would match the naming scheme used elsewhere.

model/cluster.go
```diff
@@ -105,3 +112,63 @@ var clusterVersionMatcher = regexp.MustCompile(`^(([0-9]{1,3}.[0-9]{1,3}.[0-9]{1
...
+type utilityVersions struct {
+	Prometheus string
+	Nginx      string
+	Fluentbit  string
+}
+
+// GetUtilityVersion fetches the desired version of a utility from the
+// Cluster object
+func (c *Cluster) GetUtilityVersion(utility string) (string, error) {
```
gabrieljackson: As you probably know, the caching-on-cluster-object in this function is a departure from how we simply return metadata objects for things like the provisioner metadata.

Did you find that simply returning an object here led to needing to unmarshal the same object multiple times for each supervisor flow? Considering that each time the supervisors wake up to work they will be re-querying the database for cluster objects, the caching done here should be short-lived and will need to be redone anyway.

I am mainly curious if this may just be extra complexity for now. If we do want to cache metadata, it may be best handled in a PR that converts all of these values to that behavior.

Thoughts?

gigawhitlocks: augh, DMing you, I apparently pushed changes I didn't mean to while resolving conflicts from your PR @gabrieljackson

gigawhitlocks: I just didn't want to do json deserialization/serialization every time someone requested this object. To be honest, I don't like storing the data as a []byte but our data models are a leaky abstraction, reflecting the underlying SQL data type. So if it's a BYTEA in the database, it has to be a []byte in Cluster, which means that since that []byte is actually JSON, we can either serialize/deserialize every single time, or we can implement an intermediate Cluster representation in memory after taking it out of the database, or we can revert back to my old schema with typed fields for each piece of metadata that we want to store, or we can cache the native representation of the deserialized data like this and use a Getter, which is the simplest solution to avoid (de)serializing the JSON every time we want to update or read from this field.

internal/provisioner/fluentbit.go
gigawhitlocks: yes it does, for Prometheus I pass `cluster` for other reasons and it looks like I just dumbly copied that pattern

cmd/cloud/cluster.go
gigawhitlocks: Moved the part that has to do with defaults and isn't just basic processing to the api side

gigawhitlocks:
The current version of this is now updated to work as requested but I still need to go through and do cleanup and the implementation is not very DRY. If you want to look at the overall design, though, that should be in place now with the concerns brought up in this PR, and please feel free to leave me comments about the cleanup if you see anything that concerns you before I get to it.

Also, there is not yet a client implementation for the new API endpoint, `api/cluster/{id}/utilities` but the API works and responds like so (in this example Fluentbit is the only service with a pinned version):

```shell
curl localhost:8075/api/cluster/eg9k6ah78p8gzmyjmsbhx74pcy/utilities 2>/dev/null| jq .
{
  "desiredVersions": {
    "Prometheus": "",
    "Nginx": "",
    "Fluentbit": "2.8.4"
  },
  "actualVersions": {
    "Prometheus": "prometheus-10.4.0",
    "Nginx": "nginx-ingress-1.29.6",
    "Fluentbit": "fluent-bit-2.8.4"
  }
}
```

I'll also fix the style check failures in the morning.

gigawhitlocks:
Now there is also a client implementation. Best explained through some example usage:
```
~/g/s/g/m/mattermost-cloud » cloud cluster
Manipulate clusters managed by the provisioning server.

Usage:
  cloud cluster [command]

Available Commands:
  create       Create a cluster.
  delete       Delete a cluster.
  get          Get a particular cluster.
  installation Manipulate cluster installations managed by the provisioning server.
  list         List created clusters.
  provision    Provision/Reprovision a cluster's k8s operators.
  state-report Shows information regarding changing cluster state.
  update       Updates a cluster's configuration.
  upgrade      Upgrade k8s on a cluster.
  utilities    Show metadata regarding utility services running in a cluster.

Flags:
  -h, --help            help for cluster
      --server string   The provisioning server whose API will be queried. (default "http://localhost:8075")

Use "cloud cluster [command] --help" for more information about a command.
```
```
~/g/s/g/m/mattermost-cloud » cloud cluster utilities
Usage:
  cloud cluster utilities [flags]

Flags:
      --cluster string   The id of the cluster whose utilities are to be fetched.
  -h, --help             help for utilities

Global Flags:
      --server string   The provisioning server whose API will be queried. (default "http://localhost:8075")

ERRO[2020-01-31T13:13:32-06:00] command failed                                error="required flag(s) \"cluster\" not set"
```
```
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 1yj86sia3b8bzx1a8t611j46ca
{
    "desiredVersions": {
        "Prometheus": "",
        "Nginx": "",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "prometheus-10.4.0",
        "Nginx": "nginx-ingress-1.29.6",
        "Fluentbit": "fluent-bit-2.8.7"
    }
}
```

gigawhitlocks:
This PR is stable pending further review.

model/cluster_utility.go
```diff
@@ -0,0 +1,156 @@
...
+	return ""
+}
+
+// setUtilityVersion will assign the version in desiredVersion to the
+// utility whose name's string representation matches one of the known
+// utilities with a version field in utilityVersion struct in the
+// first argument
+func setUtilityVersion(versions *utilityVersions, utility, desiredVersion string) {
+	switch utility {
+	case "prometheus":
```
jwilander: nit: Might be worth constants for these

model/cluster_utility.go
```diff
@@ -0,0 +1,156 @@
+package model
```
jwilander: What about some small unit tests for this file?

internal/api/cluster.go
```diff
@@ -22,10 +22,12 @@ func initCluster(apiRouter *mux.Router, context *Context) {
...
 	clusterRouter.Handle("", addContext(handleGetCluster)).Methods("GET")
 	clusterRouter.Handle("", addContext(handleRetryCreateCluster)).Methods("POST")
-	clusterRouter.Handle("", addContext(handleUpdateCluster)).Methods("PUT")
+	clusterRouter.Handle("", addContext(handleUpdateClusterConfiguration)).Methods("PUT")
 	clusterRouter.Handle("/provision", addContext(handleProvisionCluster)).Methods("POST")
-	clusterRouter.Handle("/kubernetes/{version}", addContext(handleUpgradeCluster)).Methods("PUT")
+	clusterRouter.Handle("/kubernetes/{version}", addContext(handleUpgradeKubernetes)).Methods("PUT")
 	clusterRouter.Handle("", addContext(handleDeleteCluster)).Methods("DELETE")
+
+	clusterRouter.Handle("/utilities", addContext(handleGetAllUtilityMetadata)).Methods("GET")
```
jwilander: Tests for new endpoints and new endpoint features?

model/cluster_utility.go
gigawhitlocks: yep, makes sense, I'll write some tests

internal/api/cluster.go
gigawhitlocks: yup I'll get some tests on this too

model/cluster_utility.go
gigawhitlocks: yeah that makes sense

model/client.go
```diff
@@ -203,6 +204,22 @@ func (c *Client) GetClusters(request *GetClustersRequest) ([]*Cluster, error) {
...
+// GetClusterUtilities returns the metadata for all utilities running in the given cluster.
+func (c *Client) GetClusterUtilities(clusterID string) (*UtilityMetadata, error) {
+	resp, err := c.doGet(c.buildURL("/api/cluster/%s/utilities", clusterID))
+	if err != nil {
+		return nil, err
+	}
+	defer closeBody(resp)
+
+	switch resp.StatusCode {
+	case http.StatusAccepted:
```
gabrieljackson: Will need to be statusOK if the above change is made

internal/provisioner/nginx.go
```diff
@@ -43,13 +90,10 @@ func (n *nginx) Create() error {
...
-func (n *nginx) Upgrade() error {
-	return nil
+		desiredVersion:      n.desiredVersion,
+	}
 }
 
-func (n *nginx) Destroy() error {
-	return nil
+func (n *nginx) Name() string {
+	return "nginx"
```
gabrieljackson: Constant?

cmd/cloud/cluster.go
```diff
@@ -247,6 +265,33 @@ var clusterListCmd = &cobra.Command{
...
+
+		metadata, err := client.GetClusterUtilities(clusterID)
+		if err != nil {
+			return err
+		}
+
+		err = printJSON(metadata)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
gigawhitlocks: Yeah, easier said than done (in a robust way) because of the way Helm reports versions :\

internal/api/cluster.go
```diff
@@ -449,3 +474,18 @@ func handleDeleteCluster(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	clusterID := vars["cluster"]
+	c.Logger = c.Logger.WithField("cluster", clusterID).WithField("action", "get-utilities")
+
+	cluster, err := c.Store.GetCluster(clusterID)
+	if err != nil {
+		c.Logger.WithError(err).Errorf("failed to look up cluster %s", clusterID)
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusAccepted)
```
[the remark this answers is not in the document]
gabrieljackson: To follow up on this, all of our other `Get` APIs return OK, so this is more of a standards thing

cmd/cloud/cluster.go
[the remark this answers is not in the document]
gabrieljackson: Yeah, agreed. Helm 3 and the Go library may be our path forward to that.

internal/provisioner/fluentbit.go
```diff
@@ -34,23 +36,61 @@ func newFluentbitHandle(provisioner *KopsProvisioner, awsClient aws.AWS, kops *k
...
+func (f *fluentbit) DesiredVersion() string {
+	return f.desiredVersion
+}
+
+func (f *fluentbit) ActualVersion() string {
+	return f.actualVersion
+}
+
+func (f *fluentbit) Name() string {
+	return "fluentbit"
```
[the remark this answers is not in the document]
gigawhitlocks: good eye

cmd/cloud/cluster.go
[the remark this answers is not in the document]
gigawhitlocks: Yeah and at the very least this change should make the transition to that easier.

internal/supervisor/cluster.go
```diff
@@ -171,10 +171,10 @@ func (s *ClusterSupervisor) createCluster(cluster *model.Cluster, logger log.Fie
...
 		if cluster.Version != version {
 			cluster.Version = version
-			err = s.store.UpdateCluster(cluster)
-			if err != nil {
-				logger.WithError(err).Warnf("failed to set cluster version to %s", version)
-			}
+		}
+		err = s.store.UpdateCluster(cluster)
+		if err != nil {
+			logger.WithError(err).Warnf("failed to save cluster with version %s", version)
```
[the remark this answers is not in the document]
gigawhitlocks: No, not unless I am misunderstanding you. But there is a change in behavior here.
In the old behavior we checked to see if the k8s version had changed and wrote those changes to the Cluster representation in the database if that was the case.

Now, we write the Cluster object regardless, to persist changes to the Cluster object made by ProvisionCluster and/or CreateCluster (in this case, I need to save the actual versions of the utilities in the Cluster object to the database).

So this behavior is a little different but regarding the error string (and in the one below, your other comment), the error is only going to be the reason that the Cluster object couldn't be saved to the database for whatever reason that might be, at this time.

internal/api/cluster.go
```diff
@@ -301,9 +326,9 @@ func handleUpdateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	outputJSON(c, w, cluster)
 }
 
-// handleUpgradeCluster responds to PUT /api/cluster/{cluster}/kubernetes/{version},
+// handleUpgradeKubernetes responds to PUT /api/cluster/{cluster}/kubernetes/{version},
 // upgrading the cluster to the given Kubernetes version.
-func handleUpgradeCluster(c *Context, w http.ResponseWriter, r *http.Request) {
+func handleUpgradeKubernetes(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks: Is this change OK? I think it makes the meaning of the method clearer.

internal/api/cluster.go
```diff
@@ -262,9 +288,9 @@ func handleGetCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	outputJSON(c, w, cluster)
 }
 
-// handleUpdateCluster responds to PUT /api/cluster/{cluster}, updating a cluster's
+// handleUpdateClusterConfiguration responds to PUT /api/cluster/{cluster}, updating a cluster's
 // configuration.
-func handleUpdateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
+func handleUpdateClusterConfiguration(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks: Is this change OK? I think it makes the meaning of the method clearer.

cmd/cloud/cluster.go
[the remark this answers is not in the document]
gigawhitlocks: Actually thinking about it a little more, this shouldn't be too bad because of the way I've laid out the interface.. I should be able to use string manipulation to extract the version from any given chart because it can be a utility-specific implementation.. so that shouldn't be brittle. I will go ahead and do this so that the result in the JSON struct is formatted the same and if/when we want we can just compare those fields on the client side.

tl;dr after thinking about it, I can reliably format actualVersions and desiredVersions identically without too much trouble or creating brittle code, so I will do that.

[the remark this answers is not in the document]
gigawhitlocks: Updated the implementation for each Utility to have the method ActualVersion() strip the Helm chart name prefix from each Utility when returning the value.

I realized after you brought this up that this should be perfectly reliable since it can vary for each Utility and it should be easy to continue the pattern if/when we move to Helm 3 and/or Operators, even if we make that change piecemeal:

https://github.com/mattermost/mattermost-cloud/pull/143/commits/606ab4da45091dc2403ded69473b1d9bbe9769d1

I was originally worried that doing this would be brittle but since it can be implemented specifically for each utility, this should be perfectly robust.

internal/api/cluster.go
gabrieljackson: Fine with me!

internal/api/cluster.go
gabrieljackson: 👍

internal/supervisor/cluster.go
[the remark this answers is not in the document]
gabrieljackson: Oh yeah, that is exactly what I was referring to. I just was wondering if the error should now be something more like "unable to save updated cluster" because it isn't just the version changing anymore.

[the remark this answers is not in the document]
gigawhitlocks: Ha! OK I see now, that's the intended meaning of the error message, but apparently it's not clear. I'll reword it to be clearer.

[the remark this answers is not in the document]
gigawhitlocks: Should be better now https://github.com/mattermost/mattermost-cloud/pull/143/commits/388c01c5f041013c3072e234ed46996f3b49e1e3

internal/supervisor/cluster.go
```diff
@@ -197,10 +197,10 @@ func (s *ClusterSupervisor) provisionCluster(cluster *model.Cluster, logger log.
...
```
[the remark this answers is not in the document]
gigawhitlocks: Fixed here https://github.com/mattermost/mattermost-cloud/pull/143/commits/388c01c5f041013c3072e234ed46996f3b49e1e3

internal/api/cluster.go
gigawhitlocks: resolved via https://github.com/mattermost/mattermost-cloud/pull/143/commits/e916f8ac439314502f179367532e54f8d78533be

model/cluster_utility.go
gigawhitlocks: tests for this file are in this commit https://github.com/mattermost/mattermost-cloud/pull/143/commits/5bd517e86af2583ee4f18f2c0c6bc860526778f9

model/cluster_utility_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	require.Equal(t, c.UtilityMetadata, []byte(nil))
+	err := c.SetUtilityActualVersion(NGINX, "1.9.9")
+	require.NoError(t, err)
+
+	require.NotEqual(t, []byte(nil), c.UtilityMetadata)
+
+	version, err := c.ActualUtilityVersion(NGINX)
+	require.NoError(t, err)
+	require.Equal(t, "1.9.9", version)
```
jwilander: nit: I prefer to use `require` only for steps in the test that must pass to properly test the next step or to prevent a panic and `assert` when doing the check you're actually focused on. That way you have a chance to see multiple failures in one run of the test, instead of having to run, fix, run, fix, etc.

gigawhitlocks: Oh okay, sounds good!

PR #144, opened 2020-01-23
set booleans flags to false as default

cpanato:
#### Summary
Setting the boolean flags to false as default

The reason behind this change is to make easier in k8s deployment. 
Tested the current way and looks like we cannot set a false when needed because it set as a string and the command does not recognize that

snippet of the deployment
```
containers:
      - args:
        - server
        - --debug
        - "true"
        - --state-store
        - mattermost-kops-state-test
        - --private-dns
        - $(PRIVATE_DNS)
        - --keep-filestore-data
        - $(KEEP_FILESTORE_DATA)
        - --keep-database-data
        - $(KEEP_DATABASE_DATA)
        - --database
        - $(DATABASE)
```

the `KEEP_FILESTORE_DATA` is coming from a secret and it value is `false`

when the pod start we see it not set as false
```
time="2020-01-23T08:44:32Z" level=info msg="Starting Mattermost Provisioning Server" cluster-installation-supervisor=true cluster-resource-threshold=80 cluster-supervisor=true debug=true installation-supervisor=true instance=XOXOXOOX keep-database-data=true keep-filestore-data=true private-dns=XOXOXOOX state-store=XOXOXOOX store-version=0.11.0 use-existing-aws-resources=true working-directory=/mattermost-cloud
```

the proposed change set all flags to false and in the deployment, we just need to set the flag and not pass any value, like
```
containers:
      - args:
        - server
        - --debug
        - --keep-filestore-data
        - --keep-database-data
        - --state-store
        - mattermost-kops-state-test
        - --private-dns
        - $(PRIVATE_DNS)
        - --database
        - $(DATABASE)
```
which makes easier


cpanato:
found another way to inject the correct value.

But still, I think this is a nice to have.

PR #145, opened 2020-01-23
add Https status and write the header in the correct order

cpanato:
#### Summary
- add Https status for the missing API
- reorder the write of the headers otherwise, the header we are trying to add will not be added.

### DIFF
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -54,6 +54,7 @@ func handleGetClusters(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
 	outputJSON(c, w, clusters)
 }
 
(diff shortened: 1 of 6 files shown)


internal/api/webhook.go
```diff
@@ -43,8 +43,8 @@ func handleCreateWebhook(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
 	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
gabrieljackson:
```suggestion
	w.WriteHeader(http.StatusAccepted)
```

internal/api/cluster.go
```diff
@@ -54,6 +54,7 @@ func handleGetClusters(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
gabrieljackson:
```suggestion
	w.WriteHeader(http.StatusAccepted)
```

internal/api/webhook.go
cpanato: ahh my bad, overlooked
thanks for catching that!

PR #148, opened 2020-02-03
Prevent group deletion with installation members

gabrieljackson:
This corrects group deletion behavior so that only empty groups
can be deleted. Also, the installation filter is now able to filter
installations based on their group ID.

https://mattermost.atlassian.net/browse/MM-22252

### DIFF
--- a/internal/api/group_test.go
+++ b/internal/api/group_test.go
@@ -10,9 +10,9 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/mattermost/mattermost-cloud/internal/api"
-	"github.com/mattermost/mattermost-cloud/model"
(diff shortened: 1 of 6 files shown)


internal/api/group_test.go
```diff
@@ -351,13 +351,43 @@ func TestDeleteGroup(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	installation1, err := client.CreateInstallation(&model.CreateInstallationRequest{
+		OwnerID:  "owner",
+		Version:  "version",
+		DNS:      "dns.example.com",
+		Affinity: model.InstallationAffinityIsolated,
+	})
```
gigawhitlocks: nit: in most places we have an empty newline after closing a stanza; I would personally like to keep doing that, here and in a few other places.

PR #153, opened 2020-02-06
Always log Helm stdout to Info level

gigawhitlocks:
Today we hit an issue where Helm unhelpfully only provided `exit code 1` in our logs for debugging. @cpanato suggested that it would be better to see the Helm output at a log level higher than debug, and also it should be valuable to see the Helm output whether or not the process errors out or succeeds.

This PR puts that change into effect.


gigawhitlocks:
Yeah I have no problem with that long term, or ditching Helm long term

gabrieljackson:
Opened https://github.com/mattermost/mattermost-cloud/pull/155 as a continuation of this work.

PR #154, opened 2020-02-06
MM-22353 Add a migration to insert default values for UtilityMetadata

gigawhitlocks:
This change inserts an empty JSON structure into the Cluster table in the UtilityMetadata column to resolve the issues seen upgrading clusters in production.

To test this, I build HEAD~1 of the provisioner (database schema version .11), created a cluster, killed the provisioner, rebuilt with these changes, migrated the database from .11 to .13, restarted the provisioner, and was able to call `cloud cluster provision` against the existing cluster.

After provisioning, the utilities are pinned to stable by default. I was also able to successfully test pinning them at a specific version after the migration:
```
~/g/s/g/m/mattermost-cloud » cloud cluster provision --cluster 9q85a9ajojnu98osrrfycgj77a
```
after provision:
```
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 9q85a9ajojnu98osrrfycgj77a
{
    "desiredVersions": {
        "Prometheus": "",
        "Nginx": "",
        "Fluentbit": ""
    },
    "actualVersions": {
        "Prometheus": "10.4.0",
        "Nginx": "1.30.0",
        "Fluentbit": "2.8.7"
    }
}
```
Reprovision to pin versions:
```
~/g/s/g/m/mattermost-cloud » cloud cluster provision --cluster 9q85a9ajojnu98osrrfycgj77a --prometheus-version 10.4.0 --nginx-version 1.30.0 --fluentbit-version 2.8.7
~/g/s/g/m/mattermost-cloud » cloud cluster utilities --cluster 9q85a9ajojnu98osrrfycgj77a
{
    "desiredVersions": {
        "Prometheus": "10.4.0",
        "Nginx": "1.30.0",
        "Fluentbit": "2.8.7"
    },
    "actualVersions": {
        "Prometheus": "10.4.0",
        "Nginx": "1.30.0",
        "Fluentbit": "2.8.7"
    }
}
```

Lines 596-610 should protect manually transitioned clusters from being further altered by this migration, but should receive extra attention during this review as I am not 100% certain how to test that scenario.

### DIFF
--- a/internal/store/migrations.go
+++ b/internal/store/migrations.go
@@ -606,7 +606,7 @@ var migrations = []migration{
 
 		// if there are UtilityMetadata fields that are not empty, this
 		// table has already been manually migrated
-		if numRows > 1 {
+		if numRows > 0 {
 			return nil
 		}
 
--- a/internal/store/migrations.go
+++ b/internal/store/migrations.go
@@ -591,4 +591,29 @@ var migrations = []migration{
 
 		return nil
 	}},
+	{semver.MustParse("0.12.0"), semver.MustParse("0.13.0"), func(e execer) error {
+		// Assign default values to UtilityMetadata if values do not
+		// exist, otherwise, keep existing values
+		result, err := e.Exec(`SELECT * FROM Cluster WHERE UtilityMetadata != ''`)
+		if err != nil {
+			return err
+		}
+
+		numRows, err := result.RowsAffected()
+		if err != nil {
+			return err
+		}
+
+		// if there are UtilityMetadata fields that are not empty, this
+		// table has already been manually migrated
+		if numRows > 1 {
+			return nil
+		}
+
+		_, err = e.Exec(`
+				UPDATE Cluster
+				SET	UtilityMetadata = '{}';
+		 `)
+		return err
+	}},
 }
--- a/internal/store/migrations.go
+++ b/internal/store/migrations.go
@@ -591,28 +591,12 @@ var migrations = []migration{
 
 		return nil
 	}},
-	{semver.MustParse("0.12.0"), semver.MustParse("0.13.0"), func(e execer) error {
+	{semver.MustParse("0.12.0"), semver.MustParse("0.12.1"), func(e execer) error {
 		// Assign default values to UtilityMetadata if values do not
 		// exist, otherwise, keep existing values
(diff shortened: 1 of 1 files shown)


internal/store/migrations.go
```diff
@@ -591,4 +591,29 @@ var migrations = []migration{
 
 		return nil
 	}},
+	{semver.MustParse("0.12.0"), semver.MustParse("0.13.0"), func(e execer) error {
```
gabrieljackson: nit: this is probably the perfect use-case for a schema dot release
```suggestion
	{semver.MustParse("0.12.0"), semver.MustParse("0.12.1"), func(e execer) error {
```

internal/store/migrations.go
```diff
@@ -591,4 +591,29 @@ var migrations = []migration{
...
+		if err != nil {
+			return err
+		}
+
+		numRows, err := result.RowsAffected()
+		if err != nil {
+			return err
+		}
+
+		// if there are UtilityMetadata fields that are not empty, this
```
gabrieljackson: nit:
```suggestion
		// If there are UtilityMetadata fields that are not empty, this
```

internal/store/migrations.go
```diff
@@ -591,4 +591,29 @@ var migrations = []migration{
...
+		// table has already been manually migrated
+		if numRows > 0 {
+			return nil
+		}
```
gabrieljackson: This would work if every cluster has been migrated, but I am worried about a scenario where a portion of the clusters have utility metadata, but some don't.

To test this, build two clusters with the schema 11, then use latest master and schema 12 to `provision` one of the clusters. This should work and update the metadata field for that cluster. Then use this branch and update your schema to perform the migration. Check that the provisioned cluster metadata is intact and non-provisioned metadata is set to this migration's value.

gigawhitlocks: Yeah okay this does flush out a bug, I see where I had a gap in my testing, and I see how to fix it, so I am doing that now. Good catch, thanks.

gigawhitlocks:
Turns out this whole thing boils down to a single SQL statement.

gigawhitlocks:
I manually tested with your scenario :)

gigawhitlocks:
Just needs a sign-off from @stylianosrigas or @cpanato now :)

PR #156, opened 2020-02-11
MM-15369 Add API endpoint to allow changing version of Installation Groups

gigawhitlocks:
This change stores the desired Group version in the database, and exposes an API endpoint to do so. It does nothing else. 

I'm not sure if this resolves MM-15369 or not, as the story was originally written with the intent that the change would be put into effect on the Installations in question, so we can discuss what we want to do with the JIRA story for the sake of tracking.

### DIFF
--- a/internal/api/group.go
+++ b/internal/api/group.go
@@ -33,7 +33,7 @@ func handleChangeMattermostVersion(c *Context, w http.ResponseWriter, r *http.Re
 
 	groupID, ok := vars["group"]
 	if !ok {
-		c.Logger.Error("couldn't get group ID from request parameters")
+		c.Logger.Error("failed to get group ID from request parameters")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
@@ -41,7 +41,7 @@ func handleChangeMattermostVersion(c *Context, w http.ResponseWriter, r *http.Re
 
 	version, ok := vars["version"]
 	if !ok {
-		c.Logger.Error("couldn't get version from request parameters")
+		c.Logger.Error("failed to get version from request parameters")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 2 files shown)


gigawhitlocks:
As part of input validation, should the API endpoint throw an error if the version provided isn't SemVer?

gabrieljackson:
Probably not. Version is a value that can be any docker tag, including the short sha tags we create and use heavily for PR testing.

internal/api/group.go
```diff
@@ -21,6 +21,54 @@ func initGroup(apiRouter *mux.Router, context *Context) {
...
 	groupRouter.Handle("", addContext(handleUpdateGroup)).Methods("PUT")
 	groupRouter.Handle("", addContext(handleDeleteGroup)).Methods("DELETE")
+
+	groupRouter.Handle("/mattermost/{version:[A-Za-z0-9.]+}", addContext(handleChangeMattermostVersion)).Methods("PUT")
+}
+
+// handleChangeMattermostVersion responds to PUT
+// /api/group/{group}/mattermost/{version}, upgrading or downgrading
+// the installation to the Mattermost version embedded in the request.
+func handleChangeMattermostVersion(c *Context, w http.ResponseWriter, r *http.Request) {
```
gabrieljackson: nit: `handleChangeMattermostVersion` is a really generic name when we probably want something more like `handleChangeGroupMattermostVersion`

internal/api/group.go
```diff
@@ -21,6 +21,54 @@ func initGroup(apiRouter *mux.Router, context *Context) {
...
+	vars := mux.Vars(r)
+
+	groupID, ok := vars["group"]
+	if !ok {
+		c.Logger.Error("couldn't get group ID from request parameters")
```
gabrieljackson: nit: I would prefer our errors start with `failed to` like the most of the other API error messages, but not a big deal

gigawhitlocks: ah, sorry, you left a comment like this on a previous PR and I didn't understand the reason for the wording change so it didn't "stick" with me; uniform language is important and I will fix this before merge.

internal/api/group.go
gigawhitlocks: I think that's a great suggestion and actually I don't think it's a nitpick either, names are important!

PR #157, opened 2020-02-11
MM-22391 Run helm init during provision

gigawhitlocks:
To resolve the issues seen when reprovisioning a cluster with a provisioner whose pod has been restarted, the easiest / simplest / lowest-complexity approach is to simply run `helm init` during `provision` as well as during `create`, before other Helm operations.

Running `helm init` adds little execution time cost and is a no-op on a cluster where re-initializing the tool is unnecessary, so adding it to `UpgradeUtilityGroup`, which is called when `cloud cluster provision` is called, should ensure that Helm is initialized properly before `provision` is called.

### DIFF
--- a/internal/provisioner/utility_group.go
+++ b/internal/provisioner/utility_group.go
@@ -135,7 +135,7 @@ func (group utilityGroup) UpgradeUtilityGroup() error {
 	logger := group.provisioner.logger.WithField("helm-init", "UpgradeUtilityGroup")
 	err := helmInit(logger, group.kops)
 	if err != nil {
-		logger.WithError(err).Errorf("couldn't re-initialize Helm in the cluster")
+		logger.WithError(err).Error("couldn't re-initialize Helm in the cluster")
 	}
 
 	for _, utility := range group.utilities {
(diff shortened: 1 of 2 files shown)


internal/provisioner/utility_group.go
```diff
@@ -132,6 +132,12 @@ func (group utilityGroup) DestroyUtilityGroup() error {
 
 // UpgradeUtilityGroup reapplies the chart for the UtilityGroup. This will cause services to upgrade to a new version, if one is available.
 func (group utilityGroup) UpgradeUtilityGroup() error {
+	logger := group.provisioner.logger.WithField("helm-init", "UpgradeUtilityGroup")
+	err := helmInit(logger, group.kops)
+	if err != nil {
+		logger.WithError(err).Errorf("couldn't re-initialize Helm in the cluster")
```
gabrieljackson: nit:
```suggestion
		logger.WithError(err).Error("Couldn't re-initialize Helm in the cluster")
```

gigawhitlocks: I don't know why I'm constantly doing this :joy:

gabrieljackson: lol, me either. My linter complains about this stuff. Maybe you can turn that on for yours somehow?

gigawhitlocks: What linter are you using? `golint` and `gopls` complain about a number of things but not this.

gabrieljackson: Lately I have been using https://github.com/golangci/golangci-lint

PR #159, opened 2020-02-14
MM21928: adds ability to snapshot RDS database

gsagula:
#### Summary

This is the very first PR towards migration. It adds the ability to snapshot a AWS database. In addition to that, there is also a small enhancement done to tooling/aws and testing. We can now generate mock libraries for all AWS api in usage.

  
#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-21928


Makefile
```diff
@@ -21,8 +21,23 @@ GOFLAGS ?= $(GOFLAGS:)
 BUILD_TIME := $(shell date -u +%Y%m%d.%H%M%S)
 BUILD_HASH := $(shell git rev-parse HEAD)
 
+################################################################################
+
+AWS_SDK_URL := github.com/aws/aws-sdk-go
+AWS_SDK_VERSION := $(shell find go.mod -type f -exec cat {} + | grep ${AWS_SDK_URL} | cut -c28-)
```
gigawhitlocks: This is a tiny nit-pick but can we change the last part of these pipelines to `| awk '{print $NF}'`? It's ever-so-slightly less brittle, because it will always select the last column rather than using -c to count characters from the beginning.

Also, what is the behavior of this if the targetted Go module is not present?

Makefile
```diff
@@ -91,3 +106,21 @@ get-kubectl: ## Download kubectl only if it's not available. Used in the docker
 .PHONY: install
 install: build
 	go install ./...
+
+# Generate mocks from the interfaces.
+.PHONY: mocks
+mocks:
+	@if [ ! -f $(GOPATH)/bin/mockery ]; then \
+		echo "installing mockery ..." && \
+ 		env GO111MODULE=off $(GO) get -u github.com/vektra/mockery/.../; \
```
gigawhitlocks: I'm not sure that I want a `Makefile` to install dependencies onto my system. 1/5 but our other dependencies, e.g. Helm, are just listed on the README so that the developer can install and manage them him or herself.

In my opinion, I think it's probably OK to just assume `mockery` is installed in the `Makefile`.

internal/tools/aws/client_test.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+
+		DBName:               "mattermost",
+		DBUser:               "admin",
+		DBPassword:           "secret",
+		RDSParamGroupCluster: "mattermost-provisioner-rds-cluster-pg",
+		RDSParamGroup:        "mattermost-provisioner-rds-pg",
+		RDSAvailabilityZones: []string{"us-east-1a", "us-east-1b", "us-east-1c"},
+		GroupID:              "id-0000000000000000",
+		SecretString:         "{\"MasterUsername\":\"mmcloud\",\"MasterPassword\":\"oX5rWueZt6ynsijE9PHpUO0VUWSwWSxqXCaZw1dC\"}",
```
gigawhitlocks: These will be more readable with raw string literals (use backticks instead of quotes and then you do not need to escape all the quotes inside the string) https://golang.org/ref/spec#String_literals

gabrieljackson:
I will do a review asap, but can you add a `.gitattributes` file to mark the new generated mocks? Example from the operator: https://github.com/mattermost/mattermost-operator/blob/master/.gitattributes

It just helps github make the PRs a little more easier to review.

internal/tools/aws/client.go
```diff
@@ -51,6 +75,21 @@ type api interface {
...
+	return &Client{
+		api: &apiInterface{},
+
+		acm:            acm.New(sess),
+		ec2:            ec2.New(sess),
+		iam:            iam.New(sess),
+		rds:            rds.New(sess),
+		s3:             s3.New(sess),
+		route53:        route53.New(sess),
+		secretsManager: secretsmanager.New(sess),
```
gsagula: This could be done more efficiently if we could use construct at first time only. However, if I remember it correctly, there as general concern around session that should be able to pick up changes to the environment. If we can modify session in only one place, we could potentially use a singleton pattern for this client. Session is concurrent-safe for reading but not for writing.

internal/tools/aws/client.go
```diff
@@ -51,6 +75,21 @@ type api interface {
 	listTagsForCertificate(*acm.ACM, *acm.ListTagsForCertificateInput) (*acm.ListTagsForCertificateOutput, error)
 }
 
+// NewAWSClient returns a new AWS client.
+func NewAWSClient(sess *session.Session) *Client {
```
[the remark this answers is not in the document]
gsagula: My bad, I should had stated that I wanted to get a feedback in this part before changing all the rest. Regarding to your first part of the comment, client always needs a session and idealistic we should not be constructing  a new one every time that we make a call. Either client and or session should be initialized once and then passed to the varies parts of the system via context or something similar (idealistically). However, IRC there is a requirement which the application may need to pick up configuration changes from the environment (not sure). I think @gabrieljackson can comment on that.

internal/tools/aws/client.go
```diff
@@ -1,9 +1,21 @@
 package aws
 
 import (
+	"github.com/aws/aws-sdk-go/aws/session"
```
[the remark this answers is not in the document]
gsagula: Good point, I will change that.

internal/tools/aws/client.go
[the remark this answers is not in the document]
gabrieljackson: Regarding configuration changes being picked up dynamically, that was regarding using tags to lookup AWS resources and how we wouldn't just want to do that once and then store them all as we could make configuration changes to the account and the already-running provisioner would have stale data afterwards.

This is a bit different in that the changes to session would involve updating the provisioner directly, so I am sorta 0/5 on setting sessions once or as needed for AWS calls.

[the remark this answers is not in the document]
gabrieljackson: I am also a little confused about the intention of `NewAWSClient` as it relates to the original way to create a client: https://github.com/mattermost/mattermost-cloud/pull/159/files#diff-e359fe546a61f4fa6bd4af55be2c38e5L55

I understand that we will be doing multiple PRs to move to the new mocks+client, but perhaps you could outline how you envision it working.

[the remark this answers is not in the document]
gsagula: > This is a bit different in that the changes to session would involve updating the provisioner directly, so I am sorta 0/5 on setting sessions once or as needed for AWS calls.

I was referring to a conversation about setting default region in code vs let the session to grab it from the environment. We have two use-cases here that can change the design quite a bit: 

1) setting session once, it's more efficient since a new client and session don't need to be construct every time that we make a call to AWS. The drawback is that the provisioner needs to be restarted always when a configuration changes.

2) creating session on every call, as explained above, it's very inefficient but besides to that, it requires some weird abstractions such are `getClient()` so the code can be tested with a mocked implementation of the api interface. This also makes more difficult for using the interfaces already provided by AWS. If we decided that we really need this use case, the current way how we construct session will need to change since it is currently using [deprecated code](https://github.com/aws/aws-sdk-go/blob/3da93ec4fb4bf582b00a1dc0c5ee87e16d6d446e/aws/session/session.go#L71).

If the system doesn't need to load configuration dynamically without being restarted I think session and client should be constructed once and passed around throughout the whole life-time of the system.

> I am also a little confused about the intention of NewAWSClient as it relates to the original way to create a client: https://github.com/mattermost/mattermost-cloud/pull/159/files#diff-e359fe546a61f4fa6bd4af55be2c38e5L55

I didn't want to remove the old way just yet because of testing and all the call sites that still use the old api interface. The original way to create client requires to manually maintain an implement an interface which the new way doesn't. The new way uses the SDK interfaces to generate mocks from a proper mocking framework. If we decided to go this way I will change all the call sites in this PR and fix the tests that already exist. It's also very likely that I will turn client and session into a singleton.

gsagula:
@gabrieljackson I pushed some modifications based on your comments. PTAL

Changes were made mostly based on this https://github.com/mattermost/mattermost-cloud/pull/159#discussion_r381338971

.gitattributes
```diff
@@ -0,0 +1 @@
+internal/mocks/**/*  linguist-generated
```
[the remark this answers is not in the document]
gsagula: I think we nee to mer it first:
```
$ git check-attr linguist-generated internal/mocks/**/*
internal/mocks/aws-sdk/ACMAPI.go: linguist-generated: set
internal/mocks/aws-sdk/EC2API.go: linguist-generated: set
internal/mocks/aws-sdk/IAMAPI.go: linguist-generated: set
internal/mocks/aws-sdk/RDSAPI.go: linguist-generated: set
internal/mocks/aws-sdk/Route53API.go: linguist-generated: set
internal/mocks/aws-sdk/S3API.go: linguist-generated: set
internal/mocks/aws-sdk/SecretsManagerAPI.go: linguist-generated: set
internal/mocks/aws-tools/AWS.go: linguist-generated: set
internal/mocks/logger/Ext1FieldLogger.go: linguist-generated: set
internal/mocks/logger/FieldLogger.go: linguist-generated: set
internal/mocks/logger/Formatter.go: linguist-generated: set
internal/mocks/logger/Hook.go: linguist-generated: set
internal/mocks/logger/StdLogger.go: linguist-generated: set
internal/mocks/model/Database.go: linguist-generated: set
internal/mocks/model/Filestore.go: linguist-generated: set
internal/mocks/model/InstallationDatabaseStoreInterface.go: linguist-generated: set
```

internal/tools/aws/client.go
```diff
@@ -2,12 +2,45 @@ package aws
...
 )
 
+// AWSClient is a singleton instance of an AWS client.
+var AWSClient *Client
+
+func init() {
+	// Create a singleton instance of an AWS session.
+	sess, err := NewAWSSessionWithLogger(log.WithField("tools-aws", "client"))
+	if err != nil {
+		log.Fatalf("failed to initialize AWS session: %s", err.Error())
```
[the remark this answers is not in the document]
gsagula: Very good point. Based on our discussion, I changed it to allow multiple clients and also not crash in case of session failure. If a session fails the client still instantiates but obviously any call made to aws will also fail. Still under the session failure scenario, the instance will attempt to get a heath session on every api call, if it succeed, the client instance will construct all clients with the healthy session and cache it.

internal/tools/aws/client_test.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func TestClientSuite(t *testing.T) {
+	suite.Run(t, new(ClientTestSuite))
+}
+
+type Mocks struct {
+	AWS   *Client
+	API   *testlib.AWSMockedAPI
+	LOG   *testlib.MockedFieldLogger
```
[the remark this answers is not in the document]
gsagula: If you inject a logger instance you can't assert that your log messages are correct. For example:
```
        a.Mocks.LOG.WithFieldArgs("security-group-ids", a.GroupID).Once()
	a.Mocks.LOG.WithFieldString("db-subnet-group-name", SubnetGroupName(a.VPCa)).Once()
	a.Mocks.LOG.WithFieldString("db-cluster-name", CloudID(a.InstallationA.ID)).Once()
	a.Mocks.LOG.WithFieldString("secret-name", RDSSecretName(CloudID(a.InstallationA.ID))).Once()
	a.Mocks.LOG.WithFieldString("db-cluster-name", CloudID(a.InstallationA.ID)).Once()
```

internal/tools/aws/client_test.go
```diff
@@ -0,0 +1,184 @@
...
+		ResourceID:           "WSxqXCaZw1dC",
+
+		EndpointsA: []string{"example1.mattermost.com", "example2.mattermost.com"},
+		EndpointsB: []string{"example1.mattermost.com"},
+		DNSNameA:   "mattermost.com",
+		DNSNameB:   "mattermost-cloud.com",
+	}
+}
+
+// This will take care of reseting the mocks on every run. Any new mocked library should be added here.
```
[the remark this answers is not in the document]
gsagula: Thanks!

gsagula:
@gabrieljackson Thanks for the review in this one. I made the modifications based on your feedback and our team discussion. Now, these code should support as many AWS clients we need. I also modified the client in the way that it never crashes when session fails. Please let me know if you want me to go through all the changes with you.

internal/tools/aws/client_test.go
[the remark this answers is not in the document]
gabrieljackson: Ah, makes sense. Thanks!

internal/tools/aws/client.go
```diff
@@ -73,6 +61,45 @@ type Client struct {
...
+	store   model.InstallationDatabaseStoreInterface
+	service *Service
+	config  *aws.Config
+	mux     *sync.Mutex
+}
+
+// Service contructs an AWS session if not yet successfully done and returns AWS clients.
+func (c *Client) Service() *Service {
+	if c.service == nil {
+		sess, err := NewAWSSessionWithLogger(c.config, log.WithField("tools-aws", "client"))
```
gabrieljackson: Does it makes sense to pass in a logger like we do most other places?

internal/tools/aws/client.go
```diff
@@ -73,6 +61,45 @@ type Client struct {
...
+		if err != nil {
+			log.Errorf("failed to initialize AWS session: %s", err.Error())
```
gabrieljackson: nit:
```suggestion
			log.WithError(err).Error("failed to initialize AWS session")
```

internal/tools/aws/client.go
gsagula: 👍

internal/tools/aws/client.go
gsagula: 👍

gsagula:
My initial idea was to use the official mock library and mockgen but then I saw that we've already been using mockery for other prokects. I can change it in a follow up PR.

gsagula:
@gigawhitlocks When you get a chance, could you give another pass on this one, please? Thanks!

internal/tools/aws/client.go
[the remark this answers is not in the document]
gigawhitlocks: This was resolved with no change, but you did say you were open to this change, so I guess this got overlooked.

internal/tools/aws/client_test.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func TestClientSuite(t *testing.T) {
+	suite.Run(t, new(ClientTestSuite))
+}
+
+type Mocks struct {
+	AWS   *Client
+	API   *testlib.AWSMockedAPI
+	LOG   *testlib.MockedFieldLogger
```
gigawhitlocks: LOG should be Log to follow Go style, no all-caps variables unless they are acronyms

internal/tools/aws/cluster_management.go
```diff
@@ -57,7 +57,7 @@ func getClusterResourcesForVPC(vpcID string) (ClusterResources, error) {
 		Values: []*string{aws.String("private")},
 	})
 
-	privateSubnets, err := GetSubnetsWithFilters(privateSubnetFilter)
+	privateSubnets, err := a.GetSubnetsWithFilters(privateSubnetFilter, logger)
```
gigawhitlocks: Is this pattern preferable to making `logger` a member of `Client` so that it can be accessed inside of these methods instead of being passed in each time because of the ability to add context to the loggers when passing them in?

If so, should we add fields to these loggers when passing them in? Otherwise it seems like a lot of easily eliminated repetition to me.

internal/tools/aws/session_test.go
```diff
@@ -0,0 +1,13 @@
+package aws
+
+// func (a *AWSTestSuite) TestSingletonSession() {
```
gigawhitlocks: Please don't check in commented-out code unless there's a very good reason. There's always version control history to fetch old code; leave a comment here if needed.

gsagula: That was an accident haha

internal/tools/aws/cluster_management.go
gsagula: That seems to be a common pattern in this repo. In fact, somewhere in this PR (if not in this exactly case) I was told to pass logger in.

internal/tools/aws/client.go
```diff
@@ -24,43 +39,72 @@ type AWS interface {
...
-	tagResource(*ec2.EC2, *ec2.CreateTagsInput) (*ec2.CreateTagsOutput, error)
-	untagResource(*ec2.EC2, *ec2.DeleteTagsInput) (*ec2.DeleteTagsOutput, error)
-	describeImages(svc *ec2.EC2, input *ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error)
-
-	getACMClient() (*acm.ACM, error)
-	listCertificates(*acm.ACM, *acm.ListCertificatesInput) (*acm.ListCertificatesOutput, error)
-	listTagsForCertificate(*acm.ACM, *acm.ListTagsForCertificateInput) (*acm.ListTagsForCertificateOutput, error)
-}
+// Service contructs an AWS session if not yet successfully done and returns AWS clients.
+func (c *Client) Service(logger log.FieldLogger) *Service {
```
gigawhitlocks: Is there a reason that `logger` is not part of the `Client` struct, and is instead passed into this method every time? This pattern creates a lot of repetition in other files, where I do not see the extra power gained by including it as an argument utilized, which simply makes calls to this method more verbose. I think we should either take advantage of the call sites of this method by providing `logger` arguments that add more context from those call sites or eliminate this argument and set a `logger` per `Client` that can be manually overridden if necessary to add more context since we are not doing that in the majority of call sites of this method.

gsagula: https://github.com/mattermost/mattermost-cloud/pull/159#discussion_r389662499

internal/tools/aws/client.go
```diff
@@ -24,43 +39,72 @@ type AWS interface {
...
+	if c.service == nil {
+		sess, err := NewAWSSessionWithLogger(c.config, logger)
```
gigawhitlocks: Service() must be called with `logger` every time, however then `session` is cached in the `Client` object, which means that subsequent calls with a different `logger` will return a `*Service` with the logger initialized the last time, ignoring the one passed in. This would be unexpected behavior. Either the logger should change when it is passed in, or the logger should be part of the Client as outlined in the other comment on this method.

internal/tools/aws/acm_test.go
```diff
@@ -1,55 +1,3 @@
...
-	_, err := a.GetCertificateSummaryByTag("banana", testRoute53TagValue)
-	assert.Error(t, err)
-}
-
-func TestGetCertificateByTagEmptyValue(t *testing.T) {
-	a := Client{api: &mockAPI{}}
-	_, err := a.GetCertificateSummaryByTag(testParsedRoute53TagKey, "")
-	assert.Error(t, err)
-}
+// TODO(gsagula): add tests for ACM.
```
gigawhitlocks: What is needed to convert these tests to use the new mocking framework?

internal/tools/aws/client.go
gigawhitlocks: See my other comment, then, the behavior here is incorrect if the logger is passed in. Only the first one supplied per `Client` is actually effective.

I'd like to see extra human-friendly context added to the logger that's passed in, if we are going to the trouble to pass the logger in each time, too, to aid in debugging, since we create a new logger for each call site. Otherwise we're repeating ourselves for no reason, but if there's unique context added at each call site, that would be a good reason to pass the logger in as an argument and will aid with debugging. 

@gabrieljackson can weigh in if I'm totally off base here, since I'm still getting used to this pattern myself.

gigawhitlocks:
In the future please open 2 PRs for 2 features (in this case it would be 1 for mocks and 1 for the snapshot feature) because disentangling the features makes reviewing a bit tougher. It would also be better to have these features as separate commits on `master` for the sake of `git bisect` in the future if a bug is found in the code, but I don't think that will be possible without extricating the code and opening another PR for the mocks, which I don't expect to be done this late in the review process. Just something to keep in mind for the future.

Other than that and a few small things this is looking good!

internal/tools/aws/client.go
gsagula: > Service() must be called with logger every time, however after the first call session is cached in the Client object, which means that subsequent calls with a different logger will return a *Service with the logger initialized the first time, ignoring the one passed in. This would be unexpected behavior. 

That's not how the above code works. Client may be cached at first call if `NewAWSSessionWithLogger()` does not return an error, otherwise it will return a new service without logger. I don't see how this will produce an unexpected behaviour. 

> Either the logger should change when it is passed in, or the logger should be part of the Client as outlined in the other comment on this method.

I agree in part with that, but I was asked to revert this exactly suggested behaviour https://github.com/mattermost/mattermost-cloud/pull/159#discussion_r389662499

What do you mean by `logger should change when it is passed in`?

gigawhitlocks: NewAWSSessionWithLogger is only called if `c.service` is `nil`, otherwise `c.service` is `nil`, which will include the logger from the last time `c.Service()` was called.

gigawhitlocks: So if you reuse the client, and call `c.Service(logger)`, `logger` may be the old value passed the first time `c.Service()` is called, instead of the new provided value, which I can't imagine is the expected or intended behavior.

gigawhitlocks: So for instance when `c.Service(logger)` is called here https://github.com/mattermost/mattermost-cloud/pull/159/files/1237b244124d3b3fc991f2cb3638d9a90779daf8#diff-4d22c3a7103d8871c0b8cf1a5c67db3cR54 , a cached `logger` from `c` will be used if `c.service` exists, and if not, the passed-in `logger` will be used, when I believe the behavior we desire is for the new logger to always be used when it's passed in. Otherwise, it's impossible to add debugging context at call time, and that's the point of always passing in the logger, I believe.

It's not clear that this is the case in the current version of the code because although `logger` is passed in every time, it is done without adding new debugging context (e.g. `logger.WithFields`) so the functionality appears equivalent.

internal/tools/aws/database.go
```diff
@@ -41,11 +46,45 @@ func (d *RDSDatabase) Provision(store model.InstallationDatabaseStoreInterface,
 
 // Teardown removes all AWS resources related to a RDS database.
 func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
-	err := rdsDatabaseTeardown(d.installationID, keepData, logger)
+	logger.Info("Tearing down AWS RDS database")
+	awsID := CloudID(d.installationID)
+
```
gigawhitlocks: add:
`logger = logger.WithField("db-cluster-name", awsID)` here and then remove the WithField portion on lines 62 and 64 for cleanup and additional context when `logger ` is passed to other functions on lines 52 and 58

gsagula:
> In the future please open 2 PRs for 2 features (in this case it would be 1 for mocks and 1 for the snapshot feature) because disentangling the features makes reviewing a bit tougher. It would also be better to have these features as separate commits on `master` for the sake of `git bisect` in the future if a bug is found in the code, but I don't think that will be possible without extricating the code and opening another PR for the mocks, which I don't expect to be done this late in the review process. Just something to keep in mind for the future.
> 
> Other than that and a few small things this is looking good!

Thanks for the review but I'm a bit confused here. In the same PR, you initially asked me why I didn't change all the call sites related to the client change. I'm having a hard time to see how would be possible to address that without breaking all the tests. Besides to that, code had to be refactored in order to support DI. I will keep that in mind in the future but ideally we should be more careful on shipping code without proper tests or ready for it.

internal/tools/aws/database.go
```diff
@@ -41,11 +46,45 @@ func (d *RDSDatabase) Provision(store model.InstallationDatabaseStoreInterface,
...
+	if err != nil {
+		return errors.Wrap(err, "unable to delete RDS secret")
+	}
+
+	if !keepData {
+		err = d.client.rdsEnsureDBClusterDeleted(awsID, logger)
+		if err != nil {
+			return errors.Wrap(err, "unable to delete RDS DB cluster")
+		}
+		logger.WithField("db-cluster-name", awsID).Debug("AWS RDS DB cluster deleted")
```
gigawhitlocks: return after this log message and eliminate `else`

gigawhitlocks: Or flip the conditional and short-circuit (`if keepData { log; return nil}`)  to eliminate the `else`, up to you.

internal/tools/aws/acm_test.go
gsagula: I've already did it in a separated PR.

internal/tools/aws/client.go
gsagula: I see what you mean, so should we go back to what was initially? @gabrieljackson Any objection here since you requested this change? https://github.com/mattermost/mattermost-cloud/pull/159#discussion_r389662499

internal/tools/aws/filestore.go
```diff
@@ -35,19 +37,39 @@ func (f *S3Filestore) Provision(logger log.FieldLogger) error {
...
 	}
 
+	if keepData {
+		logger.WithField("s3-bucket-name", awsID).Info("AWS S3 bucket was left intact due to the keep-data setting of this server")
+		return nil
+	}
+
+	err = f.awsClient.s3EnsureBucketDeleted(awsID, logger)
+	if err != nil {
+		return errors.Wrap(err, "unable to teardown AWS S3 filestore")
```
gigawhitlocks: can we change the language in this error to "unable to ensure AWS S3 filestore was deleted" or something similar, just so that it's not identical to the error on line 46?

internal/tools/aws/database.go
gsagula: This code was there before but I will change it anyway.

internal/tools/aws/filestore.go
gsagula: Same as above, code was already there but doesn't hurt to improve it.

gigawhitlocks:
> Thanks for the review but I'm a bit confused here. In the same PR, you initially asked me why I didn't change all the call sites related to the client change. I'm having a hard time to see how would be possible to address that without breaking all the tests. Besides to that, code had to be refactored in order to support DI. I will keep that in mind in the future but ideally we should be more careful on shipping code without proper tests or ready for it.

I was suggesting that the change to add a mocking tool and generating mocks could have been done in a separate PR from the one that actually adds the new feature. Perhaps it could not be done in separate PRs. I agree that we shouldn't ship changes without proper tests.

internal/tools/aws/client.go
gigawhitlocks: No, I think if we want `logger` to be passed in then:
```
if c.service != nil {
c.service.logger = logger
}
```

Right? Otherwise passing in a logger only does something the first time `Session()` is called. If that's the desired behavior @gabrieljackson will have to explain to me why that's desired, I suppose. It doesn't make sense to me why we would ignore the `logger` argument in cases where `c.service` is already initialized.

internal/tools/aws/cluster_management.go
gigawhitlocks: Can we change `logger` to `logger.WithFields()` and then add a field to add debugging context at the call sites, to take advantage of passing the logger in each time?

internal/tools/aws/client.go
gsagula: The general idea here is that session may return error but we don't want to panic. Instead, we give a dummy client to the caller which will obviously return an error. This allows the system to keep trying call AWS until a healthy session is constructed. In the case of a failed session, logger is functionally irrelevant anyway. I agree that ignore logger in the case of a cached service but it was the best that I could come up based on the change request previously made. Not sure if this answers your question.

gigawhitlocks: It does not. My concern is not about an error case.
Say a user of this method has a `*Client` `c`, and some `logger` in their current scope.
The user calls the method like so:
`c.Session(logger.WithField("foo", "bar")).Something()`
Then, in the same scope, the user calls the method like this:
`c.Session(logger.WithField("bar", "baz")).Something()`
The second call to `Something()` will be called with `c.session` containing field `foo=bar`, and the `c.session` containing fields `bar=baz` will be silently ignored, which is surprising. If this is the desired behavior, I don't understand why that is the case.

internal/tools/aws/database.go
```diff
@@ -49,21 +49,24 @@ func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
...
 
-	if !keepData {
-		err = d.client.rdsEnsureDBClusterDeleted(awsID, logger)
-		if err != nil {
-			return errors.Wrap(err, "unable to delete RDS DB cluster")
-		}
-		logger.WithField("db-cluster-name", awsID).Debug("AWS RDS DB cluster deleted")
-	} else {
+	if keepData {
 		logger.WithField("db-cluster-name", awsID).Info("AWS RDS DB cluster was left intact due to the keep-data setting of this server")
```
gigawhitlocks: WithField() is redundant here, now, because we add it on line 52. You can just do `logger.Info(...`

internal/tools/aws/database.go
```diff
@@ -49,21 +49,24 @@ func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
...
+		return nil
+	}
+
+	err = d.client.rdsEnsureDBClusterDeleted(awsID, logger)
+	if err != nil {
+		return errors.Wrap(err, "unable to delete RDS DB cluster")
 	}
 
+	logger.WithField("db-cluster-name", awsID).Debug("AWS RDS DB cluster deleted")
```
gigawhitlocks: same here, `WithField` is redundant now

internal/tools/aws/database.go
gsagula: Ah right, forgot to change it. Good catch!

gabrieljackson:
Alright, there are a few different topics being discussed that I will try to weigh in on once here.

It definitely would have been preferable to have two PRs for this change: one for the mock and one for the RDS snapshot feature. That said, as already mentioned, it doesn't make sense to change that now. Just something for next time.

Regarding the logger:
1. iirc, I was specifically asking to not use the root shared logrus logger object directly.
2. Now, that leaves us two options like was discussed. One is we give a logger to the AWS client and use that and one is to give all/some of the AWS functions a logger in the signature. I wasn't specifically asking for either for this PR, but as was discussed, there are pros/cons to both. We add a lot of structured logger fields as we progress in code and pass the logger down. This is a very good/useful thing and will help us a ton in the long run. By passing a logger in directly, you can retain all of the context and fields of that logger when we then act on something in AWS. This is the pro here and the con is it is clunky to pass loggers all the time. Traditionally, we pass the logger in most functions that may need to log, but different structs that play important roles are also allowed to store a logger for their own purposes. I can see the AWS library being a combination of both a client-owned logger and a passed in logger depending on the situation. That said, I am totally 4/5 of us testing things and seeing how it goes. It is easy enough to change and bit outside of the scope of this PR.

gsagula:
> Alright, there are a few different topics being discussed that I will try to weigh in on once here.
> 
> It definitely would have been preferable to have two PRs for this change: one for the mock and one for the RDS snapshot feature. That said, as already mentioned, it doesn't make sense to change that now. Just something for next time.
> 
> Regarding the logger:
> 
> 1. iirc, I was specifically asking to not use the root shared logrus logger object directly.
> 2. Now, that leaves us two options like was discussed. One is we give a logger to the AWS client and use that and one is to give all/some of the AWS functions a logger in the signature. I wasn't specifically asking for either for this PR, but as was discussed, there are pros/cons to both. We add a lot of structured logger fields as we progress in code and pass the logger down. This is a very good/useful thing and will help us a ton in the long run. By passing a logger in directly, you can retain all of the context and fields of that logger when we then act on something in AWS. This is the pro here and the con is it is clunky to pass loggers all the time. Traditionally, we pass the logger in most functions that may need to log, but different structs that play important roles are also allowed to store a logger for their own purposes. I can see the AWS library being a combination of both a client-owned logger and a passed in logger depending on the situation. That said, I am totally 4/5 of us testing things and seeing how it goes. It is easy enough to change and bit outside of the scope of this PR.
> 
> Edit:
> A few additional thoughts on passing in loggers. It is true that one of the greatest benefits of passing them in is that additional fields can then be appended for better structured logging, but keep in mind that even if no fields are given, it is highly likely that there will already be more fields than those of a base client-owned logger. For instance, the AWS client we use throughout most of the code is made very early on in server startup. Later, the supervisors are using their own loggers and adding fields before doing something like claiming a VPC. By the time the VPC claim happens, there will be more fields in that passed in logger. Just something to keep in mind.

Thanks for the detailed response, I will go ahead and move logger into the client struct as it was previously done. This logger that is passed to the AWS handler is only effective when the application runs in production. This will at least eliminate the unexpected behaviour well pointed out by @gigawhitlocks .

gsagula:
@gigawhitlocks Mind take another look when you get a chance, please?

internal/tools/aws/session.go
```diff
@@ -7,7 +7,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// NewAWSSessionWithLogger initialized a singleton AWS session instance with logging handler. This method should be called first in the code.
+// NewAWSSessionWithLogger initialized an AWS session instance with logging handler for debuging only.
```
gigawhitlocks: nit/type: "initialized" should be "initializes"

not a big deal :P

PR #160, opened 2020-02-14
MM-22503 Add MattermostEnv to Group objects

gigawhitlocks:
Adds a member MattermostEnv with a new type to Group objects and includes JSON marshaling/unmarshaling when going to/from the database.

@gabrieljackson this code is the same as it was when I linked you my WIP branch, just with the commits cleaned up and the tests for Groups updated to manually check each field value instead of doing a require.Equal() on each object, which causes tests to fail when a struct is read into and out of the database again, since the address of that object necessarily changes.

Resolves MM-22503


internal/store/group.go
```diff
@@ -10,15 +10,42 @@ import (
...
 
+type rawGroup struct {
+	ID            string
+	Name          string
+	Description   string
+	Version       string
+	CreateAt      int64
+	DeleteAt      int64
+	MattermostEnv []byte
+}
```
gabrieljackson: With a slightly different tactic, you can just embed the group struct directly here. The benefit of this is that you don't need to remember to update this when adding new group values. Also, tests pass when doing it this way without modification.

Example: https://github.com/mattermost/mattermost-cloud/pull/161/files#diff-f9b9c67a53c130978d472f7bb6412687R24-R27

internal/store/group_test.go
```diff
@@ -1,22 +1,94 @@
...
+				compareGroups(t, g1, g2)
+				break
+			}
+		}
+		if !found {
+			t.Fatal("groups of lists did not match", list1, list2)
+			continue
+		}
+	}
+}
```
gabrieljackson: As mentioned above, tests should pass without the need for this logic. Comparing manually means we will have to make sure we don't forgot to check new values as they are added to groups.

internal/store/migrations.go
```diff
@@ -600,4 +600,16 @@ var migrations = []migration{
 		 `)
 		return err
 	}},
+	{semver.MustParse("0.12.1"), semver.MustParse("0.13.0"), func(e execer) error {
+		// first rename "Group" to Group because those quotes are an accident
```
gabrieljackson: Old comment?

internal/store/group.go
gigawhitlocks: Hmm, I thought there was some reason that I didn't just embed. Maybe not. I'll give this a shot, thanks.

internal/store/migrations.go
gigawhitlocks: Yes, whoops

internal/store/group_test.go
gigawhitlocks: I don't this this is right, `assert.Equal` insists that pointer addresses are identical, which obviously we get new addresses when we write an object to the database and then read it out into a new object. 

I recognize the drawbacks of checking each value manually, but I wasn't able to see with our testing library a way to indicate that pointers should be dereferenced and values compared in a generic way, short of using reflection.

gigawhitlocks:
> seems to be missing any CLI option to set the env value for groups

Didn't realize we wanted CLI as part of this change. I'll add that today.

internal/store/group.go
gigawhitlocks: Ah I see, it was a silly naming problem that was preventing me from doing this in the first place.

gigawhitlocks:
This change is now rebased on top of https://github.com/mattermost/mattermost-cloud/pull/161 and should not be merged until after that is merged. The history of this change will also probably need to be squashed at that time. That's why the Do Not Merge label has been applied.

The diff between my change and @gabrieljackson's is available here https://github.com/mattermost/mattermost-cloud/compare/installation-update-overhaul...gigawhitlocks:group-mattermostEnv in case the PR diff is hard to read until his change is merged to `master`.

internal/store/group_test.go
```diff
@@ -1,22 +1,69 @@
...
+	mattermostEnv := model.EnvVarMap{
+		"Var1": model.EnvVar{
+			Value: "Var1Value",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					APIVersion: "1",
+					FieldPath:  "some/path/neat",
+				},
+
+				// TODO isw 2/17/2020
```
cpanato: what is `isw`?

gigawhitlocks: My initials ;)

model/group_test.go
```diff
@@ -57,6 +58,12 @@ func TestGroupFromReader(t *testing.T) {
 			Version:     "version",
 			CreateAt:    10,
 			DeleteAt:    20,
+			MattermostEnv: EnvVarMap{
```
cpanato: do we have test cases that the Mattermost env is nil or in a bad shape?

internal/store/group_test.go
cpanato: nice!
can you do something like
```
TODO: (ian)
```
or 
`TODO: (gigawhitlocks)`

to make explicit otherwise i will ask you again in a day, i will forgot this :joy:

cpanato:
can we have some documentation as well in how the group works? 
like:
- how to create a group
- how to update a group
- how to add installations to a group
- how to remove a installation from a group
....

I know this is not part of this PR we can make it in a followup

cc @stylianosrigas

internal/store/group_test.go
gigawhitlocks: Yes absolutely! I completely understand

model/group_test.go
gigawhitlocks: Now we do :)

gigawhitlocks: (and we already had them in the `store` package, as I only added the MattermostEnv to 1 of the groups that gets tested there, so most of the test cases have MattermostEnv as `nil` actually)

gigawhitlocks:
Also I manually tested creating a group with the `master` code and then updating the provisioner and doing `group list` to make sure we did not see a panic on the old group. The group created prior to the upgrade shows a env var map of simply `{}` in this case, tested against Postgres.

internal/api/group.go
```diff
@@ -110,10 +110,18 @@ func handleCreateGroup(c *Context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = createGroupRequest.MattermostEnv.Validate()
```
gabrieljackson: nit: would like to see all validation move to a `Validate` method for `createGroupRequest` like we do for other requests, but we can do this in a follow-up.

internal/api/group_test.go
```diff
@@ -240,16 +240,18 @@ func TestCreateGroup(t *testing.T) {
 
 	t.Run("valid", func(t *testing.T) {
 		group, err := client.CreateGroup(&model.CreateGroupRequest{
-			Name:        "name",
-			Description: "description",
-			Version:     "version",
+			Name:          "name",
+			Description:   "description",
+			Version:       "version",
+			MattermostEnv: model.EnvVarMap{"foo": model.EnvVar{Value: "bar"}},
```
gabrieljackson: nit: the upgrade group test seems a little less brittle with the way you set the env var once and used the variable to compare later.

internal/api/group_test.go
```diff
@@ -355,14 +360,16 @@ func TestUpdateGroup(t *testing.T) {
...
 		err = client.UpdateGroup(&model.PatchGroupRequest{
-			ID:          group1.ID,
-			Name:        sToP("name2"),
-			Description: sToP("description2"),
-			Version:     sToP("version2"),
+			ID:            group1.ID,
+			Name:          sToP("name2"),
+			Description:   sToP("description2"),
+			Version:       sToP("version2"),
+			MattermostEnv: model.EnvVarMap{"bar": model.EnvVar{Value: "baz"}},
```
gabrieljackson: nit: same here

internal/store/group_test.go
```diff
@@ -1,22 +1,69 @@
 package store
 
 import (
+	corev1 "k8s.io/api/core/v1"
```
gabrieljackson: nit: strange placement above standard library imports

internal/store/group_test.go
```diff
@@ -1,22 +1,69 @@
...
+	mattermostEnv := model.EnvVarMap{
+		"Var1": model.EnvVar{
+			Value: "Var1Value",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					APIVersion: "1",
+					FieldPath:  "some/path/neat",
+				},
+
+				// TODO @gigawhitlocks 2/17/2020
```
gabrieljackson: Thanks for the todo like we discussed. 👍

internal/store/migrations.go
```diff
@@ -606,6 +606,14 @@ var migrations = []migration{
 				ALTER TABLE Installation
 				ADD COLUMN MattermostEnvRaw BYTEA NULL;
 		`)
+		return err
+	}},
+	{semver.MustParse("0.13.0"), semver.MustParse("0.14.0"), func(e execer) error {
+		// Add Mattermost Env column for groups.
+		_, err := e.Exec(`
+						ALTER TABLE "Group"
+						ADD COLUMN MattermostEnvRaw BYTEA NULL;
```
gabrieljackson: nit: seems to be overly indented

model/env.go
```diff
@@ -49,3 +51,25 @@ func (em *EnvVarMap) ToEnvList() []corev1.EnvVar {
...
+
+// ToJSON converts the EnvVarMap to a JSON object represented as a
+// []byte
+func (em *EnvVarMap) ToJSON() ([]byte, error) {
+	b, err := json.Marshal(em)
+	if err != nil {
+		return []byte(nil), err
+	}
+
+	return b, nil
```
gabrieljackson: `json.Marshal()` returns the same thing as the parent function here. I would either wrap the error if there is one or just return the marshal call directly.

model/group_request.go
```diff
@@ -70,6 +72,11 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 		group.Version = *p.Version
 	}
 
+	if len(p.MattermostEnv) != 0 {
```
gabrieljackson: With this check, are we unable to unset all the env vars if we want? It seems like that may be the case.

internal/store/migrations.go
gigawhitlocks: 😬

model/env.go
gigawhitlocks: 🤔 Yeah, you're right. I think for some reason I was thinking just returning the call to `json.Marshal` directly was in conflict with the project style? Idk, good catch.

model/group_request.go
gigawhitlocks: No, it is the case that we are unable to unset all env vars with this check. This is a logical error, and instead on this line I should be checking `if p.MattermostEnv != nil`. This would mean that we can unset all env vars by providing an empty map, however an uninitialized map will be a no-op.

internal/api/group.go
gigawhitlocks: Easy enough, done.

model/env.go
gigawhitlocks: Might even just want to get rid of this method and call `json.Marshal` directly at the call site? I'm unsure.

gabrieljackson: Yeah, that was one of the reasons I originally didn't make this method, but I am totally fine keeping it.

model/group_request.go
```diff
@@ -90,6 +90,17 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 	return applied
 }
 
+// Validate validates the values of a group patch request
+func (p *PatchGroupRequest) Validate() error {
+
```
gabrieljackson: nit: extra newline

gabrieljackson: nit: extra newline

gabrieljackson:
Putting the No Not Merge label on for commit message cleanup after the infra review is done.

PR #161, opened 2020-02-16
Add Mattermost Env for Installations

gabrieljackson:
Due to changes that were made in recent operator versions, this
includes a few key alterations to the cloud server:
 - Add Mattermost Env configuration to the installation object.
   Both Value and ValueFrom are supported which enables basic env
   var value setting along with more-advanced values such as those
   from kubernetes secrets. Currently, the installation CLI only
   supports basic values. The missing ValueFrom support will be
   added later.
 - Update Go modules with replacements to handle operator changes
 - Renamed instances of upgrade installation to update installation.
   This was done to better reflect what installation changes are
   happening. Ever since being able to set and change the license
   value was introduced, this functionality is not specific to
   versions or upgrades.
 - Removed InstallationStateUpgradeInProgress as a valid state for
   updating an installation. The behavior of updating in this state
   is unclear, so it will be re-added later if possible.

https://mattermost.atlassian.net/browse/MM-22502


gabrieljackson:
Two notes:

1. Sorry it got so big. Most of the code is new tests.
2. Due to the change on states from "upgrade" to "update", you will want to make sure that no installations are upgrading before running this version of the code.

internal/store/installation.go
```diff
@@ -127,19 +174,25 @@ func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation) e
 
 // UpdateInstallation updates the given installation in the database.
 func (sqlStore *SQLStore) UpdateInstallation(installation *model.Installation) error {
-	_, err := sqlStore.execBuilder(sqlStore.db, sq.
+	envJSON, err := json.Marshal(installation.MattermostEnv)
+	if err != nil {
+		errors.Wrap(err, "unable to marshal mattermost env")
```
gigawhitlocks: nit: "mattermost env" -> "MattermostEnv" just so that it's clear when this error is printed that the issue is with the field MattermostEnv and not some generic "mattermost environment"

internal/store/migrations.go
```diff
@@ -600,4 +600,16 @@ var migrations = []migration{
 		 `)
 		return err
 	}},
+	{semver.MustParse("0.12.1"), semver.MustParse("0.13.0"), func(e execer) error {
+		// Add Mattermost Env column for installations.
+		_, err := e.Exec(`
+				ALTER TABLE Installation
+				ADD COLUMN MattermostEnvRaw BYTEA NULL;
```
gigawhitlocks: Do the existing `store` tests cover the case where this is `nil`? I think they should, but just to be sure.. don't want to hit that same issue I did..

model/env.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// EnvVar contains the value source for a given environement variable.
+type EnvVar struct {
+	Value     string               `json:"value,omitempty"`
+	ValueFrom *corev1.EnvVarSource `json:"valueFrom,omitempty"`
+}
+
+// EnvVarMap is a map of multiple env var names to their values.
+type EnvVarMap map[string]EnvVar
```
gigawhitlocks: 0/5 should EnvVarMap be a `map[string]*EnvVar` instead?

model/installation_states.go
```diff
@@ -121,9 +121,8 @@ func validTransitionToInstallationStateCreationRequested(currentState string) bo
 func validTransitionToInstallationStateUpgradeRequested(currentState string) bool {
 	switch currentState {
 	case InstallationStateStable,
-		InstallationStateUpgradeRequested,
-		InstallationStateUpgradeInProgress,
-		InstallationStateUpgradeFailed:
+		InstallationStateUpdateRequested,
+		InstallationStateUpdateFailed:
```
gigawhitlocks: It's invalid to go from `InstallationStateStable` to `InstallationStateUpdateInProgress` now? Maybe this was always invalid and the intent is to go through `InstallationStateUpdateRequested` first? e.g. Stable -> UpdateRequested -> UpdateInProgress?

gigawhitlocks:
Also next time, if you are doing a big rename like UpgradeInstallation->UpdateInstallation it would help if that refactor was in its own commit. Not a big deal, just something that would be nice.

cmd/cloud/installation.go
```diff
@@ -214,3 +232,26 @@ var installationShowStateReport = &cobra.Command{
...
+	if len(rawInput) == 0 {
+		return nil, nil
+	}
+
+	envVarMap := make(model.EnvVarMap)
+
+	for _, env := range rawInput {
+		kv := strings.Split(env, "=")
+		if len(kv) != 2 {
+			return nil, errors.Errorf("%s is not in a valid env format; expecting KEY_NAME=VALUE", env)
```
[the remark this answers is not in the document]
gabrieljackson: Are you concerned that something like `--mattermost-env=KEY1=VALUE1` won't work? That works with this code along with `--mattermost-env KEY1=VALUE1`.

The library handles anything before the actual flag value.

internal/api/installation.go
```diff
@@ -210,39 +211,36 @@ func handleUpgradeInstallation(c *Context, w http.ResponseWriter, r *http.Reques
...
 
-		err := c.Store.UpdateInstallation(installation)
-		if err != nil {
-			c.Logger.WithError(err).Error("failed to mark installation for upgrade")
-			w.WriteHeader(http.StatusInternalServerError)
-			return
-		}
+	err = c.Store.UpdateInstallation(installation)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to mark installation for update")
```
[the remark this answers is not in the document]
gabrieljackson: Yeah, that's exactly right. Originally the endpoint was just version and then license came afterwards. The intent of the word upgrade was just with versions. I don't know about you, but I find it confusing when upgrade would mean that everything is staying the same except one env var being removed as an example.

internal/store/migrations.go
gabrieljackson: Oh, I meant to tell you that I did a bunch of testing on this and nil/null is just fine (and a little easier to read that the empty struct)

gabrieljackson:
You are totally right about me pushing the update name change into this PR. I sorta started to use the word update on newer functions/names and found it confusing so figured it would be better to make them all align. That change is a bit bigger than I thought so I should have done it separately.

I will be sure to do that when I get in a similar situation again.

model/installation_states.go
gabrieljackson: So this change is saying that it is no longer possible to go from `InstallationStateUpdateInProgress` to `InstallationStateUpgradeRequested`. The reason for this is I am a bit concerned about the edge cases on what could be happening when the installation is being updated and then is asked to be updated again.

Although, I don't have a conclusive thought on why this should be yet, I want to be safe until we can test it.

model/env.go
gabrieljackson: Either seems fine, but the only reason I am 0.5/5 on this is because this method made it cleaner to use elsewhere in the code. I will change it if you think it's better that way.

cmd/cloud/installation.go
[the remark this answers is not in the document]
gigawhitlocks: No I'm concerned with the case where VALUE1 contains an = e.g. `--mattermost-env=sshkey=AAAAB3NzaC1yc2EAAAADAQABAAABgQDBVn3ekvzpFu61EnPQ1J5pSMkd/I4WZv9vVI28qxArK5oALRNU+Q6ko0sx1LFToAs6lCXTLUj+XI+59f5ZI2vCp8n6v0A69hCj41Ztm8dIBz2S4hj8fhlfKW4+KlGAaxZDD5gAq+Jua+ejexOK3uSBcljt0kvbVvcKf9ngakPvxI2Eeji10nazOUkA+wvOvNLcc2b8evaXp9GTm+Y5nSD4QEBZ0w6t2wTm4JAa+PJIQ4Cqv22vhrZLiUL78GsVYCTdt+zC4YOVxYsdfqktQ4ZSQQ34RHhBg+V5Rul0vrHzTL87xFJ/Cgfg7L5bJgihhixy1LFz5j2AEYJ6xfhij++QSRpAriUQOsjFjEHvVxW4VbHe9l508AO37XNM5tdq5dG70UkBuXqpdcU6fmK0wJf/vPyOC5V49RVK2Y5RsOpfK4T9WiUHErQqS2T/FUF7OuWssJfz4DohnBDr2Uoi2GdMqDaHkjcAPgoV1AJHLPPzfI7QknoBcvcd0aoo+Up3sBE=`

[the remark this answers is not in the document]
gigawhitlocks: Maybe not the _best_ example because I think SSH keys will still work without the trailing `=` but hopefully it makes the point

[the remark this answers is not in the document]
gabrieljackson: Oh, yeah you are right about that. I can't see any reason not to support that kind of thing so I will make the change.

Thanks.

gabrieljackson:
Thanks for the review so far @gigawhitlocks. I believe I have responded or addressed to all your comments.

internal/api/installation.go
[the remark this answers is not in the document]
gigawhitlocks: Yes, thank you for explaining

model/env.go
gigawhitlocks: No, whatever makes it cleaner. I guess in theory if this map got big there might be performance implications but that's probably a premature optimization on my part.

model/installation_states.go
gigawhitlocks: OK sounds good, mostly I wanted to make sure this line didn't get removed by accident. I don't think there's a reason to keep around states/state transitions that are not known to be necessary & are not tested, so this is a good change imho, thanks for explaining.

cpanato:
@gabrieljackson we need to do the same for the probes, or at least if we have any custom values in the clusterinstallation manifest the provisioner not override those

gabrieljackson:
@cpanato, were you able to spend some time to review this? I would like to get it in to get this merged for @gigawhitlocks PR.

Thanks

cpanato:
@gabrieljackson sure, sorry for the delay, will review that tomorrow my morning

cpanato:
also there are others to review as well 😅 @gabrieljackson

cpanato:
please lets talk about this: https://github.com/mattermost/mattermost-cloud/pull/161#issuecomment-587585919 @gabrieljackson

PR #164, opened 2020-03-10
Initial installation group config inheritance

gabrieljackson:
 - Adds foundational logic for merging installation and group
   configuration. This is only exposed via the get installation
   API endpoint for now. No supervisor logic has been altered.
 - Prevent saving installation objects once they have merged
   group config. This is to prevent errors when the provisioner
   eventually acts on group config changes.
 - Logging and group store cleanup.

https://mattermost.atlassian.net/browse/MM-22504

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -34,6 +34,8 @@ func init() {
 	installationDeleteCmd.MarkFlagRequired("installation")
 
 	installationGetCmd.Flags().String("installation", "", "The id of the installation to be fetched.")
+	installationGetCmd.Flags().Bool("include-group-config", false, "Whether to include group configuration in the installation or not.")
+	installationGetCmd.Flags().Bool("include-group-config-overrides", false, "Whether to include a group configuration override summary in the installation or not.")
 	installationGetCmd.MarkFlagRequired("installation")
 
 	installationListCmd.Flags().String("owner", "", "The owner by which to filter installations.")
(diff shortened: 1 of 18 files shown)


model/installation.go
```diff
@@ -23,6 +24,13 @@ type Installation struct {
 	DeleteAt       int64
 	LockAcquiredBy *string
 	LockAcquiredAt int64
+	GroupOverrides map[string]string `json:"GroupOverrides,omitempty"`
+
+	// configconfigMergedWithGroup is set when the installation configuration
+	// has been overridden with group configuration. This value can then be
+	// checked later to determine whether the installation is safe to save or
+	// not.
+	configMergedWithGroup bool
```
jwilander: Private fields don't get serialized to JSON do they? If so might be worth adding `json:"-"`

gabrieljackson: Yeah, they don't so we good there.

cmd/cloud/installation.go
```diff
@@ -34,6 +34,8 @@ func init() {
 	installationDeleteCmd.MarkFlagRequired("installation")
 
 	installationGetCmd.Flags().String("installation", "", "The id of the installation to be fetched.")
+	installationGetCmd.Flags().Bool("include-group-config", false, "Whether to include group configuration in the installation or not.")
```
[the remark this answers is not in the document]
gabrieljackson: I agree that we would want the defaults to be whatever the provisioner is actually using to make updates. Because groups are currently still a no-op I left that info out. I will change the defaults when group values start impacting the installations.

PR #166, opened 2020-03-13
logs: add machine readable format flag for logging

cpanato:
#### Summary
Add option to enable machine-readable logging format which in this case is JSON 😄
to set use `machine-readable-logs` flag
This will enable the Mattermost Cloud Monitoring to parse and read more easily the logs.

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-23236

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -44,6 +44,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("keep-database-data", true, "Whether to preserve database data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
+	serverCmd.PersistentFlags().Bool("machine-readable-logs", false, "Output the logs in machine readable format.")
 }
 
 var serverCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/cloud/server.go
```diff
@@ -44,6 +44,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("keep-database-data", true, "Whether to preserve database data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
+	serverCmd.PersistentFlags().Bool("machine-readable-logs", true, "Output the logs in machine readable format. Default to true.")
```
gabrieljackson: I am pretty sure we agreed yesterday that not changing the default log behavior would be best. Shouldn't this be false for the default?

cpanato: i think so 😇 
but in terms of deployment and operation, this is the best option, fewer changes in the environments.
but if everybody wants to have as false I can make it.

cpanato: set to false

cmd/cloud/server.go
```diff
@@ -44,6 +44,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("keep-database-data", true, "Whether to preserve database data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
+	serverCmd.PersistentFlags().Bool("machine-readable-logs", false, "Output the logs in machine readable format. Default to true.")
```
gabrieljackson: This would need to be updated, but the default flag value is printed automatically so we should just remove this part.

cpanato: done

PR #167, opened 2020-03-18
MM-23257 Adding support for public-nginx and cert-manager

stylianosrigas:
With this PR the provisioner will be able to deploy:

- Public NGINX 
- Cert Manager
- Cluster Issuer

 Helm charts are used for NGINX and Cert-Manager and manifest for the Cluster Issuer.


model/cluster_utility.go
```diff
@@ -23,6 +27,10 @@ const (
 	NginxDefaultVersion = "1.30.0"
 	// FluentbitDefaultVersion defines the default version for the Helm chart
 	FluentbitDefaultVersion = "2.8.7"
+	// CertManagerDefaultVersion defines the default version for the Helm chart
+	CertManagerDefaultVersion = "v0.13.1"
```
gigawhitlocks: Is the v in this version string intentional or a typo?

stylianosrigas: @gigawhitlocks its intentional. Thats how they define the version for certmanager.

gigawhitlocks: Perfect, thought that might be the case.

internal/provisioner/certmanager.go
```diff
@@ -0,0 +1,170 @@
...
+		defer cancel()
+		pod, err := k8sClient.WaitForPodRunning(ctx, "cert-manager", pod.GetName())
+		if err != nil {
+			return err
+		}
+		logger.Infof("Successfully deployed cert-manager-webhook pod %q", pod.Name)
+	}
+
+	//After the webhook is up and running it needs ~2sec before being able to accept connections and being able to deploy the cluster issuer.
+	time.Sleep(5 * time.Second)
```
gigawhitlocks: Can we poll the endpoint until it comes up, or something like that, instead of just a sleep? 0/5 on this, just a thought.

internal/provisioner/utility_group.go
```diff
@@ -90,9 +110,16 @@ func newUtilityGroupHandle(kops *kops.Cmd, provisioner *KopsProvisioner, cluster
 
 // CreateUtilityGroup  creates  and  starts  all of  the  third  party
 // services needed to run a cluster.
-func (group utilityGroup) CreateUtilityGroup() error {
+func (group utilityGroup) CreateUtilityGroup(kops *kops.Cmd) error {
```
gigawhitlocks: The utilityGroup has a member `kops` of type `*kops.Cmd`. Do we need to have `kops` passed in as an argument here (and in Update as well)? `kops` is instantiated in `newUtilityGroupHandle`, unless for some reason here we need a different instance of that type?

gigawhitlocks:
Looks good! Just had a couple of questions :)

internal/provisioner/certmanager.go
stylianosrigas: I don't like it either but the thing is that as soon as the pod is app it takes it like a second to prepare the hook. Its not sth changing in every run, its an application thing.

internal/provisioner/utility_group.go
stylianosrigas: Thanks for the catch, totally missed that. Fixing now :)

internal/provisioner/certmanager.go
cpanato: i will check this check is being made in the cluster-api, I think I saw something like that

cpanato: you can do something like
```
$ kubectl wait --for condition=Available --timeout 300s apiservice v1alpha2.cert-manager.io
```
this will return when the condition met, and error when timeout


just need to check what is the service you want to check

stylianosrigas: thats great thanks, I will have a look.

stylianosrigas: Pushed a fix for that

stylianosrigas:
@cpanato @gigawhitlocks please have another look. I pushed a fix for the sleep time :)

stylianosrigas:
@gigawhitlocks @gabrieljackson @cpanato Can you please have one last look? I pushed a fix using retry for the cluster issuer creation.

gigawhitlocks:
Double approve from me :laughing:

cpanato:
@stylianosrigas please update the PR and resolve the conflict

PR #171, opened 2020-03-23
Add group API locks

gabrieljackson:
 - Group update and delete API endpoints now lock the group object
   before making changes.
 - Removed group version API endpoint. The update endpoint handles
   version and other group configuration changes.
 - Refactored group request objects and streamlined validation.

https://mattermost.atlassian.net/browse/MM-23442

### DIFF
--- a/internal/api/group.go
+++ b/internal/api/group.go
@@ -21,54 +21,28 @@ func initGroup(apiRouter *mux.Router, context *Context) {
 	groupRouter.Handle("", addContext(handleGetGroup)).Methods("GET")
 	groupRouter.Handle("", addContext(handleUpdateGroup)).Methods("PUT")
 	groupRouter.Handle("", addContext(handleDeleteGroup)).Methods("DELETE")
-
-	groupRouter.Handle("/mattermost/{version:[A-Za-z0-9.]+}", addContext(handleChangeGroupMattermostVersion)).Methods("PUT")
 }
 
-// handleChangeGroupMattermostVersion responds to PUT
-// /api/group/{group}/mattermost/{version}, upgrading or downgrading
-// the installation to the Mattermost version embedded in the request.
-func handleChangeGroupMattermostVersion(c *Context, w http.ResponseWriter, r *http.Request) {
+// handleGetGroup responds to GET /api/group/{group}, returning the group in question.
+func handleGetGroup(c *Context, w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-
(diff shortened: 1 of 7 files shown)


internal/api/group.go
```diff
@@ -21,54 +21,28 @@ func initGroup(apiRouter *mux.Router, context *Context) {
...
 	groupRouter.Handle("", addContext(handleUpdateGroup)).Methods("PUT")
 	groupRouter.Handle("", addContext(handleDeleteGroup)).Methods("DELETE")
-
-	groupRouter.Handle("/mattermost/{version:[A-Za-z0-9.]+}", addContext(handleChangeGroupMattermostVersion)).Methods("PUT")
 }
 
-// handleChangeGroupMattermostVersion responds to PUT
-// /api/group/{group}/mattermost/{version}, upgrading or downgrading
-// the installation to the Mattermost version embedded in the request.
-func handleChangeGroupMattermostVersion(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks: I guess the right way to change this setting is now through using `handleUpdateGroup`?

jwilander:
I'll let @gigawhitlocks handle the dev review

model/group_request.go
```diff
@@ -25,18 +25,22 @@ func NewCreateGroupRequestFromReader(reader io.Reader) (*CreateGroupRequest, err
 		return nil, errors.Wrap(err, "failed to decode create group request")
 	}
 
-	if createGroupRequest.Name == "" {
-		return nil, errors.New("must specify name")
-	}
-	if createGroupRequest.Version == "" {
-		return nil, errors.New("must specify version")
+	err = createGroupRequest.Validate()
```
gigawhitlocks: :+1: I feel like you told me to move this line here during my PR so my apologies for missing it, assuming that's what happened. Regardless, I'm happy to see this moved.

internal/api/group.go
gabrieljackson: Yep, that's exactly right! The version endpoint was from the original group code where we didn't see a need for anything other than version so we streamlined it.

The cloud client hasn't actually used that endpoint in a while.

model/group_request.go
gabrieljackson: No worries! It's there now.

gabrieljackson:
Thanks for the review!

internal/api/group.go
gigawhitlocks: Perfect :+1:

PR #173, opened 2020-03-24
MM-21931: authorize and revoke ingress access of an RDS instance

gsagula:
#### Summary
Introduces  RDS database migration interface and the implementation of a couple operations for setting and tearing down the migration process. This PR addresses the need for authorize and revoke ingress access from one RDS instance to another RDS instance. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-21931


internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+				UserIdGroupPairs: []*ec2.UserIdGroupPair{
+					{
+						Description: aws.String("Ingress Traffic from other RDS instance"),
+						GroupId:     slaveInstanceSG.GroupId,
+					},
+				},
+			},
+		},
+	})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
gabrieljackson: Can we not type assert the error here? When checking AWS errors elsewhere I follow their guidelines and type check them. This seems like it could be brittle.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+		return "", err
+	}
+
+	logger.WithField("migration-installation-id", d.masterInstallationID).Info("database migration setup completed")
```
gabrieljackson: Nit: Capital letter start for log messages
```suggestion
	logger.WithField("migration-installation-id", d.masterInstallationID).Info("Database migration setup completed")
```

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+				ToPort:     aws.Int64(3306),
+				UserIdGroupPairs: []*ec2.UserIdGroupPair{
+					{
+						GroupId: slaveInstanceSG.GroupId,
+					},
+				},
+			},
+		},
+	})
+	if err != nil && !strings.Contains(err.Error(), "rule does not exist") {
```
gabrieljackson: Same question as above.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+		return "", err
+	}
+
+	logger.WithField("migration-installation-id", d.masterInstallationID).Info("database migration teardown completed")
+
+	return model.DatabaseMigrationStatusTeardownComplete, nil
+}
+
+// Replicate starts the process for replicating an master RDS database. This method must return an resplication status or an error.
```
gabrieljackson:
```suggestion
// Replicate starts the process for replicating a master RDS database. This method must return a replication status or an error.
```

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+func (d *RDSDatabaseMigration) Replicate(logger log.FieldLogger) (string, error) {
```
gsagula: I think this should return `not implemented` error. I will change it.

internal/tools/aws/database_migration.go
gsagula: Good call! I think I saw how to do that somewhere.

internal/tools/aws/database_migration.go
gsagula: Good call! I think I saw how to do that somewhere.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, instance := range output.DBInstances {
+		for _, vpcSG := range instance.VpcSecurityGroups {
+			sgOutput, err := d.awsClient.Service().ec2.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
+				GroupIds: []*string{vpcSG.VpcSecurityGroupId},
+			})
+			if err != nil {
+				return nil, err
+			}
+			if len(sgOutput.SecurityGroups) == 1 && strings.Contains(*sgOutput.SecurityGroups[0].GroupName, "-db-sg") {
```
gigawhitlocks: I can't seem to find where the security group name's naming scheme is forced to include `-db-sg` anywhere, I suppose at security group creation time? Can you help me find that? If it's not enforced and is just by convention, I think the convention should be encoded so that it is enforced, and this hardcoded string `"-db-sg"` should be updated to use a constant (presumably the constant would also be useful at convention-enforcement time).

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+func (d *RDSDatabaseMigration) Replicate(logger log.FieldLogger) (string, error) {
+	return "", errors.New("not implemented")
+}
+
+func (d *RDSDatabaseMigration) describeDBInstanceSecurityGroup(instanceID string) (*ec2.SecurityGroup, error) {
+	output, err := d.awsClient.Service().rds.DescribeDBInstances(&rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: aws.String(instanceID),
+	})
+	if err != nil {
+		return nil, err
```
gigawhitlocks: Just a reminder to be sure to `Wrap` these bare `err`s with some more context when you `return` them, before merge. I know there can be a lot of churn with these statements during review so I assume you intend to do this before merge and just haven't yet.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return "", err
+	}
+
+	logger.WithField("migration-installation-id", d.masterInstallationID).Info("database migration setup completed")
+
+	return model.DatabaseMigrationStatusSetupComplete, nil
+}
+
+// Teardown removes access from one RDS database to another and rollback any previous database configuration.
```
gigawhitlocks: The actual rollback is coming in another PR right? Or am I overlooking it?

gigawhitlocks:
Looks good! Just had small questions.

internal/tools/aws/database_migration.go
gigawhitlocks: It's also fine if this :point_up: is going to be in the next PR :)

gsagula: This is probably in Terraform:
```
mattermost-cloud-dev-provisioning-1024000-db-sg
```

internal/tools/aws/database_migration.go
gsagula: Next PR. It would be way too much for this one.

gigawhitlocks: Sounds good :+1:

internal/tools/aws/database_migration.go
gigawhitlocks: Hmm, if it's from Terraform I guess I don't feel too strongly about it one way or the other. A comment on that line saying where the value comes from might be good, but it's fine even without it.

gsagula: That's a very good point. I can try to find some other metadata but it will be from Terraform.

gsagula:
@gigawhitlocks @gabrieljackson Thank you for the review. I did the changes that you suggested. Feel free to take another look.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,163 @@
...
+// Setup sets access from one RDS database to another and sets any configuration needed for replication.
+func (d *RDSDatabaseMigration) Setup(logger log.FieldLogger) (string, error) {
+	masterInstanceSG, err := d.describeDBInstanceSecurityGroup(RDSMasterInstanceID(d.masterInstallationID))
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to setup database migration for installation id: %s", d.masterInstallationID)
+	}
+
+	slaveInstanceSG, err := d.describeDBInstanceSecurityGroup(RDSMigrationInstanceID(d.slaveInstallationID))
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to setup database migration for installation id: %s", d.masterInstallationID)
```
gabrieljackson: Should the error include that the slaveInstance was unable to be found? 0/5 on this.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return nil, errors.Errorf("security group for RDS DB instance %s not found", instanceID)
+}
+
+func isSecurityGroupRuleNotFound(err error) bool {
+	if aerr, ok := err.(awserr.Error); ok {
+		switch aerr.Code() {
+		// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#CommonErrors
+		case "InvalidPermission.NotFound":
```
gabrieljackson: Is there not an AWS error constant for this? I believe I used constants when doing this check elsewhere.

internal/tools/aws/database_migration.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+	return false
+}
+
+func isSecurityGroupRuleDuplicate(err error) bool {
+	if aerr, ok := err.(awserr.Error); ok {
+		switch aerr.Code() {
+		// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#CommonErrors
+		case "InvalidPermission.Duplicate":
```
gabrieljackson: Same here.

internal/tools/aws/database_migration.go
gsagula: I couldn't find it in the SDK code base. I may ask about it in GH repo.

internal/tools/aws/database_migration.go
gsagula: Please, see my answer above.

internal/tools/aws/database_migration.go
gsagula: https://github.com/aws/aws-sdk-go/issues/3235

gsagula: I will add a comment in code.

internal/tools/aws/database_migration.go
gsagula: ` slaveInstance ID was unable to be found` will probably be available in the wrapped error, but what we were trying migrate was the master installation. I think the error message could contain more information though.

PR #176, opened 2020-03-25
Add group supervisor for installation updates

gabrieljackson:
The group supervisor is responsible for managing rolling updates
to installations when a group's configuration is updated. It does
this by moving installations in the group to an update-requested
state while respecting the group's MaxRolling setting.

Note that the cloud server still does not apply merged group config
to installations so the group supervisor is disabled by default.
This will be changed in a future commit.

This change also corrects a SQL query bug as well as a bug with
how group merged config overrides are reported.

https://mattermost.atlassian.net/browse/MM-23154

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -36,6 +36,7 @@ func init() {
 	serverCmd.PersistentFlags().String("database", "sqlite://cloud.db", "The database backing the provisioning server.")
 	serverCmd.PersistentFlags().String("listen", ":8075", "The interface and port on which to listen.")
 	serverCmd.PersistentFlags().Bool("cluster-supervisor", true, "Whether this server will run a cluster supervisor or not.")
+	serverCmd.PersistentFlags().Bool("group-supervisor", false, "Whether this server will run an installation group supervisor or not.")
 	serverCmd.PersistentFlags().Bool("installation-supervisor", true, "Whether this server will run an installation supervisor or not.")
 	serverCmd.PersistentFlags().Bool("cluster-installation-supervisor", true, "Whether this server will run a cluster installation supervisor or not.")
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
@@ -90,9 +91,10 @@ var serverCmd = &cobra.Command{
 		}
 
 		clusterSupervisor, _ := command.Flags().GetBool("cluster-supervisor")
-		clusterInstallationSupervisor, _ := command.Flags().GetBool("cluster-installation-supervisor")
+		groupSupervisor, _ := command.Flags().GetBool("group-supervisor")
 		installationSupervisor, _ := command.Flags().GetBool("installation-supervisor")
-		if !clusterSupervisor && !installationSupervisor && !clusterInstallationSupervisor {
+		clusterInstallationSupervisor, _ := command.Flags().GetBool("cluster-installation-supervisor")
+		if !clusterSupervisor && !installationSupervisor && !clusterInstallationSupervisor && !groupSupervisor {
 			logger.Warn("Server will be running with no supervisors. Only API functionality will work.")
 		}
 
@@ -109,6 +111,7 @@ var serverCmd = &cobra.Command{
 
 		logger.WithFields(logrus.Fields{
 			"cluster-supervisor":              clusterSupervisor,
+			"group-supervisor":                groupSupervisor,
 			"installation-supervisor":         installationSupervisor,
 			"cluster-installation-supervisor": clusterInstallationSupervisor,
 			"store-version":                   currentVersion,
@@ -152,6 +155,9 @@ var serverCmd = &cobra.Command{
 		var multiDoer supervisor.MultiDoer
 		if clusterSupervisor {
 			multiDoer = append(multiDoer, supervisor.NewClusterSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, logger))
+		}
+		if groupSupervisor {
+			multiDoer = append(multiDoer, supervisor.NewGroupSupervisor(sqlStore, instanceID, logger))
 		}
 		if installationSupervisor {
 			multiDoer = append(multiDoer, supervisor.NewInstallationSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, clusterResourceThreshold, keepDatabaseData, keepFilestoreData, resourceUtil, logger))
--- a/internal/store/group.go
+++ b/internal/store/group.go
@@ -78,7 +78,7 @@ func (sqlStore *SQLStore) GetUnlockedGroupsPendingWork() ([]*model.Group, error)
 			Select("ID").
 			From("Installation").
 			Where("GroupID = ?", rawGroup.ID).
-			Where("GroupSequence IS NULL OR GroupSequence != ?", rawGroup.Sequence).
+			Where("(GroupSequence IS NULL OR GroupSequence != ?)", rawGroup.Sequence).
 			Where("DeleteAt = 0").
 			Limit(1)
 		err = sqlStore.selectBuilder(sqlStore.db, &installations, installationBuilder)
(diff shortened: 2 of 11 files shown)


cmd/cloud/server.go
```diff
@@ -36,6 +36,7 @@ func init() {
 	serverCmd.PersistentFlags().String("database", "sqlite://cloud.db", "The database backing the provisioning server.")
 	serverCmd.PersistentFlags().String("listen", ":8075", "The interface and port on which to listen.")
 	serverCmd.PersistentFlags().Bool("cluster-supervisor", true, "Whether this server will run a cluster supervisor or not.")
+	serverCmd.PersistentFlags().Bool("group-supervisor", false, "Whether this server will run an installation group supervisor or not.")
```
gigawhitlocks: I must've missed the end of the discussion during the meeting; I thought we settled on `true` as the default for this. What was the reasoning behind a `false` default, if you don't mind my asking?

cmd/cloud/server.go
```diff
@@ -90,9 +91,10 @@ var serverCmd = &cobra.Command{
 		}
 
 		clusterSupervisor, _ := command.Flags().GetBool("cluster-supervisor")
-		clusterInstallationSupervisor, _ := command.Flags().GetBool("cluster-installation-supervisor")
+		groupSupervisor, _ := command.Flags().GetBool("group-supervisor")
 		installationSupervisor, _ := command.Flags().GetBool("installation-supervisor")
-		if !clusterSupervisor && !installationSupervisor && !clusterInstallationSupervisor {
+		clusterInstallationSupervisor, _ := command.Flags().GetBool("cluster-installation-supervisor")
+		if !clusterSupervisor && !installationSupervisor && !clusterInstallationSupervisor && !groupSupervisor {
```
gigawhitlocks: Could we break this check into 3 and log more-specifically what won't work if each supervisor is disabled? 0/5 on this, just a thought.

cmd/cloud/server.go
gabrieljackson: Haha, I remember completely the opposite. @jwilander, can you weigh in with what you remember?

internal/store/group.go
```diff
@@ -88,12 +88,96 @@ func (sqlStore *SQLStore) GetUnlockedGroupsPendingWork() ([]*model.Group, error)
...
+		return nil, err
+	}
+	count, err = stableResult.value()
+	if err != nil {
+		return nil, err
+	}
+	metadata.InstallationStableCount = count
+	metadata.InstallationNonStableCount = metadata.InstallationTotalCount - metadata.InstallationStableCount
+
+	if metadata.InstallationNonStableCount < 0 {
```
gigawhitlocks: How is this condition possible? :confused:

cmd/cloud/server.go
jwilander: I'm pretty sure @cpanato just wanted control for what was turned on and not to have stuff just start on by default. E.g. we have 4 provisioner pods right now with 1 tagged for cluster only and 3 for installation only. If this was on by default and pushed those would all start doing group stuff automatically

jwilander: Aka I recall the default being false

internal/store/group.go
gabrieljackson: It's a catch against the subtraction above. The only way it would happen is if the total query or the stable query returned something incorrect due to either code error or DB error. It's basically a "this should never happen comment".

cmd/cloud/server.go
gigawhitlocks: Sounds good, I remembered the discussion but couldn't remember @cpanato 's opinion so thanks for the reminder of why we chose this default.

gabrieljackson: That's what I remember as well. Basically default `true` would make stuff just start happening if they didn't manage to update the start commands on the pods in time.

cmd/cloud/server.go
gabrieljackson: Do you mean 4 or more checks? One for each supervisor?

The logic check here is less about warning that a supervisor is disabled and more about warning that all supervisors are disabled. If that were true the server would just be doing API stuff like I mentioned in the log.

Do you think this should be changed so individual warnings are given? I don't want to clog up the server startup logs with warnings that sorta aren't warnings but only 0/5.

internal/supervisor/group.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Supervise schedules the required work on the given group.
+func (s *GroupSupervisor) Supervise(group *model.Group) {
+	logger := s.logger.WithFields(log.Fields{
+		"group": group.ID,
+	})
+
+	groupLock := newGroupLock(group.ID, s.instanceID, s.store, logger)
+	if !groupLock.TryLock() {
```
gigawhitlocks: Do we want to log an error here? Or a warning?

internal/supervisor/group.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, id := range groupMetadata.InstallationIDsToBeRolled {
+		if groupMetadata.InstallationNonStableCount+moved >= group.MaxRolling {
+			// We have bumped up against the max rolling count with the new
+			// installations added to the rolling pool.
+			break
+		}
+
+		installationLock := newInstallationLock(id, s.instanceID, s.store, logger)
+		if !installationLock.TryLock() {
+			return
```
gigawhitlocks: Same question here, do we want a log message when this happens? Though in this case, I guess it has a potential to be noisy here. 0/5 just a thought.

internal/supervisor/group_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (s *mockGroupStore) UnlockInstallation(installationID, lockerID string, force bool) (bool, error) {
+	if s.UnlockChan != nil {
+		close(s.UnlockChan)
+	}
+	return true, nil
+}
+
+func TestGroupSupervisorDo(t *testing.T) {
```
gigawhitlocks: Do we have a test case that I'm overlooking that exercises the maxRolling feature/setting?

internal/store/group.go
gigawhitlocks: OK good that's what I was hoping :joy: 
Basically my question was a sanity check for myself. :)

cmd/cloud/server.go
gigawhitlocks: >Do you think this should be changed so individual warnings are given?
I'm actually not sure, I guess my concern was that the user might not be aware of limitations imposed if one of the supervisors (but not all) was disabled.

In case my original comment was unclear, I wasn't asking you to break it into three checks, I was looking to bring your attention to the log message and ask, really ask, if you thought we _should_ do that. If it doesn't make sense to you to do that, just resolve this conversation and move on :)

gigawhitlocks:
Looks good overall, no major problems that stand out to me, just some easy questions about logging.

cmd/cloud/server.go
gigawhitlocks: Ohhhh gotcha, okay, thanks

internal/supervisor/group.go
gabrieljackson: The lock objects have a logger in them already and log on error.

internal/supervisor/group.go
gabrieljackson: The lock objects have a logger in them already and log on error. Example: https://github.com/mattermost/mattermost-cloud/pull/176/files#diff-2e128f917c08be9b9e3f93249985f459R31

internal/supervisor/group_test.go
gabrieljackson: Nothing directly for that yet. I have the stub for the test here and will update it on the final PR when all the group merge stuff is turned on.

internal/supervisor/group.go
gabrieljackson: For more context, the lockers are usually deferred so we have them handle their own logging needs.

gigawhitlocks: Perfect

gigawhitlocks:
Thanks for answering my questions @gabrieljackson :+1:

PR #178, opened 2020-03-27
Mm 22392 - Adding support for NGINX for new installations

stylianosrigas:
With this PR:

- New installations should stop using dedicated Load Balancer service
- NGINX Load Balancer service should be used by new installations
- Certificate, TLS and caching should be handled by Ingress Annotations

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -68,6 +68,7 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 		Spec: mmv1alpha1.ClusterInstallationSpec{
 			Size:          installation.Size,
 			Version:       translateMattermostVersion(installation.Version),
+			Image:         installation.Image,
 			IngressName:   installation.DNS,
 			MattermostEnv: installation.MattermostEnv.ToEnvList(),
 			UseIngressTLS: true,
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -52,11 +52,6 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 		return errors.Wrapf(err, "failed to create namespace %s", clusterInstallation.Namespace)
 	}
 
-	certificateSummary, err := awsClient.GetCertificateSummaryByTag(aws.DefaultInstallCertificatesTagKey, aws.DefaultInstallCertificatesTagValue, logger)
-	if err != nil {
-		return errors.Wrapf(err, "failed to fetch AWS certificate ARN for tag %s:%s", aws.DefaultInstallCertificatesTagKey, aws.DefaultInstallCertificatesTagValue)
-	}
-
 	mattermostInstallation := &mmv1alpha1.ClusterInstallation{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "ClusterInstallation",
@@ -71,17 +66,31 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 			},
 		},
 		Spec: mmv1alpha1.ClusterInstallationSpec{
-			Size:                   installation.Size,
-			Version:                translateMattermostVersion(installation.Version),
-			Image:                  installation.Image,
-			IngressName:            installation.DNS,
-			UseServiceLoadBalancer: true,
-			MattermostEnv:          installation.MattermostEnv.ToEnvList(),
-			ServiceAnnotations: map[string]string{
-				"service.beta.kubernetes.io/aws-load-balancer-backend-protocol":        "tcp",
-				"service.beta.kubernetes.io/aws-load-balancer-ssl-cert":                *certificateSummary.CertificateArn,
-				"service.beta.kubernetes.io/aws-load-balancer-ssl-ports":               "https",
-				"service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "120",
+			Size:          installation.Size,
+			Version:       translateMattermostVersion(installation.Version),
+			IngressName:   installation.DNS,
+			MattermostEnv: installation.MattermostEnv.ToEnvList(),
+			UseIngressTLS: true,
+			IngressAnnotations: map[string]string{
+				"cert-manager.io/cluster-issuer":                       "letsencrypt-prod",
+				"kubernetes.io/ingress.class":                          "nginx",
+				"kubernetes.io/tls-acme":                               "true",
+				"nginx.ingress.kubernetes.io/proxy-buffering":          "on",
+				"nginx.ingress.kubernetes.io/proxy-body-size":          "100m",
+				"nginx.ingress.kubernetes.io/proxy-send-timeout":       "600",
+				"nginx.ingress.kubernetes.io/proxy-read-timeout":       "600",
+				"nginx.ingress.kubernetes.io/proxy-max-temp-file-size": "0",
+				"nginx.ingress.kubernetes.io/ssl-redirect":             "true",
+				"nginx.ingress.kubernetes.io/configuration-snippet": `
+				  proxy_force_ranges on;
+				  add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
+				  proxy_cache mattermost_cache;  
+				  proxy_cache_revalidate on; 
+				  proxy_cache_min_uses 2;
+				  proxy_cache_use_stale timeout; 
+				  proxy_cache_lock on;
+				  proxy_cache_key "$host$request_uri$cookie_user";`,
+				"nginx.org/server-snippets": "gzip on;",
 			},
 		},
 	}
--- a/internal/provisioner/kops_utils.go
+++ b/internal/provisioner/kops_utils.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost-cloud/internal/tools/k8s"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
 	"github.com/mattermost/mattermost-cloud/internal/tools/terraform"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/pkg/errors"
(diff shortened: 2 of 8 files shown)


internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,49 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+
+	kopsMetadata, err := model.NewKopsMetadata(cluster.ProvisionerMetadata)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse provisioner metadata")
+	}
+
+	if kopsMetadata.Name == "" {
+		logger.Infof("Cluster %s has no name, assuming cluster installation never existed.", cluster.ID)
+		return "", nil
+	}
```
gabrieljackson: Hmm, does this make sense? Not returning an error means the caller will think everything is fine and an endpoint was returned. I am pretty sure we don't need this check because we need the name to exist for the following code to use the name to get the endpoint. They can return an error if the name is not valid for some reason.

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,49 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+			if service.Status.LoadBalancer.Ingress != nil {
+				endpoint := service.Status.LoadBalancer.Ingress[0].Hostname
+				if endpoint == "" {
+					return "", errors.New("loadbalancer endpoint value is empty")
+				}
+
+				return endpoint, nil
+			}
+		}
+	}
```
gabrieljackson: I don't think we want to have this wrapped in a `for` loop. It looks like it could get stuck forever in certain cases.

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,49 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
 		}
 	}
 }
+
+// GetNGINXLoadBalancerEndpoint returns the load balancer endpoint of the NGINX service.
+func (provisioner *KopsProvisioner) GetNGINXLoadBalancerEndpoint(cluster *model.Cluster, namespace string, logger log.FieldLogger) (string, error) {
```
gabrieljackson: The provisioner is one of the few things that actually has its own logger. I would use that instead of passing one in. This is an example of how you can set up the logger right in the beginning of this method:

```
	logger := provisioner.logger.WithFields(log.Fields{
		"cluster":      cluster.ID,
		"nginx-namespace": namespace,
	})
```

gabrieljackson:
If you could clean up your commit history that would be great, but not a big deal.

internal/provisioner/kops_utils.go
stylianosrigas: Yeap you are right. Check not needed, removed.

internal/provisioner/kops_utils.go
stylianosrigas: Thanks for catching this, for shouldn't be there, removed ;)

internal/provisioner/kops_utils.go
stylianosrigas: Changed to use provisioner logger. Thanks :)

stylianosrigas:
> If you could clean up your commit history that would be great, but not a big deal.

@gabrieljackson squashed everything into one commit ;)

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -71,16 +66,22 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
...
-				"service.beta.kubernetes.io/aws-load-balancer-backend-protocol":        "tcp",
-				"service.beta.kubernetes.io/aws-load-balancer-ssl-cert":                *certificateSummary.CertificateArn,
-				"service.beta.kubernetes.io/aws-load-balancer-ssl-ports":               "https",
-				"service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "120",
+			Size:          installation.Size,
+			Version:       translateMattermostVersion(installation.Version),
+			IngressName:   installation.DNS,
+			MattermostEnv: installation.MattermostEnv.ToEnvList(),
+			UseIngressTLS: true,
+			IngressAnnotations: map[string]string{
```
gsagula: Since data is passing through a ALB, can you enable compression here? You could perhaps just use the defaults for gzip-level, gzip-min-length and gzip-types.

https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#gzip-level

stylianosrigas: @gsagula thanks good idea. I enabled gzip which will apply the default values for level, length  and types.

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,52 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+	if err != nil {
+		return "", err
+	}
+
+	services, err := k8sClient.Clientset.CoreV1().Services(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return "", err
+	}
+	for _, service := range services.Items {
+		if service.Status.LoadBalancer.Ingress != nil {
```
gsagula: Any use-case that ingress could contain 0 elements? Just for sake can you take the length of this array instead of inspecting the pointer?
```
if  len(service.Status.LoadBalancer.Ingres) > 0 {  ...  }
```

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,52 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+			endpoint := service.Status.LoadBalancer.Ingress[0].Hostname
```
gsagula: How about removing the extra allocation?
```
if len(service.Status.LoadBalancer.Ingres) > 0 {
	endpoint := &service.Status.LoadBalancer.Ingress[0].Hostname
	if *endpoint == "" {
		return "", errors.New("loadbalancer endpoint value is empty")
	}
	return *endpoint, nil
}
```
or just
```
for _, service := range services.Items {
        if len(service.Status.LoadBalancer.Ingres) > 0 && service.Status.LoadBalancer.Ingress[0].Hostname != "" {
	         return service.Status.LoadBalancer.Ingress[0].Hostname, nil	
        }
        return "", errors.New("loadbalancer endpoint value is empty") 
}
```

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,52 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to export kubecfg")
+	}
+
+	k8sClient, err := k8s.New(kops.GetKubeConfigPath(), logger)
+	if err != nil {
+		return "", err
+	}
+
+	services, err := k8sClient.Clientset.CoreV1().Services(namespace).List(metav1.ListOptions{})
```
gsagula: I'm not sure how big list can be but can't we take advantage `ListOptions` for filtering?

stylianosrigas: @gsagula I don't think its needed because it filters initially by namespace and in the nginx case its gets only one service back.

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -71,16 +66,23 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
...
+				"cert-manager.io/cluster-issuer":                       "letsencrypt-prod",
+				"kubernetes.io/ingress.class":                          "nginx",
+				"kubernetes.io/tls-acme":                               "true",
+				"nginx.ingress.kubernetes.io/proxy-buffering":          "on",
+				"nginx.ingress.kubernetes.io/proxy-body-size":          "100m",
+				"nginx.ingress.kubernetes.io/proxy-send-timeout":       "600",
+				"nginx.ingress.kubernetes.io/proxy-read-timeout":       "600",
+				"nginx.ingress.kubernetes.io/proxy-max-temp-file-size": "0",
+				"nginx.ingress.kubernetes.io/ssl-redirect":             "true",
+				"nginx.ingress.kubernetes.io/configuration-snippet":    "  proxy_force_ranges on;\n  add_header Strict-Transport-Security \"max-age=31536000; includeSubDomains\" always;\n  proxy_cache mattermost_cache;\n  proxy_cache_revalidate on;\n  proxy_cache_min_uses 2;\n  proxy_cache_use_stale timeout;\n  proxy_cache_lock on;\n  proxy_cache_key \"$host$request_uri$cookie_user\";",
```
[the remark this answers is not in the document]
stylianosrigas: I tried with back-tick initially but then it does not understand for some reason the \n to change line and it breaks the config.

internal/provisioner/kops_utils.go
stylianosrigas: yeap like that done ;)

internal/provisioner/kops_utils.go
stylianosrigas: thanks, done :)

internal/provisioner/kops_provisioner_cluster_installation.go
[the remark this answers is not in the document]
gsagula: With back-tick you can do multi-line. Try this https://play.golang.org/p/8R_EW5fWy07

If doesn't work, no worries. It was more a nit thing :)

internal/provisioner/kops_utils.go
gsagula: No worries then, thanks!

internal/provisioner/kops_provisioner_cluster_installation.go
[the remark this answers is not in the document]
stylianosrigas: @gsagula Managed to make it work with back-ticks. Thanks, looks a lot better now.

internal/provisioner/kops_utils.go
gabrieljackson: I think you removed this check from another function above instead of from here. We probably still want to remove this.

internal/provisioner/kops_utils.go
```diff
@@ -110,3 +111,47 @@ func getLoadBalancerEndpoint(ctx context.Context, namespace string, logger log.F
...
+	if err != nil {
+		return "", err
+	}
+	for _, service := range services.Items {
+		if len(service.Status.LoadBalancer.Ingress) > 0 && service.Status.LoadBalancer.Ingress[0].Hostname != "" {
+			return service.Status.LoadBalancer.Ingress[0].Hostname, nil
+		}
+	}
+	return "", nil
```
gabrieljackson, line 156: I think we want to return an error here, right? The endpoint wasn't found so we should return a default error I think.

internal/provisioner/kops_utils.go
stylianosrigas: @gabrieljackson Thanks, reverted back the check from the other function and removed this one.

internal/provisioner/kops_utils.go
stylianosrigas: @gabrieljackson Removed the check and added default error return.

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -71,17 +66,31 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 			},
 		},
 		Spec: mmv1alpha1.ClusterInstallationSpec{
-			Size:                   installation.Size,
-			Version:                translateMattermostVersion(installation.Version),
-			Image:                  installation.Image,
```
gabrieljackson, line 76: Doesn't look like we are setting the image anymore after the change.

stylianosrigas: @gabrieljackson Sorry missed that somehow. Fixed.

PR #179, opened 2020-03-30
Enable group functionality

gabrieljackson:
 - Enables group configuration merging by default when creating
   and updating installations.
 - Provides API functionality for updating group config such as
   Max Rolling.
 - Provides API functionality for returning installation objects
   with or without group configuration merged in. Default setting
   is to return merged config.
 - Provides API functionality to return installations belonging to
   a single group.
 - Adds Image configuration merging logic.
 - Many, many more tests.

Note: the group supervisor which manages lazy installation updates
due to group configuration updates still defaults to not-running
with default server flags. This can be manually enabled for now
and the default will be changed later.

https://mattermost.atlassian.net/browse/MM-22506

### DIFF
--- a/cmd/cloud/group.go
+++ b/cmd/cloud/group.go
@@ -13,6 +13,7 @@ func init() {
 	groupCreateCmd.Flags().String("description", "", "An optional description for this group of installations.")
 	groupCreateCmd.Flags().String("version", "stable", "The Mattermost version for installations in this group to target.")
 	groupCreateCmd.Flags().String("image", "mattermost/mattermost-enterprise-edition", "The Mattermost Container image to use.")
+	groupCreateCmd.Flags().Int64("max-rolling", 1, "The maximum number of installations that can be updated at one time when a group is updated")
 	groupCreateCmd.Flags().StringArray("mattermost-env", []string{}, "Env vars to add to the Mattermost App. Accepts format: KEY_NAME=VALUE. Use the flag multiple times to set multiple env vars.")
 	groupCreateCmd.MarkFlagRequired("name")
 
@@ -21,6 +22,7 @@ func init() {
 	groupUpdateCmd.Flags().String("description", "", "An optional description for this group of installations.")
 	groupUpdateCmd.Flags().String("version", "", "The Mattermost version for installations in this group to target.")
 	groupUpdateCmd.Flags().String("image", "mattermost/mattermost-enterprise-edition", "The Mattermost Container image to use.")
+	groupUpdateCmd.Flags().Int64("max-rolling", 0, "The maximum number of installations that can be updated at one time when a group is updated")
 	groupUpdateCmd.Flags().StringArray("mattermost-env", []string{}, "Env vars to add to the Mattermost App. Accepts format: KEY_NAME=VALUE. Use the flag multiple times to set multiple env vars.")
 	groupUpdateCmd.MarkFlagRequired("group")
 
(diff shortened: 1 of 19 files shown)


model/group_request.go
```diff
@@ -53,9 +48,27 @@ func (request *CreateGroupRequest) Validate() error {
...
+		return nil, errors.Wrap(err, "invalid group create request")
+	}
+
+	return &createGroupRequest, nil
+}
+
 // PatchGroupRequest specifies the parameters for an updated group.
 type PatchGroupRequest struct {
 	ID            string
+	MaxRolling    *int64
```
gigawhitlocks: Why is this a pointer to an int64 instead of just an int64?

gabrieljackson: For that patch logic it's cleaner/easier to use pointers(except for ID of course which is how the patch is correlated).

You can see how the other options are pointers as well.

As we know, the group patching logic is different from other updates (for reasons I can't remember or was never told). I plan to update everything else eventually to standardize.

gigawhitlocks: Do you mean the patch logic that appears starting on line 96 in `Apply`? Pardon me if I'm being thick, but the only difference I can see caused by using pointers in the `PatchGroupRequest` type definition is that now for each member, you have to check to see if it's `nil` before you check the value, which doesn't seem cleaner to me?

Can you tell me what I'm clearly not seeing? Sorry, again, if I'm just missing something obvious. If this is the current version of one of the assignments in `Apply`:

```go
	if p.Name != nil && *p.Name != group.Name {
		applied = true
		group.Name = *p.Name
	}
```

then if `p.Name` was a `string` instead of a `*string` then the patch would be
```go
if p.Name != group.Name {
    applied = true
    group.Name = p.Name
}
```
There must be something I'm not understanding.

gigawhitlocks: Or are we doing it like this because `nil` represents no change in a patch? 

In other words, and I don't know if this analogy helps, but if this was a functional language, `group.Name` in the above snipped might be type `Option[String]` of either type `Some[String]` or `None`?

gigawhitlocks:
Looks really good, besides that one point of confusion which I'm 90% sure is just me being confused, everything else looks great as far as I can tell :tada:

model/group_request.go
gabrieljackson: Yes, exactly. `nil` is way to "fall through" if no value is there and gives you another option beyond empty string.

We do this elsewhere like in the Mattermost server config. I think the intent was to have the nil check+patch logic for all types of cloud server object updates, but that was forgotten up till now.

Either way, yes, `nil` represents no change.

cmd/cloud/group.go
```diff
@@ -21,6 +22,7 @@ func init() {
 	groupUpdateCmd.Flags().String("description", "", "An optional description for this group of installations.")
 	groupUpdateCmd.Flags().String("version", "", "The Mattermost version for installations in this group to target.")
 	groupUpdateCmd.Flags().String("image", "mattermost/mattermost-enterprise-edition", "The Mattermost Container image to use.")
+	groupUpdateCmd.Flags().Int64("max-rolling", 0, "The maximum number of installations that can be updated at one time when a group is updated")
```
cpanato, line 25: why is zero here?

gabrieljackson: Ah, I needed a value here that basically meant "don't change the original value". The value `1` isn't good because you may want to actually set max rolling to 1.

I can do a follow-up PR if we find this confusing going forward.

PR #180, opened 2020-03-30
Version should actually not be required for update

gigawhitlocks:
`--version` is marked as a required field for `cloud installation update` and it should not be, as the user may only wish to use `--mattermost-env`. Perhaps there should be a check to make sure one of the flags is used, but in any case, the tool shouldn't force you to specify `--version` in this case.

Pretty sure this was my mistake originally, whoops!


gigawhitlocks:
Yup, that makes sense. Want me to open a story to track this? It's obviously super minor.

gabrieljackson:
I am 0/5 on opening a ticket unless it starts to eat up some serious time.

gigawhitlocks:
OK I'm going to just leave this PR open and I'll do the extra testing needed to carry this over the finish line after the other stuff that's higher priority today.

gabrieljackson:
I just realized that not having a group version will be a requirement for using groups with the cloud plugin. I am going to spike on doing the work I mentioned in my earlier comment.

gigawhitlocks:
Oh that does certainly change things. Do you want me to close this PR?

gabrieljackson:
I am going to close this as this concern is covered in https://github.com/mattermost/mattermost-cloud/pull/193

PR #181, opened 2020-03-30
MM-23550 Add DELETE permission for S3 buckets for cloud user

gigawhitlocks:
Resolves [MM-23550](https://mattermost.atlassian.net/browse/MM-23550) by adding a `DELETE` permission to the IAM user created for a cloud instance at creation time. MM-23550 was caused by a 403 issued by AWS due to the missing permission on this user.

One thing to note, that I'm not sure about, is that I am seeing a couple of error logs generated by the AWS library that I don't see without this change:

```
INFO[2020-03-30T17:13:34-05:00] Provisioning AWS S3 filestore                 installation=j5grybimefb5jqnysefxtcrhna instance=4jj1shoyn3riipqa37h7ztarbw
ERRO[2020-03-30T17:13:34-05:00] [aws] POST 404 Not Found https://iam.amazonaws.com/ {"UserName":"cloud-j5grybimefb5jqnysefxtcrhna"}  aws-operation-name=GetUser aws-service-id=IAM instance=4jj1shoyn3riipqa37h7ztarbw tools-aws=client
ERRO[2020-03-30T17:13:35-05:00] [aws] POST 404 Not Found https://iam.amazonaws.com/ {"PolicyArn":"arn:aws:iam::926412419614:policy/cloud-j5grybimefb5jqnysefxtcrhna"}  aws-operation-name=GetPolicy aws-service-id=IAM instance=4jj1shoyn3riipqa37h7ztarbw tools-aws=client
```

However, debugging shows that these come from a part of the code where we attempt to get a user to see if it exists and then the actions taken if it does not exist are part of normal execution, so these ERROR logs are themselves in error, but I'm not exactly sure how to suppress them, because they bubble up out of the `aws-sdk-go` library.

### DIFF
--- a/internal/tools/aws/iam.go
+++ b/internal/tools/aws/iam.go
@@ -168,6 +168,7 @@ func (a *Client) iamEnsurePolicyCreated(awsID, policyARN string, logger log.Fiel
 					"s3:PutObject",
 					"s3:ListBucket",
 					"s3:PutObjectAcl",
+					"s3:DeleteObject",
 				},
 				Resource: fmt.Sprintf("arn:aws:s3:::%s/*", awsID),
 			},


gigawhitlocks:
Unfortunately I think it almost definitely means we have been leaking buckets

gabrieljackson:
Okay, this may be an early warning that we should prioritize E2E tests for AWS in the shorter term. Something to discuss with the team.

lieut-data:
@gigawhitlocks, has this change been deployed yet? @prapti reported the same issue recently as part of https://mattermost.atlassian.net/browse/MM-24374.

gigawhitlocks:
I would think so but I'm not sure. @cpanato and/or @stylianosrigas should know for sure.

stylianosrigas:
@gigawhitlocks @lieut-data It is merged in test, so yeah the latest master is running for cloud provisioner.

cpanato:
@gigawhitlocks @lieut-data @prapti yes it is deployed, but it is valid only for new test servers

lieut-data:
@prapti, do you know when the server in question for https://mattermost.atlassian.net/browse/MM-24374 was originally created?

prapti:
> @prapti, do you know when the server in question for https://mattermost.atlassian.net/browse/MM-24374 was originally created?

@lieut-data The server was originally created around Apr 3 but updated last Thursday, Apr 16 with 5.22 updates.

lieut-data:
@prapti, thank you! The permissions would have been configured from the April 3rd date, likely before this hit `master`, even though the server has since been updated. If you destroy and re-create, it should now behave as expected.

prapti:
Ah, thanks @lieut-data and everyone else. Sorry for the obsolete bug confusion :)

PR #182, opened 2020-03-31
MM-22708: added support to encrypt RDS database at rest

gsagula:
#### Summary
This PR adds support for encrypting RDS instances per installation. It also extends AWS tools by adding a method that allows creating symmetric encryption key.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-22708


Makefile
```diff
@@ -119,8 +119,8 @@ mocks:
 	$(GOPATH)/bin/mockgen -source $(AWS_SDK_PATH)/service/acm/acmiface/interface.go -package mocks -destination ./internal/mocks/aws-sdk/acm.go
 	$(GOPATH)/bin/mockgen -source $(AWS_SDK_PATH)/service/iam/iamiface/interface.go -package mocks -destination ./internal/mocks/aws-sdk/iam.go
 	$(GOPATH)/bin/mockgen -source $(AWS_SDK_PATH)/service/route53/route53iface/interface.go -package mocks -destination ./internal/mocks/aws-sdk/route53.go
+	$(GOPATH)/bin/mockgen -source $(AWS_SDK_PATH)/service/kms/kmsiface/interface.go -package mocks -destination ./internal/mocks/aws-sdk/kms.go
```
gabrieljackson, line 122: This is unrelated to this PR, but do we want to add a CI/CD check that runs `make mocks` and checks that there is no difference between the mocks generated and the ones in the PR? We do this in the operator for the generated code and manifests. I think it would help us catch errors when we forgot to generate mocks.

internal/tools/aws/database.go
```diff
@@ -183,7 +184,12 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
 		return err
 	}
 
-	err = d.client.rdsEnsureDBClusterCreated(awsID, *vpcs[0].VpcId, rdsSecret.MasterUsername, rdsSecret.MasterPassword, logger)
+	encryptionKey, err := d.client.kmsCreateSymmetricKey(awsID, KMSAliasNameRDS(awsID), "Key used for encrypting RDS database")
```
gabrieljackson, line 187: What happens if the key is created on an attempt and then an error happens later? When the provisioner runs this provision flow again, will is return an error or will the encryption key be returned on the already-created key?

gsagula:
> I am wondering if we are cleaning up KMS keys on database teardown? I don't see any steps to remove the key so we will slowly leak keys.

Do we want to request deletion on keys? Keep in mind that we will need to keep the keys for as long as snapshots exist if we want to restore the customer's data. Since when/how we want to delete keys will require more thinking, can we keep it out of this PR?

gabrieljackson:
So, we already have a `keep-data` flag for RDS and S3. I would think we could just use this option to either delete the KMS key or not. The database resources are left or cleaned up based on this flag.

I would think it would be safe to clean up the KMS key based on this option or not.

gsagula:
> So, we already have a `keep-data` flag for RDS and S3. I would think we could just use this option to either delete the KMS key or not. The database resources are left or cleaned up based on this flag.

I think we clean up the resource and snapshot on database deletion. I have to check that but we should only delete the key when absolutely no data is retained.

Makefile
gsagula: I think it's a great idea but I may just throw a ticket in the backlog for that.

internal/tools/aws/database.go
gsagula: No, even if [CreateAlias](https://github.com/mattermost/mattermost-cloud/pull/182/files#diff-d5ff833767547cbc731fa146b4d30688R20) fails it will still be able to re-run. However, if creating alias failed we should schedule deletion for the created key.

Makefile
gabrieljackson: Yep, sounds good.

gabrieljackson:
I agree that we want to keep the keys until we absolutely don't have any snapshots or data those keys would decrypt.

My desire to cleanup keys is more about getting that code in and exercising it in our testing environment to make sure it is working correctly. In production, you are right that we will have a vetted flow for when and how to delete the data + keys.

Either way, this can be done in a follow-up PR.

Tagging @jwilander for additional thoughts if he has any.

gsagula:
> My desire to cleanup keys is more about getting that code in and exercising it in our testing environment to make sure it is working correctly. In production, you are right that we will have a vetted flow for when and how to delete the data + keys.

It makes sense @gabrieljackson . I will add the code for scheduling the deletion. I hope that deletion doesn't require any special permissions though.

gsagula:
I added a few methods that may facilitate the usage of this api in the future. PTAL

internal/tools/aws/database.go
```diff
@@ -183,7 +184,20 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
 	}
 
-	err = d.client.rdsEnsureDBClusterCreated(awsID, *vpcs[0].VpcId, rdsSecret.MasterUsername, rdsSecret.MasterPassword, logger)
+	encryptionKey, err := d.client.kmsCreateSymmetricKeyWithAlias(awsID, KMSAliasNameRDS(awsID), "Key used for encrypting RDS database")
+	if err != nil && !IsErrorCode(err, kms.ErrCodeAlreadyExistsException) {
+		// If key is created but alias did not, this key should be scheduled for deletion.
+		if encryptionKey != nil {
+			deletionKeyErr := d.client.kmsScheduleKeyDeletion(*encryptionKey.KeyId, 7)
+			if deletionKeyErr != nil {
+				logger.WithError(deletionKeyErr).Warnf("Failed to schedule encryption key %s for deletition", *encryptionKey.KeyId)
```
gabrieljackson, line 193:
```suggestion
				logger.WithError(deletionKeyErr).Warnf("Failed to schedule encryption key %s for deletion", *encryptionKey.KeyId)
```

Also, do we want to log this at Error level instead?

internal/tools/aws/database.go
```diff
@@ -183,7 +184,20 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+			}
+		}
+
+		return errors.Wrapf(err, "unable provision RDS database for installation %s", installationID)
```
gabrieljackson, line 197: This error should probably be a bit more specific like:
```suggestion
		return errors.Wrapf(err, "unable to create RDS encryption key for installation %s", installationID)
```

internal/tools/aws/kms.go
```diff
@@ -0,0 +1,74 @@
...
+func (a *Client) kmsScheduleKeyDeletion(keyID string, days int64) error {
+	input := kms.ScheduleKeyDeletionInput{
+		KeyId: aws.String(keyID),
+	}
+
+	if days == -1 {
+		input.PendingWindowInDays = aws.Int64(DefaultScheduledEncryptionKeyDeletion)
+	} else {
+		input.PendingWindowInDays = aws.Int64(days)
+	}
```
gabrieljackson, line 66: Is it worth checking for other negative values and error handling them in some way or just let the upstream AWS call handle that? (assuming you can't provide a negative value there)

internal/tools/aws/kms.go
```diff
@@ -0,0 +1,74 @@
...
+
+	_, err := a.Service().kms.ScheduleKeyDeletion(&input)
+	if err != nil {
+		return errors.Wrapf(err, "unabled to describe encryption key ID %s", keyID)
```
gabrieljackson, line 70:
```suggestion
		return errors.Wrapf(err, "unable to schedule deletion of encryption key ID %s", keyID)
```

internal/tools/aws/database.go
gsagula: Yes, I think error is better.

internal/tools/aws/kms.go
gsagula: I'm not sure how they handle negative numbers, but upstream should return error for any number less than 7 days. Looking at the code again, I just realized that my 45 days default will also produce an error. Number should be between 7 and 30 days. I will fix it :)

https://github.com/aws/aws-sdk-go/blob/2d94d44e83349d20c26033450e35d295a3f77eb1/service/kms/api.go#L13442

internal/tools/aws/constants.go
```diff
@@ -108,6 +108,9 @@ const (
 	// DefaultAWSClientRetries supplies how many time the AWS client will retry a failed call.
 	DefaultAWSClientRetries = 3
 
-	// DefaultScheduledEncryptionKeyDeletion is the default number of days that AWS will take to delete an encryption key.
-	DefaultScheduledEncryptionKeyDeletion = 45
+	// KMSMaxTimeEncryptionKeyDeletion is the maximum number of days that AWS will take to delete an encryption key.
+	KMSMaxTimeEncryptionKeyDeletion = 30
+
+	// KMSMinTimeEncryptionKeyDeletion is the minimum number of days that AWS will take to delete an encryption key.
+	KMSMinTimeEncryptionKeyDeletion = 7
```
gabrieljackson, line 115: I don't see these used anywhere. Is that intended?

gabrieljackson: Sorry, meant I don't see the first `KMSMaxTimeEncryptionKeyDeletion` value used anywhere.

gsagula: The intent is to be used in the same way that KMSMinTimeEncryptionKeyDeletion but I guess we can delete it if not being used.

gabrieljackson: I just wanted to make sure. Thanks

PR #186, opened 2020-04-01
Improve group leave logic

gabrieljackson:
 - Provides an option when leaving groups for installations to
   retain the group-merged configuration that was used while in
   the group or to revert back to the original config. Default
   behavior is to retain group-merged config.
 - Prevent leaving groups when installations are in invalid states.
 - When leaving groups, installations are forced into an update-
   requested state to ensure configuration is correct.

https://mattermost.atlassian.net/browse/MM-22505


internal/api/installation.go
```diff
@@ -325,21 +325,57 @@ func handleJoinGroup(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			// so, we will get a merged copy of the installation out and will
+			// manually update the necessary values.
+			mergedInstallation, err := c.Store.GetInstallation(installationID, true, false)
+			if err != nil {
+				c.Logger.WithError(err).Error("failed to get group-merged installation")
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+
+			installation.Version = mergedInstallation.Version
```
gigawhitlocks, line 375: Refresh my memory here, but in normal operation, is the merge config written to the Installation struct, overwriting whatever configuration may have been there before the Installation joined the group? Or is it safe to assume that until this line, `installation.Version` will refer to the version the Installation had before joining the group, and that data is only actually overwritten here?

internal/api/installation_test.go
```diff
@@ -779,21 +783,49 @@ func TestLeaveGroup(t *testing.T) {
...
 
-		err = client.LeaveGroup(installation1.ID)
+		err = client.LeaveGroup(installation1.ID, &model.LeaveGroupRequest{RetainConfig: true})
 		require.EqualError(t, err, "failed with status code 409")
 	})
 
 	t.Run("while in group 1", func(t *testing.T) {
-		err = client.LeaveGroup(installation1.ID)
-		require.NoError(t, err)
+		t.Run("don't retain group config", func(t *testing.T) {
```
gigawhitlocks, line 791: What is the outcome of a test where an Installation has configuration A and is joined to a group with configuration B and then leaves the group with `retain-config == true`, as opposed to this test which sets these values once the Installation is already part of the Group?

internal/api/installation.go
gabrieljackson: The answer is: it totally depends on how you ask for the installation.

https://github.com/mattermost/mattermost-cloud/blob/c1621a2453a697d1d830be07f2dd7bf90f35a90e/internal/store/installation.go#L61

The function for getting installations will do whatever you tell it to so you can ask for what you want.

internal/store/installation.go
```diff
@@ -197,15 +197,15 @@ func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation) e
 		SetMap(map[string]interface{}{
 			"ID":               installation.ID,
 			"OwnerID":          installation.OwnerID,
+			"GroupID":          installation.GroupID,
+			"GroupSequence":    nil,
```
gigawhitlocks, line 201: Just double-check that we want `nil` here and not an `int` value since `SetMap` takes a `map[string]interface{}` so the compiler won't enforce types signatures here. It's not obvious to me which would be correct and it'd be an easy mistake to make -- feel free to just resolve this comment once you read it if this is as-intended :)

internal/api/installation_test.go
gabrieljackson: Installations are not directly updated when they join a group so their original config remains. When leaving the group, you have the option of updating the installation to the group-merged version as it leaves (resulting in a no-op basically) or just leaving without the merge which could result in it updating its behavior.

The tests here try to cover both options. I think that answered your question?

internal/store/installation.go
gigawhitlocks, line 201: Just out of curiosity, for the sake of style could we start sorting these sorts of object definitions alphabetically unless there's a more obvious sorting order, just so that changes like this are less likely to occur (where some lines just move for no real reason)?

model/client.go
```diff
@@ -605,8 +605,15 @@ func (c *Client) JoinGroup(groupID, installationID string) error {
 }
 
 // LeaveGroup removes an installation from its group, if any.
-func (c *Client) LeaveGroup(installationID string) error {
-	resp, err := c.doDelete(c.buildURL("/api/installation/%s/group", installationID))
+func (c *Client) LeaveGroup(installationID string, request *LeaveGroupRequest) error {
+	u, err := url.Parse(c.buildURL("/api/installation/%s/group", installationID))
+	if err != nil {
+		return err
```
gigawhitlocks, line 611: Would there be any benefit to wrapping this error?

internal/api/installation_test.go
gigawhitlocks: Maybe I'm wrong but the test as-written doesn't seem like it actually tests leaving without the merge because the data is written to the Installation after it's joined the group. 

My concern is that when the Installation is joined to the group, the configuration held by the Installation is overwritten by the Group's config (the merged config). Then, when the Installation is moved out of the Group, we "preserve" the Installation config, but it was already overwritten on Join.

internal/api/installation.go
gigawhitlocks: OK cool so we don't ever overwrite the values in the Installation fields themselves, new values are always stored in the Group and we just check the overrides values (I do remember this from your previous PRs). Perfect.

internal/api/installation_test.go
gigawhitlocks: After reading your answer here https://github.com/mattermost/mattermost-cloud/pull/186#discussion_r401795700 and thinking a bit more, I think that this test is _probably_ fine as-is but I'll leave my concern above just in case I did spot something you think is worthwhile.

internal/api/installation.go
gabrieljackson: Yeah, to be super clear. If you update the installation directly, the installation DB object is updated. For group updates, it's the group object that is updated. Then, when acting on installations it's always the merged set of what both are at that moment in time.

internal/api/installation_test.go
gabrieljackson: Oh, there are plenty of other tests in various places that check installation config merge behavior when joined to groups if that is the primary concern.

This test is in the leave group test suite and so I was just focusing on that flow working correctly.

internal/store/installation.go
gabrieljackson: So, I did move that because I thought it made sense to have the various "ID"s in one spot near the top. That at least made it easier to read the JSON when I am listing installations so I figured we would match it here. Totally 0/5 on alphabetical or other ordering if that is what we want to do for store calls, but I do think trying to "logically" display the JSON payloads is helpful.

model/client.go
gabrieljackson: Probably, noticed we don't really wrap errors much in the API package we also typically log in lowercase letters for w/e reason. lol

Maybe the package should just have a follow-up clean and better-logging PR. Good feedback though.

PR #188, opened 2020-04-03
Allow for empty image and/or version group values

gabrieljackson:
This opens up groups to be used for specific purposes such as only
setting environment variables for the installations in the group.

This change also fixes a bug on group creation where max rolling
was not set correctly.

https://mattermost.atlassian.net/browse/MM-23840

### DIFF
--- a/model/group_request.go
+++ b/model/group_request.go
@@ -28,14 +28,8 @@ func (request *CreateGroupRequest) SetDefaults() {
 
 // Validate validates the values of a group create request.
 func (request *CreateGroupRequest) Validate() error {
-	if request.Name == "" {
+	if len(request.Name) == 0 {
 		return errors.New("must specify name")
-	}
-	if request.Version == "" {
-		return errors.New("must specify version")
-	}
-	if request.Image == "" {
-		return errors.New("must specify image")
 	}
 	if request.MaxRolling < 1 {
 		return errors.New("max rolling must be 1 or greater")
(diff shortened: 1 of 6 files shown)


model/group_request.go
```diff
@@ -126,6 +120,9 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 
 // Validate validates the values of a group patch request
 func (p *PatchGroupRequest) Validate() error {
+	if p.Name != nil && len(*p.Name) == 0 {
+		return errors.New("must specify name")
```
gigawhitlocks, line 124: no error if `p.Name == nil` ?

model/group_request.go
```diff
@@ -126,6 +120,9 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 
 // Validate validates the values of a group patch request
 func (p *PatchGroupRequest) Validate() error {
+	if p.Name != nil && len(*p.Name) == 0 {
```
angeloskyratzakos, line 123: If the value is `nil`, it won't be caught here, right?

angeloskyratzakos: maybe we need `||`?

gabrieljackson: Thanks for checking on this. I believe this logic is correct after reviewing it. We are basically checking if p.Name is not nil and then if it isn't nil we see if the value is blank `""`. If this line turns out to be true it means the update request is trying to update the name of the group to "", which we should return an error on.

PR #189, opened 2020-04-06
MM-21930: Added api for restoring a RDS database from the most recent snapshot

gsagula:
#### Summary

This PR enhances the database migration api for restoring a RDS DB instance from the most recent snapshot taken. By doing that, there is no need to persist any data regarding snapshots. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-21930


internal/tools/aws/session.go
```diff
@@ -37,10 +37,10 @@ func NewAWSSessionWithLogger(config *aws.Config, logger log.FieldLogger) (*sessi
 				"aws-operation-name": r.Operation.Name,
 			})
 
-			if r.HTTPResponse.StatusCode >= 400 {
+			if r.HTTPResponse.StatusCode >= 500 {
```
gigawhitlocks, line 40: I've been meaning to address this, but in this part of the code, we should just always log `logger.Debug` because sometimes an error from AWS isn't an error in our application and logging `Error` here bubbles up the error as though it is an application-layer error. So I think we should get rid of these conditionals and just `logger.Debug(buffer.String())` here no matter what. What do you think?

gsagula: It wasn't supposed to be in the commit. I agree with having it only active for debugging. It was initially like that.

gabrieljackson:
@gsagula, looks like we have a merge conflict. I will review as soon as that is resolved.

gsagula:
Thanks @gabrieljackson I'm looking into that!

jwilander:
Closing as this isn't priority anymore

PR #190, opened 2020-04-06
Scan image for security vulnerabilities 

cpanato:
#### Summary
Add security image scan, in a very similar way that we do in GitLab.
you can check the scan report in the `scan image` step and there in the artifact section to get the report.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,6 @@ jobs:
         docker tag mattermost/mattermost-cloud:test mattermost/mattermost-cloud:$TAG
         docker push mattermost/mattermost-cloud:$TAG
 
-
   push-docker:
     executor:
       name: default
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -116,6 +134,10 @@ workflows:
         - check-style
         - test-sqlite
         - test-postgres
+    - scan-image:
+        context: matterbuild-docker
+        requires:
+        - push-docker-pr
```
gabrieljackson, line 140: Do we want to scan the image before pushing it? Not sure we would want to start blocking pushes right now if the scan failed, but also not sure what we have in mind for scan automation.

cpanato: we need to push because the scanner fetches from the registry. however i can see if that works without pushing

PR #191, opened 2020-04-06
dockerfile: add cert-manager manifests to the image

cpanato:
#### Summary
add missing manifests to the image


cpanato:
we are deploying in test and notice this was missing. to not block our deployment we just request one review and we are merging when green

PR #193, opened 2020-04-06
Refactor installation update logic

gabrieljackson:
 - Changes the installation update logic to use patching like what
   is done with groups. This removes the requirement for passing
   in more installation update values than those being used to
   actually affect the installation. In doing so, the API endpoint
   is now easier to use and less error-prone.
 - Changes the installation update API endpoint to handle change
   detection better. Previously, all installation updates would
   trigger a state change. Now, only actual updates will do so.
 - Corrects installation update logging referencing only version
   updates.
 - Both the installation and group update endpoints now return
   the patched object as stored in the database for review.

https://mattermost.atlassian.net/browse/MM-23904

### DIFF
--- a/cmd/cloud/group.go
+++ b/cmd/cloud/group.go
@@ -93,12 +93,7 @@ var groupCreateCmd = &cobra.Command{
 			return errors.Wrap(err, "failed to create group")
 		}
 
-		err = printJSON(group)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return printJSON(group)
 	},
 }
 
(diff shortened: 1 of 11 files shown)


cmd/cloud/group.go
```diff
@@ -93,12 +93,7 @@ var groupCreateCmd = &cobra.Command{
 			return errors.Wrap(err, "failed to create group")
 		}
 
-		err = printJSON(group)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return printJSON(group)
```
gigawhitlocks, line 96: I thought this (this whole hunk) was against our style guide for some reason..

gabrieljackson: I think I vaguely remember talking about this before and hope I wasn't confusing with my thoughts at the time. Basically, we don't do this normally because we try to wrap errors almost always. Because there was no wrapping I just shortened it here.

That said, I probably shouldn't have done it and instead wrapped the error.

gigawhitlocks: Yeah I definitely got confused before; thanks for the clarification of the style rule!

gabrieljackson: No worries! Definitely call me out if I am saying one thing and doing another.

cpanato:
this will go to test env as soon we promote the existing release to the other envs

PR #195, opened 2020-04-15
Check if group is deleted before joining installations

gabrieljackson:
We were already checking that groups had no installation members
before deleting them, but this change ensures groups are not
deleted before joining an installation to them.

https://mattermost.atlassian.net/browse/MM-24163

Release Note:
```release-note
Check group deletion state before joining installation
```

### DIFF
--- a/internal/api/installation.go
+++ b/internal/api/installation.go
@@ -110,15 +110,19 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	var group *model.Group
+	var status int
+	groupUnlockOnce := func() {}
 	if len(createInstallationRequest.GroupID) != 0 {
-		group, err := c.Store.GetGroup(createInstallationRequest.GroupID)
-		if err != nil {
-			c.Logger.WithError(err).Error("failed to query group")
-			w.WriteHeader(http.StatusInternalServerError)
+		group, status, groupUnlockOnce = lockGroup(c, createInstallationRequest.GroupID)
+		if status != 0 {
+			w.WriteHeader(status)
 			return
 		}
-		if group == nil {
-			w.WriteHeader(http.StatusNotFound)
+		defer groupUnlockOnce()
+		if group.IsDeleted() {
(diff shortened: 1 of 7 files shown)


internal/api/installation.go
```diff
@@ -110,15 +110,19 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
...
+	var group *model.Group
+	var status int
+	groupUnlockOnce := func() {}
 	if len(createInstallationRequest.GroupID) != 0 {
-		group, err := c.Store.GetGroup(createInstallationRequest.GroupID)
-		if err != nil {
-			c.Logger.WithError(err).Error("failed to query group")
-			w.WriteHeader(http.StatusInternalServerError)
+		group, status, groupUnlockOnce = lockGroup(c, createInstallationRequest.GroupID)
+		if status != 0 {
```
gigawhitlocks, line 118: Do we want a log message here maybe?

gabrieljackson: The locker has its own logger.

model/group.go
```diff
@@ -29,14 +29,19 @@ type GroupFilter struct {
 }
 
 // Clone returns a deep copy the group.
-func (c *Group) Clone() *Group {
+func (g *Group) Clone() *Group {
 	var clone Group
-	data, _ := json.Marshal(c)
+	data, _ := json.Marshal(g)
```
gigawhitlocks, line 34: :pray:

model/cluster_installation_test.go
```diff
@@ -126,7 +126,7 @@ func TestClusterInstallationsFromReader(t *testing.T) {
 		]`)))
 		require.NoError(t, err)
 		require.Equal(t, []*ClusterInstallation{
-			&ClusterInstallation{
+			{
```
gigawhitlocks, line 129: hey cool I didn't know you could omit the type like this, that's nice

gabrieljackson: Yeah, my linter started complaining about this stuff so I am fixing it when I see it.

PR #196, opened 2020-04-16
MM-24209: enhanced RDS encryption at rest

gsagula:
#### Summary

Few improvements to the tools/aws/database:

-  Added logic to check when alias name already exists and then reuse the encryption key
- Fixed bug which KMS return errors were being wrapped and then failing on awserror type assertion.
- Check state of the key before using it.
- Added more tests for different use cases involving encryption key. 

#### Ticket Link

Fixes: https://mattermost.atlassian.net/browse/MM-24209

### DIFF
--- a/internal/tools/aws/database.go
+++ b/internal/tools/aws/database.go
@@ -136,6 +136,7 @@ func (d *RDSDatabase) GenerateDatabaseSpecAndSecret(logger log.FieldLogger) (*mm
 
 func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.FieldLogger) error {
 	awsID := CloudID(installationID)
+
 	logger.Infof("Provisioning AWS RDS database with ID %s", awsID)
 
 	// To properly provision the database we need a SQL client to lookup which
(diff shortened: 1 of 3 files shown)


internal/tools/aws/database.go
```diff
@@ -184,22 +185,35 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+		if !IsErrorCode(err, kms.ErrCodeNotFoundException) {
+			return errors.Wrapf(err, "unable to verify encryption key %s for the RDS cluster ID %s", keyAliasName, awsID)
+		}
 
-	err = d.client.kmsCreateAlias(*encryptionKey.KeyId, KMSAliasNameRDS(awsID))
-	if err != nil && !IsErrorCode(err, kms.ErrCodeAlreadyExistsException) {
-		deletionKeyErr := d.client.kmsScheduleKeyDeletion(*encryptionKey.KeyId, KMSMinTimeEncryptionKeyDeletion)
-		if deletionKeyErr != nil {
-			logger.WithError(deletionKeyErr).Errorf("Failed to schedule encryption key %s for deletition", *encryptionKey.KeyId)
+		keyMetadata, err = d.client.kmsCreateSymmetricKey("Key used for encrypting RDS database")
```
gabrieljackson, line 195: nit: it would be really nice to include the installation ID in the key description

internal/tools/aws/database.go
```diff
@@ -184,22 +185,35 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
 
-		return errors.Wrapf(err, "unable to create a RDS encryption key alias name for installation %s", installationID)
+		err = d.client.kmsCreateAlias(*keyMetadata.KeyId, keyAliasName)
+		if err != nil {
+			scheduleErr := d.client.kmsScheduleKeyDeletion(*keyMetadata.KeyId, KMSMinTimeEncryptionKeyDeletion)
+			if scheduleErr != nil {
+				logger.WithError(scheduleErr).Errorf("Failed to schedule deletion for the encryption key %s", keyAliasName)
+			}
+			return errors.Wrapf(err, "unable to create an encryption key alias name for the RDS cluster ID %s", awsID)
+		}
```
gabrieljackson, line 207: This is the part that confuses me. If I understand the flow of this PR, it is:

1. Lookup KMS key alias
2. I that doesn't exist, create a new KMS key
3. Create an alias for that new key
4. If that fails, try to quickly schedule the key for deletion and long any errors.

The problem I see with this is that keys can still leak if step 2 works, but 3 fails.

This seems to stem from this design choice: https://github.com/mattermost/mattermost-cloud/blob/a424a80e2ac9d9545760bdc620cf0ecd60e86e76/internal/tools/aws/kms.go#L21-L22

If we were to create a tag on the KMS key when it was created, we should be able to look it up on follow-up attempts and not need to try to schedule it for deletion if anything goes wrong. This is how we make the other AWS flows resilient to failure. Is there a reason we can't do the same thing here?

jwilander:
@gsagula can you prioritize a follow-up PR to address @gabrieljackson's comment?

PR #197, opened 2020-04-16
Disable RDS encryption logic

gabrieljackson:
This will be re-enabled at a later time.

I did a single test with this locally and it seems to work, but please double-check what I commented out @gsagula

### DIFF
--- a/internal/tools/aws/database.go
+++ b/internal/tools/aws/database.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


cpanato:
what is the impact of the current deployments and the current provisioned? @gabrieljackson

gabrieljackson:
They should be torn down in a similar way to before. Logic still needs to be added to the teardown RDS flow to mark keys for deletion, but that is not currently in master.

internal/tools/aws/database.go
```diff
@@ -185,35 +184,35 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
 		return err
 	}
 
-	keyAliasName := KMSAliasNameRDS(awsID)
-	keyMetadata, err := d.client.kmsGetSymmetricKey(keyAliasName)
-	if err != nil {
-		if !IsErrorCode(err, kms.ErrCodeNotFoundException) {
-			return errors.Wrapf(err, "unable to verify encryption key %s for the RDS cluster ID %s", keyAliasName, awsID)
-		}
+	// keyAliasName := KMSAliasNameRDS(awsID)
```
gsagula, line 187: nit: no commented code. Just delete it and I can add it again.

PR #202, opened 2020-04-20
MM-24209: added support to create and tear down encryption keys with leaking resources

gsagula:
#### Summary

This PR addresses the issue of leaking encryption keys. A new service is used to locate KMS resources by tag, so create a key alias is no longer needed. See key features for more details. 

```release-note
 * Create encryption keys with tags.
 * Find encryption keys using resource groups tagging api.
 * Reuse already existent keys when creating RDS instance.
 * Validating that only one key in Enabled state exists when creating instance.
 * Remove keys when installation is deleted.
 * Eliminate the need for creating key alias which was the source for leaking KMS resources.
```

#### Testing

* unit test (acceptance) for provisioning RDS 
* unit test for logic around get all resources by tag
* (PENDING) corner cases, including unit tests for tearing down RDS (will be added after review)
* manual testing: setup and tear down RDS, verify that no keys are leaked on errors. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24209


internal/tools/aws/database.go
```diff
@@ -66,7 +68,27 @@ func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
...
 
-	logger.Debug("AWS RDS DB cluster deleted")
+	logger.Debug("AWS RDS db cluster deleted")
+
+	resourceNames, err := d.getKMSResourceNames(awsID)
+	if err != nil {
+		return errors.Wrapf(err, "unabled to get KMS resources associated with db cluster %s", awsID)
+	}
+
+	if len(resourceNames) > 0 {
```
gigawhitlocks, line 78: The `else` in this statement can be eliminated by flipping this condition and returning early if `len(resourceNames) == 0`

internal/tools/aws/database.go
```diff
@@ -66,7 +68,27 @@ func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
 		return errors.Wrap(err, "unable to delete RDS DB cluster")
 	}
 
-	logger.Debug("AWS RDS DB cluster deleted")
+	logger.Debug("AWS RDS db cluster deleted")
```
gigawhitlocks, line 71: I think this log message should be either removed or altered so that the stage that is complete here is reported in a way that is more distinct from the log message on line 91

internal/tools/aws/database.go
```diff
@@ -176,43 +200,43 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+		return err
+	}
 
-	// if *keyMetadata.KeyState != kms.KeyStateEnabled {
-	// 	return errors.Errorf("expected encryption key %s state to be Enabled, but it was %s", keyAliasName, *keyMetadata.KeyState)
-	// }
+	var keyMetadata *kms.KeyMetadata
+	if len(kmsResourceNames) > 0 {
+		enabledKeys := d.getEnabledEncryptionKeys(kmsResourceNames, logger)
+		switch len(enabledKeys) {
```
gigawhitlocks, line 219: Do we expect there to be future `case` statements added to this switch? If not, could this just be an `if/else`? It's kind of weird as a switch statement. 0/5.

internal/tools/aws/database.go
```diff
@@ -176,43 +200,43 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+				TagValue: aws.String(awsID),
+			},
+		})
+		if err != nil {
+			return errors.Wrapf(err, "unable to create an encryption key for RDS db cluster %s", awsID)
+		}
+	}
 
-	// logger.Infof("Encrypting RDS database with key %s", keyAliasName)
+	logger.Infof("Encrypting RDS database with key %s", *keyMetadata.Arn)
```
gigawhitlocks, line 237: Is there any chance here that `keyMetadata` or `*keyMetadata.Arn` might be `nil`? There might be a `nil pointer exception` lurking here.

gsagula: Unless there is a bug aws-sdk code, there is no way that the pointer to keyMetadata will be null here. I can guard this pointer but If we are going to be that defensive, then we will need to change a lot of code in this repo.

internal/tools/aws/database.go
```diff
@@ -224,3 +248,41 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+
+func (d *RDSDatabase) getEnabledEncryptionKeys(resourceNameList []*string, logger log.FieldLogger) []*kms.KeyMetadata {
+	var keys []*kms.KeyMetadata
+
+	for _, name := range resourceNameList {
+		keyMetadata, err := d.client.kmsGetSymmetricKey(*name)
+		if err != nil {
+			logger.WithError(err).Error("Failed get some of the encryption key")
+			continue
+		}
```
gabrieljackson, line 281: Only logging an error here and not returning seems dangerous. What if there are 1 or more keys that would have been returned that would have altered the rest of the flow?

internal/tools/aws/database.go
gsagula: I was thinking about the same chage 👍

internal/tools/aws/database.go
```diff
@@ -66,7 +68,27 @@ func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
...
+	}
+
+	if len(resourceNames) > 0 {
+		enabledKeys := d.getEnabledEncryptionKeys(resourceNames, logger)
+		for _, keyMetadata := range enabledKeys {
+			err = d.client.kmsScheduleKeyDeletion(*keyMetadata.KeyId, KMSMinTimeEncryptionKeyDeletion)
+			if err != nil {
+				return errors.Wrapf(err, "encryption key associated with db cluster %s could not be scheduled for deletion", awsID)
+			}
+			logger.Infof("Encryption key %s will be deleted in %d days", *keyMetadata.Arn, KMSMinTimeEncryptionKeyDeletion)
```
gabrieljackson, line 85: nit:
```suggestion
			logger.Infof("Encryption key %s scheduled for deletion in %d days", *keyMetadata.Arn, KMSMinTimeEncryptionKeyDeletion)
```

internal/tools/aws/database.go
```diff
@@ -176,43 +200,43 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
-	// }
+	var keyMetadata *kms.KeyMetadata
+	if len(kmsResourceNames) > 0 {
+		enabledKeys := d.getEnabledEncryptionKeys(kmsResourceNames, logger)
+		switch len(enabledKeys) {
+		case 1:
+			keyMetadata = enabledKeys[0]
+		default:
+			return errors.Errorf("db cluster %s should have exactly one enabled/active encryption key (found %d)", awsID, len(enabledKeys))
+		}
```
gabrieljackson, line 224: nit: a bit cleaner this way I think.
```suggestion
		enabledKeys := d.getEnabledEncryptionKeys(kmsResourceNames, logger)
		if  len(enabledKeys) != 1 {
			return errors.Errorf("db cluster %s should have exactly one enabled/active encryption key (found %d)", awsID, len(enabledKeys))
		}
		keyMetadata = enabledKeys[0]
```

internal/tools/aws/database.go
gsagula: That's a good point. I will change it. Thanks!

internal/tools/aws/database.go
gsagula: If I do that I will need to duplicate `logger.Debug("AWS RDS db cluster teardown completed")`. Also, I'd rather having only one returning point for teardown complete.

gsagula:
@gabrieljackson @gigawhitlocks Thanks for the review. I will add the rest of the tests as soon as it gets the second approval.

internal/tools/aws/database.go
gigawhitlocks: No need to be defensive at this level against bugs in library code, but you're the expert on the AWS libraries so it was not apparent to me whether it was possible for `d.client.kmsCreateSymmetricKey` to return `nil, nil` and I wanted to make sure that this was not an oversight. I do not see AWS specifying one way or the other in their documentation whether a `nil, nil` result is possible from `CreateKey`. Happy to trust you on this one.

PR #203, opened 2020-04-20
Always log to DEBUG level from AWS session

gigawhitlocks:
#### Summary
Changes the AWS session package to only log at DEBUG level. On `master` the behavior is such that we get ERROR level logs for any response from AWS >=400 level. While this does represent an HTTP error, there are a number of places where we expect an error from AWS under normal operation, so it becomes incorrect to bubble up these API errors as application errors.

This change changes all talk with AWS to DEBUG level so that the provisioner output only produces red error output in cases where there are application errors.

#### Ticket Link
No ticket link

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/tools/aws/session.go
+++ b/internal/tools/aws/session.go
@@ -37,12 +37,7 @@ func NewAWSSessionWithLogger(config *aws.Config, logger log.FieldLogger) (*sessi
 				"aws-operation-name": r.Operation.Name,
 			})
 
-			if r.HTTPResponse.StatusCode >= 400 {
-				logger.Error(buffer.String())
-			}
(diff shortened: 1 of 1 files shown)


gigawhitlocks:
Sorry about that, I must've missed that in your PR @gsagula (but I made this change locally for myself and figured I'd just open a tiny PR so it's not like I spent a long time doing this :P )

cpanato:
@gigawhitlocks add a release note fo this PR

gigawhitlocks:
It looks like you set one?

cpanato:
yes

 s/add/added

PR #208, opened 2020-04-25
MM-24541: added ability to include IAM policy document when creating symmetric encryption keys

gsagula:
#### Summary

Before this PR, encryption keys were being created with KMS service's default policy which only grants access to the root(account) user which was preventing the provisioner from accessing the encryption keys in some AWS accounts. CMKs are treated differently than other resources in AWS since they require an additional resource level policy for users accessing the key.

#### Release Note

```release-note
 * create AIM policy document 
 * look up for caller ARN using Service Token Service api
 * create 2 policy statements for IAM access to the encryption keys
 * first statement grants access to the root user so there is no risk for the key becomes unmanageable
 * second statement grants access to the caller ARN so provisioner can perform operations with the key
```

Fixes: https://mattermost.atlassian.net/browse/MM-24541

#### IMPORTANT 
This PR  introduces a new service that may require IAM permissions: [(STS) GetCallerIdentity](https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetCallerIdentity) 

#### DO NOT MERGE BEFORE:
https://github.com/mattermost/mattermost-cloud-monitoring/pull/101
cc @stylianosrigas 


jwilander:
Added the do not merge label until we get the permissions updated

gsagula:
@stylianosrigas Mind to advise me again on how to grant the same permissions as test to my IAM in dev so I can test this changes. I can't locate the policy in DEV. Thanks!

gsagula:
Tested this change in DEV using `arn:aws:iam::926412419614:user/gabriel.sagula-provisioner` (thanks @stylianosrigas for creating the test accounts). The only permissions that I had to add were for S3 and ELB. Surprisingly AWS didn't complain about (STS) GetCallerIdentity but regardless we have a PR for adding this permission. 
I still found an issue which KMS didn't like the permissions given to the caller in the key's policy document. Besides to full rights give to the root, KMS still requires the caller to have writing access to the key. The issue was resolved based on this [document](https://aws.amazon.com/premiumsupport/knowledge-center/update-key-policy-future/).

stylianosrigas:
@gsagula I am a little bit confused regarding the S3 and ELB permissions you mentioned. What are the exact permissions that I need to add to the Provisioner user?

internal/tools/aws/database.go
```diff
@@ -228,14 +230,85 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
 				TagKey:   aws.String(DefaultRDSEncryptionTagKey),
 				TagValue: aws.String(awsID),
 			},
-		})
+		}
+		keyMetadata, err = d.client.kmsCreateSymmetricKey(KMSKeyDescriptionRDS(awsID), keyTags, &keyPolicyDocument)
 		if err != nil {
-			return errors.Wrapf(err, "unable to create an encryption key for db cluster %s", awsID)
+			return errors.Wrapf(err,
+				"unable to create an encryption key for RDS DB cluster %s: key policy: %s", awsID, keyPolicyDocument)
```
gabrieljackson, line 311: Hmm, print the whole key policy document in the error? I would ask if we are sure we would want to do that or if this could be handled in a debug level message?

internal/tools/aws/database.go
```diff
@@ -228,14 +230,85 @@ func (d *RDSDatabase) rdsDatabaseProvision(installationID string, logger log.Fie
...
+					Sid:    "Allow full control of the key",
+					Effect: model.PolicyStatementEffectAllow,
+					Principal: map[string]string{
+						"AWS": rootARN.String(),
+					},
+					Action: []string{
+						"kms:*",
+					},
+					Resource: model.PolicyStatementAllResources,
+				},
```
gabrieljackson, line 270: This part makes me a little nervous. I don't see anything inherently wrong with how you constructed the `rootARN` object above, but is there any way this could go wrong? Have you thought through this in regards to possibly giving permissions to an AWS user incorrectly?

I am only trying to channel the security team here and ensure we have done our best to not open the code up to new security issues. As I said, though, this seems like it should be fine as long as the `root` user will always be one or our cloud-team-owned accounts.

jwilander:
As discussed during the call with @stylianosrigas and @gsagula this morning, we no longer need this

internal/tools/aws/database.go
gsagula: > Have you thought through this in regards to possibly giving permissions to an AWS user incorrectly?

That's what AWS does if you don't provide a policy to the key.
https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html
https://aws.amazon.com/premiumsupport/knowledge-center/update-key-policy-future/

> As I said, though, this seems like it should be fine as long as the root user will always be one or our cloud-team-owned accounts.

Yes, that's a good point and that's why I left a comment when constructing the ARN that this assumes the caller and the root in the same AWS account.

internal/tools/aws/database.go
gsagula: What was exactly your concern? I'm answering this questions because we may need this PR in the future.

PR #209, opened 2020-04-30
MM-23151: Added a new database abstraction to support multitenant RDS clusters

gsagula:
#### Summary
Before this PR, the only option available for creating installations with an RDS database was by creating an entire database cluster in AWS. The PR introduces a new command flag  `--database aws-multitenant-rds` that instructions the provisioner to look up for a pre instantiated DB cluster. By doing that, the provisioner can created multiple databases in a single RDS cluster instead of having one cluster per installation. This feature is need in order to support and scale a high volume of installations. In addition to improve scalability, this PR also cuts the installation time from ~15 minutes to less than 1 minute.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-23151

#### Release Note
```release-note
 * Added a new database implementation for handling multiple tenants in a single RDS DB cluster.
 * Added new table MultitenantDatabase so provisioner can keep track of database installations in each RDS cluster.
 * Provisioner will always do lookups in the MultitenantDatabase table first and, if nothing is available, it will try to fetch the resources directly from RDS via resource tags.
 * RDS tags are defined in `tools/aws/constants.go`, however the provisioner does not create them in AWS since this work is done via Terraform.
 * Provisioner uses these tags to identify the correct RDS clusters in order to install new databases.
 * Tag:Counter is maintained by the provisioner. The value is incremented and decremented on every new installation or uninstallation respectively.
 * There is a hardcoded limit in the number database installations that each RDS cluster can take. This limit is referenced by a constant also defined in `tools/aws/constants.go`.
 * The multitenant database follows the same design pattern of all other databases. In fact, it implements the same interface.
 * Small modifications were done to the signature of some the interface's methods . 
```


gsagula:
@stylianosrigas @gabrieljackson This is most of the work but would be great if I can have your feedback in logic, tags, etc. If you want to try it in dev, I've already created one DB cluster per VPC. You just need to create a cluster, run the installation command with: `--database aws-multitenant-rds --affinity multitenan`:

For example:
```
$ cloud installation create --database aws-multitenant-rds --dns hoarders-ca.dev.cloud.mattermost.com --owner gsagula --affinity multitenant
```

Please, let me know if you have any question. Thanks!

gsagula:
@gabrieljackson Hold on the review here. I have to push a considerable amount of changes for this PR so I it can handle @stylianosrigas work on the new instances. I let you know when it's ready. Thanks!

gsagula:
I will  fix and add more tests after the first review.

gabrieljackson:
Hmm, looks like we have failing tests still?

internal/tools/utils/utils.go
```diff
@@ -113,6 +113,8 @@ func (r *ResourceUtil) GetDatabase(installation *model.Installation) model.Datab
 		return model.NewMysqlOperatorDatabase()
 	case model.InstallationDatabaseAwsRDS:
```
gigawhitlocks, line 114: 1/5 maybe we should change this old type to `InstallationDatabaseSingleTenantRDS` for consistency with the new type?

gsagula:
> Hmm, looks like we have failing tests still?

@gabrieljackson I thought I commented about the tests situation. I just need a quick pass/feedback. I will fix the tests anyway and I let yo know when it's done.

model/database_cluster.go
```diff
@@ -0,0 +1,82 @@
+package model
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// DatabaseClusterInstallations is a container that holds a collection of installation IDs.
+type DatabaseClusterInstallations []string
```
gigawhitlocks, line 10: 1/5 on this whole thought; take it or leave it :)

I'm not sure that this type is needed. The methods defined on it are basic CRUD (all but `Remove` have implementations that are very simple to understand) and don't do anything DatabaseClusterInstallation specific, and in addition, `DatabaseClusterInstallation` isn't a type so `DatabaseClusterInstallations` still has no guarantees -- nor does it make any itself -- about the type of string held by `DatabaseClusterInstallations`. 

So for instance, if I had some `databaseClusterInstallations []string` and I wanted to operate on that, I would find `len(databaseClusterInstallations)` in my calling routine equally clear or clearer than if I had some `databaseClusterInstallations *DatabaseClusterInstallation` and I performed `databaseClusterInstallations.Size()` except that I have to look up `Size` if I'm not familiar with this type, whereas `len()` is standard so I know that one off the top of my head.

If we're preparing to add methods to this type that take advantage of unique properties of the IDs held within this list, keeping this type makes sense to me, but otherwise it seems like unnecessary indirection with little benefit.

gigawhitlocks: I feel a bit more strongly that if you keep this type, though, it should be renamed to `DatabaseClusterInstallationIDs` because it does not hold pointers to `DatabaseClusterInstallation` objects, it holds a list of IDs, and so the name is misleading as-is.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+	databaseClusters, err := store.GetDatabaseClusters()
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "could not get installation id %", d.installationID)
+	}
+	if len(databaseClusters) < 1 {
+		return nil, nil, errors.Errorf("could not find installation id %", d.installationID)
+	}
+
+	var rdsCluster *rds.DBCluster
+	for _, databaseCluster := range databaseClusters {
```
gigawhitlocks, line 138: Either moving this for loop to a helper function or adding a comment saying what it does would aid readability here 0/5

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+
+			break
+		}
+	}
+
+	if rdsCluster == nil {
+		return nil, nil, errors.Errorf("could not find an RDS endpoint for installation id %s", d.installationID)
+	}
+
+	installationSecreteName := RDSMultitenantSecretName(d.installationID)
```
gigawhitlocks, line 164: This variable name has a typo in it; you have the suffix `SecreteName` and want `SecretName`. `gorename` should make short work of this.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+	"strconv"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/rds"
+	gt "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/mattermost/mattermost-cloud/model"
```
gigawhitlocks, line 17: I believe that idiomatically we prefer to organize the imports so that Mattermost imports are separate from the 3rd party libraries.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get database cluster for RDS cluster ID %s", *lockedCluster.cluster.DBClusterIdentifier)
+	}
+
+	databaseClusterInstallations, err := databaseCluster.GetInstallations()
+	if err != nil {
+		return errors.Wrapf(err, "failed to get database cluster installations for RDS cluster ID %s", *lockedCluster.cluster.DBClusterIdentifier)
+	}
+
+	if !databaseClusterInstallations.Contains(d.installationID) {
```
gabrieljackson, line 294: I am a little confused by this. In the beginning of this method we determine that a databaseCluster has this Installation in it as that is how the RDS cluster is determined. There really should be no way for this if statement to be true, right?

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+	lockedCluster, err := d.lockRDSCluster(*vpc.VpcId, store, logger)
+	if err != nil {
+		return errors.Wrapf(err, "unable to lock a RDS cluster for installation ID %s", d.installationID)
+	}
+	defer lockedCluster.unlock(logger)
+
+	masterSecretValue, err := d.client.Service().secretsManager.GetSecretValue(&secretsmanager.GetSecretValueInput{
+		SecretId: lockedCluster.cluster.DBClusterIdentifier,
+	})
+	if err != nil && !IsErrorCode(err, secretsmanager.ErrCodeResourceNotFoundException) {
```
gigawhitlocks, line 219: I think this error check might not be right. The check says if err is not a resource not found exception.. then log "unable to find ..."

I'm a bit confused. If we're unable to find the secret, wouldn't we get a ResourceNotFoundException? What is the error case where this `if` evaluates to `true`?

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+		return errors.Wrapf(err, "unable to find the master secret for RDS cluster ID %s", *lockedCluster.cluster.DBClusterIdentifier)
+	}
+
+	close, err := d.connectRDSCluster(rdsMySQLSchemaInformationDatabase, *lockedCluster.cluster.Endpoint, DefaultMattermostDatabaseUsername, *masterSecretValue.SecretString)
+	defer close(logger)
+	if err != nil {
+		return errors.Wrapf(err, "unable to connect to RDS cluster ID %s", *lockedCluster.cluster.DBClusterIdentifier)
+	}
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(DefaultMySQLDefaultContextTimeout*time.Second))
```
gigawhitlocks, line 229: 2 things:
1) `DefaultMySQLDefaultContextTime` has `Default` in it twice and I'm not sure that was intentional.
2) Should `DefaultMySQLDefaultContextTime` be a `time` type instead of an `int` so that the unit of time measured in the actual default is unambiguous and the consumer of that object doesn't need to know that it's in seconds?

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+			{
+				Key:   aws.String(trimTagPrefix(DefaultMattermostInstallationIDTagKey)),
+				Value: aws.String(d.installationID),
+			},
+		}
+		installationSecret, err = d.createInstallationSecret(installationSecreteName, d.installationID, description, tags)
+		if err != nil {
+			return errors.Wrapf(err, "failed to create a RDS database secret for installation ID %s", d.installationID)
+		}
+	}
```
gigawhitlocks, line 272: This section is probably another good candidate for a helper method. A comment at the top of the stanza would be helpful at the least.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+
+type lockRDSClusterOutput struct {
+	unlock  func(log.FieldLogger)
+	cluster *rds.DBCluster
+}
+
+func (d *RDSMultitenantDatabase) lockRDSCluster(vpcID string, store model.InstallationDatabaseStoreInterface, logger log.FieldLogger) (*lockRDSClusterOutput, error) {
+	clusters, err := store.GetDatabaseClusters()
+	if err != nil {
+		return nil, err
```
gigawhitlocks, line 328: Probably best to `Wrap` this `err`

gsagula:
Thanks @gabrieljackson and @gigawhitlocks for the feedback. I will address your comments and work on the tests.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+					if err != nil {
+						return nil, err
+					}
+
+					if counter < DefaultRDSMultitenantDatabaseCountLimit {
+						cluster := model.DatabaseCluster{
+							ID: *rdsClusterID,
+						}
+
+						err := store.CreateDatabaseCluster(&cluster)
```
gigawhitlocks, line 378: I'm a little bit confused as to why a helper method called `lockRDSCluster` has an execution branch that may create a DatabaseCluster in the local database. I think adding some comments and breaking `lockRDSCluster` into several functions (it is quite long) would aid the clarity of this helper quite a bit.

It seems like a local representation of the RDSCluster is created if one is found in AWS but it isn't in the local database. Then the lock is stored with the new RDSCluster object in the local database. Therefore, if a new Provisioner database is used, any existing infrastructure will be considered unlocked. Is this a correct interpretation of this implementation?

gabrieljackson:
Oh one more quick thing. Before we cement in the database too much, I think we may want to try to avoid using the word `cluster` again in `DatabaseCluster` and the various methods, variables, etc. where it makes sense.

Between `Cluster`, `ClusterInstallation`, etc. it would be nice if we could go with something like `Database`, `MultitenantDatabase`, etc.

Thoughts @jwilander, @gigawhitlocks ?

jwilander:
Oh yeah agreed. I think we should only use `Cluster` in the way we have been. It's already been getting confusing when we're talking about the DB clusters whether someone means the RDS cluster or the K8s cluster. What about just `Database` for the table name?

gabrieljackson:
I think `Database` sounds best to me.

This also allows us to do things in the future like add new columns to the table like `Types` which could be `RDS` or `RDS Multitenant`.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+		return nil, errors.Wrapf(err, "failed to get RDS cluster id %s", dbClusterID)
+	}
+	if len(dbClusterOutput.DBClusters) != 1 {
+		return nil, fmt.Errorf("expected exactly one RDS cluster for installation id %s (found %d)", d.installationID, len(dbClusterOutput.DBClusters))
+	}
+
+	return dbClusterOutput.DBClusters[0], nil
+}
+
+func (d *RDSMultitenantDatabase) acquireLock(dbClusterID string, store model.InstallationDatabaseStoreInterface) (func(logger log.FieldLogger), error) {
```
gigawhitlocks, line 545: The naming here is a little confusing. A comment would help enlighten me as to the intended difference between this and `lockRDSCluster` but a more specific name for one or both of these methods would improve this. As it is, `myRDSMultitenantDatabase.lockRDSCluster()` and `myRDSMultitenantDatabase.acquireLock()` sound like they should do the same thing.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+		if err != nil {
+			logger.WithError(err).Errorf("failed to close database connection to %s", endpoint)
+		}
+	}
+
+	return close, nil
+}
+
+func (d *RDSMultitenantDatabase) createUserIfNotExist(ctx context.Context, username, password string) error {
+	query := fmt.Sprintf("CREATE USER IF NOT EXISTS '%s'@'%s' IDENTIFIED BY '%s'", username, "%", password)
```
gigawhitlocks, line 607: This type of query construction is vulnerable to SQL injection. Instead of `Sprintf` and `%s`, you need to use `?` for string interpolation in conjunction with the SQL library like this: 

https://github.com/mattermost/mattermost-cloud/blob/master/internal/store/installation.go#L64

gsagula:
> Oh yeah agreed. I think we should only use `Cluster` in the way we have been. It's already been getting confusing when we're talking about the DB clusters whether someone means the RDS cluster or the K8s cluster. What about just `Database` for the table name?

I like `Database` too. I've been calling it DB cluster because it's how AWS calls it but it can cause confusion for sure.

internal/tools/aws/database_multitenant.go
gsagula: I know, but we can't use the placeholder for "CREATE", apparently. I tried and it errors. It seems that there is a bug somewhere in  the sql library. I have to find it.

internal/tools/aws/database_multitenant.go
gsagula: You can't create constants of complex types in Go, unfortunately. The double Default is not intentional.

gigawhitlocks: Does it need to be an actual `const`? I'd rather have a `var` of the correct type personally, but I guess that's personal preference. If you don't like that, perhaps include the unit as part of the name, e.g. `DefaultMySQLContextTimeSeconds`?

gsagula: The constant unit useful if we want to fine tune it in the future. We don't need to change every single call site.

gsagula: I can change it to `DefaultMySQLContextTimeSeconds`.

internal/tools/aws/database_multitenant.go
gsagula: Nice catch, this is a bug 👍

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+
+	_, err := d.db.ExecContext(ctx, query)
+	if err != nil {
+		return errors.Wrapf(err, "creating user database user %s", username)
```
gigawhitlocks, line 611: I think the first instance of `user` in this error message was unintentional. Probably should read `"creating database user %s"`

internal/tools/aws/database_multitenant.go
gsagula: That is the correct interpretation. It's a relatively complex logic and I agree that it should be broken up in smaller helpers for clarity. My plan was to go for correctness first, get a quick review pass, add tones of testing and then make the code cleaner. However, it's great that you are pointing out all the places where it should happen so I know by a fact that it's difficult to read.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+
+	databaseClusters, err := store.GetDatabaseClusters()
+	if err != nil {
+		return errors.Wrap(err, "unable to fetch database clusters from data store")
+	}
+	if len(databaseClusters) < 1 {
+		return errors.Errorf("data store has no database clusters")
+	}
+
+	for _, databaseCluster := range databaseClusters {
```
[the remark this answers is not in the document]
gsagula: That's a great suggestion! Thanks!

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+
+			rdsCluster, err := d.describeRDSCluster(databaseCluster.ID)
+			if err != nil {
+				return errors.Wrap(err, "could not describe RDS cluster")
+			}
+
+			err = d.updateCounterTag(rdsCluster.DBClusterArn, dbInstallations.Size())
+			if err != nil {
+				return errors.Wrapf(err, "failed to update counter tag in RDS cluster id %s", databaseCluster.ID)
+			}
```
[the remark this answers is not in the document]
gsagula: Indeed! I just realized that it's missing a very important part which checks the number of databases in the actual cluster and then updates the tag not already updated. Nice catch, thanks!

internal/tools/aws/database_multitenant.go
gigawhitlocks: I did some investigation on this. When you connect to MySQL you have to add a parameter to allow parameter interpolation and then you can use parameters. The following is equivalent to your `createUserIfNotExists` using `database/sql` and the MySQL driver, except I'm using `Exec` instead of `ExecWithContext` just to simplify my example.

```go
db, err := sql.Open("mysql", "user:my-secret-pw@/mysql?interpolateParams=true")
q := "CREATE USER IF NOT EXISTS ?@? IDENTIFIED BY ?"
result, err := db.Exec(q, "ian", "%", "mypassword")
```

I ran a local MySQL Docker container and was able to execute the above successfully with the Go REPL `gore`.

gsagula: This is the error that I'm seeing when using the placeholder `Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?@? IDENTIFIED BY ?' at line 1`. However, `interpolateParams=true` did the trick. Thanks!

gsagula: Ok, too early. Can you try `result, err := db.Exec("CREATE DATABASE IF NOT EXISTS ? CHARACTER SET ?", databaseName, "utf8mb4")`?

gsagula: @gigawhitlocks It turns out that **this is a bug or at very least a limitation** of the library in place. The issue here is that when using the placeholders the library *always input single quote* which results in an invalid [SQL statement](https://dev.mysql.com/doc/refman/5.7/en/create-database.html).

db.Exec("CREATE DATABASE IF NOT EXISTS ?", "my_database", "utf8mb4"):
```
Invalid:

CREATE DATABASE IF NOT EXISTS 'my_database';
```

```
valid:

CREATE DATABASE IF NOT EXISTS my_database;
```

> I ran a local MySQL Docker container and was able to execute the above successfully with the Go REPL gore

I'm not sure if you tried the other two commands but they result in the same error:
```
 Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''cloud_ubzzobtfginpdq773nw3chr1cw'' at line 1
```

> TL;DR func RDSMySQLConnString in helpers.go should be modified to add &interpolateParams=true to the returned value and then you can change these three queries to use SQL-injection resistant interpolation.

I agree that we should always write code without vulnerabilities. However, injection is normally a concern for applications which take human inputs. In the case of this application, database, user creating and permissions is exclusively being done by the provisioner. Besides  to that, I don't think that (CREATE USER, CREATE DATABASE, and GRANT) are susceptible to [SQL-injection](https://owasp.org/www-community/attacks/SQL_Injection).

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,637 @@
...
+func NewRDSMultitenantDatabase(installationID string, client *Client) *RDSMultitenantDatabase {
+	return &RDSMultitenantDatabase{
+		client:         client,
+		installationID: installationID,
+	}
+}
+
+// Teardown removes all AWS resources related to a multitenant RDS database.
+func (d *RDSMultitenantDatabase) Teardown(store model.InstallationDatabaseStoreInterface, keepData bool, logger log.FieldLogger) error {
+	logger.Info("Tearing down multitenant RDS database cluster")
```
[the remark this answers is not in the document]
gsagula: Based on our conversation and the new requirement, we should drop the database if keep-data flag is false. I changed this to `Tearing down RDS database and database secret`

internal/tools/aws/database_multitenant.go
gsagula: At the beginning of this method we create a database name, get the claimed VPC, acquire a lock for some RDS cluster if any, and then create the database. The condition here is just in case of the update counter fails we don't try to update the store again.

gsagula: It got resolved by creating specific datastore methods for adding and removing installations.

gsagula:
I'm still writing the tests but this should be good for another pass if you want to take a look at.

jwilander:
@gigawhitlocks @gabrieljackson friendly bump to take a look

internal/tools/aws/database.go
```diff
@@ -47,7 +47,7 @@ func (d *RDSDatabase) Provision(store model.InstallationDatabaseStoreInterface,
 }
 
 // Teardown removes all AWS resources related to a RDS database.
-func (d *RDSDatabase) Teardown(keepData bool, logger log.FieldLogger) error {
+func (d *RDSDatabase) Teardown(store model.InstallationDatabaseStoreInterface, keepData bool, logger log.FieldLogger) error {
```
gigawhitlocks, line 50: 0/5 nitpick: is the name `InstallationDatabaseStoreInterface` redundant? Maybe just `InstallationDatabaseInterface` is better? The `Store` seems extraneous to me, or maybe even confusing because we refer to s3/minio as the "filestore".

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,717 @@
...
+// Helpers
+
+type validatedAndLockedRDSCluster struct {
+	unlock  func(log.FieldLogger)
+	cluster *rds.DBCluster
+}
+
+func (d *RDSMultitenantDatabase) validateAndLockRDSCluster(multitenantDatabases []*model.MultitenantDatabase, store model.InstallationDatabaseStoreInterface, logger log.FieldLogger) (*validatedAndLockedRDSCluster, error) {
+	for _, multitenantDatabase := range multitenantDatabases {
+		installations, err := multitenantDatabase.GetInstallationIDs()
```
gigawhitlocks, line 242: It would be really nice if the body of this `for` loop was broken into a helper method

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,717 @@
...
+				unlockFn(logger)
+				logger.WithError(err).Errorf("Multitenant database ID %s validation failed", multitenantDatabase.ID)
+				continue
+			}
+
+			rdsCluster, err := d.describeRDSCluster(multitenantDatabase.ID)
+			if err != nil {
+				unlockFn(logger)
+				logger.WithError(err).Errorf("Failed to describe the multitenant RDS cluster ID %s", multitenantDatabase.ID)
+				continue
```
gigawhitlocks, line 271: specifically the above mentioned helper method would let you turn these `continue` statements into `return`s, I think, which might make it easier to think about this routine.

internal/tools/aws/database.go
gsagula: I agree but given that this an old interface code, I'm not refactoring it in this PR. I will put a comment for followup changes.

internal/tools/aws/database_multitenant.go
gsagula: Is this for testing reasons, readability, or something else?

internal/tools/aws/database_multitenant.go
gsagula: I see, disregard the question above 👍

internal/tools/aws/database_multitenant.go
gsagula: Ok, I see the reason in the comment below. I will do the change.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -34,10 +34,14 @@ type KopsProvisioner struct {
 	useExistingAWSResources bool
 	resourceUtil            *utils.ResourceUtil
 	logger                  log.FieldLogger
+	store                   model.InstallationDatabaseStoreInterface
 }
 
 // NewKopsProvisioner creates a new KopsProvisioner.
-func NewKopsProvisioner(s3StateStore, owner string, useExistingAWSResources bool, resourceUtil *utils.ResourceUtil, logger log.FieldLogger) *KopsProvisioner {
+// TODO(gsagula): Consider place all this paramaters in a struct for readability.
```
gigawhitlocks, line 41: good TODO :+1:

internal/tools/aws/database.go
gigawhitlocks: Oh sorry, I thought it was a new interface. Definitely a comment is enough, or just ignore my feedback LOL

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,730 @@
...
+	databaseName := MattermostRDSDatabaseName(d.installationID)
+
+	logger = logger.WithField("rds-multitenant-database", databaseName)
+	logger.Info("Tearing down RDS database and database secret")
+
+	multitenantDatabases, err := store.GetMultitenantDatabases(&model.MultitenantDatabaseFilter{
+		InstallationID: d.installationID,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "cannot teardown multitenant database for installation ID %s", d.installationID)
```
gabrieljackson, line 64:
```suggestion
		return errors.Wrapf(err, "unable to find multitenant database for installation ID %s", d.installationID)
```

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,730 @@
...
+	}
+
+	multitenantDatabaseSecretName := RDSMultitenantSecretName(d.installationID)
+
+	_, err = d.client.Service().secretsManager.DeleteSecret(&secretsmanager.DeleteSecretInput{
+		SecretId: aws.String(multitenantDatabaseSecretName),
+	})
+	if err != nil && !IsErrorCode(err, secretsmanager.ErrCodeResourceNotFoundException) {
+		return errors.Wrapf(err, "failed to delete multitenant database secret name %s", multitenantDatabaseSecretName)
+	}
```
gabrieljackson, line 87: All of this happens after the `MultitenantStore` has been updated to remove the installation ID. Due to this, there is the possibility to leak the following secret. The store should be the final thing to be updated.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,730 @@
...
+}
+
+// GenerateDatabaseSpecAndSecret creates the k8s database spec and secret for accessing a single database inside
+// a RDS multitenant cluster.
+func (d *RDSMultitenantDatabase) GenerateDatabaseSpecAndSecret(store model.InstallationDatabaseStoreInterface, logger log.FieldLogger) (*mmv1alpha1.Database, *corev1.Secret, error) {
+	installationDatabaseName := MattermostRDSDatabaseName(d.installationID)
+
+	logger = logger.WithField("rds-multitenant-database", installationDatabaseName)
+
+	multitenantDatabases, err := store.GetMultitenantDatabases(&model.MultitenantDatabaseFilter{
```
gabrieljackson, line 106: If you create a method `GetMultitenantDatabaseForInstallationID` you can move all the logic to check that there is one, and only one, returned field to that method. 0/5 on needing to do it, but just an idea.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,730 @@
...
+
+	databaseInstallationIDs, err := store.RemoveMultitenantDatabaseInstallationID(rdsDatabaseCluster.ID, d.installationID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to remove installation ID %s from datasore", d.installationID)
+	}
+
+	err = d.updateCounterTag(rdsCluster.DBClusterArn, len(databaseInstallationIDs))
+	if err != nil {
+		return errors.Wrapf(err, "failed to update counter tag for the multitenant RDS cluster ID %s", rdsDatabaseCluster.ID)
+	}
```
gabrieljackson, line 617: This happens after the MultitenantStore has been updated to remove the installation ID. Due to this, there is the possibility to not update this tag. It's not a huge deal in this case, but the store should be the final thing to be updated.

internal/provisioner/kops_provisioner_cluster.go
gsagula: 👍 but there is a typo `Consider place all this` / `Consider place all these`. 
I will fix it!

internal/store/migrations.go
```diff
@@ -893,6 +893,35 @@ var migrations = []migration{
...
+				);
+			`)
+			if err != nil {
+				return err
+			}
+		} else {
+			_, err := e.Exec(`
+				CREATE TABLE MultitenantDatabase (
+					ID TEXT PRIMARY KEY,
+					RawInstallationIDs byte[] NOT NULL,
```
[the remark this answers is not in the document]
gsagula: Oh, I'm not enough familiar with this. Did not know that they where interchangeable. I will change it to BYTEA.

model/multitenant_database.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return installationIDs, nil
+}
+
+// MultitenantDatabaseFilter filters results based on a specific installation ID and a number of
+// installation's limit.
+type MultitenantDatabaseFilter struct {
+	InstallationID          string
+	NumOfInstallationsLimit uint32
```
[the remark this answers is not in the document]
gsagula: Sounds good, I will do the change 👍

internal/store/multitenant_database.go
```diff
@@ -0,0 +1,184 @@
...
+	if err == sql.ErrNoRows {
+		return nil, nil
+	} else if err != nil {
+		return nil, errors.Wrap(err, "failed to get multitenant database by id")
+	}
+
+	return &multitenantDatabase, nil
+}
+
+// GetMultitenantDatabases fetches the given page of created database clusters. The first page is 0.
```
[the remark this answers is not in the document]
gsagula: I was planning on add pages but I didn't see a real reason for what I need from this api.

internal/store/multitenant_database.go
```diff
@@ -0,0 +1,184 @@
...
+			installationIDs, err := database.GetInstallationIDs()
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to query multitenant databases")
+			}
+
+			if len(installationIDs) > int(filter.NumOfInstallationsLimit) {
+				multitenantDatabases = append(multitenantDatabases[:i], multitenantDatabases[i+1:]...)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
gsagula: I I understand your concern correctly, I thought paging would affect the number of database records. This code is restricting the number of installations per  mutitenant database:
```
len(installationIDs) > int(filter.NumOfInstallationsLimit)
```

model/installation_database.go
```diff
@@ -10,22 +10,32 @@ import (
 const (
 	// InstallationDatabaseMysqlOperator is a database hosted in kubernetes via the operator.
 	InstallationDatabaseMysqlOperator = "mysql-operator"
-	// InstallationDatabaseAwsRDS is a database hosted via Amazon RDS.
-	InstallationDatabaseAwsRDS = "aws-rds"
+	// InstallationDatabaseSingleTenantRDS is a database hosted via Amazon RDS.
+	InstallationDatabaseSingleTenantRDS = "aws-singletenant-rds"
```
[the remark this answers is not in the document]
gsagula: This changed was request [here](https://github.com/mattermost/mattermost-cloud/pull/209#discussion_r422214696) and I didn't realize that.

internal/tools/aws/database_multitenant.go
gsagula: Unless the installation ID is gone, I don't see how. Secrets are managed by AWS and we just reference them by a name that depends on the installations ID. It's completely independent of the provisioner's persistent data.

gabrieljackson: Right, so if the installation ID is removed on the first pass on line 80, but then we hit an error here on line 82 it will fail and eventually retry.

When it does, this check above will return from this function and the secret cleanup will be missed.

```
	if len(multitenantDatabases) == 0 {
		logger.Warnf("Installation ID %s does not belong to any of the multitenant databases. Nothing to be torn down.", d.installationID)

		return nil
	}
```

internal/tools/aws/database_multitenant.go
gsagula: The code above always return the installation IDs for the given multitenant database. This is independent of the installation being removed or not. However, if the installation ID exists and this operation is called without an error being returned, it's guaranteed that the ID was removed. This simplifies the things in case the `updateCounterTag()` returns error and needs to retry.

internal/tools/aws/database_multitenant.go
gsagula: Oh right, I forgot that I added this check. Problem of moving it before `removeRDSDatabase()` is that I need the secrets in oder to drop the database. I fix the logic.

gsagula: I'm still writing the tests for the provisioning part. I will make sure that this is being tested.

internal/tools/aws/database_multitenant.go
gabrieljackson: Hmm, I am not sure that I understand.

I am pointing out that if line 611 succeeds, but line 616 fails, then we will never reach this `removeRDSDatabase` method again with the current code and the `updateCounterTag` will never be successfully called.

internal/store/multitenant_database.go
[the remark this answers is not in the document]
gabrieljackson: No, you are correct that paging should only be used to control how many entries are returned and `model.AllPerPage` will return them all.

I see now how this is used. Thanks.

internal/tools/aws/database_multitenant.go
gsagula: Yes, that's a fact. I didn't want move all this logic into a helper for this reason. Let me work on this and add testing around it.

internal/tools/aws/database_multitenant.go
```diff
@@ -0,0 +1,728 @@
...
+}
+
+// GenerateDatabaseSpecAndSecret creates the k8s database spec and secret for accessing a single database inside
+// a RDS multitenant cluster.
+func (d *RDSMultitenantDatabase) GenerateDatabaseSpecAndSecret(store model.InstallationDatabaseStoreInterface, logger log.FieldLogger) (*mmv1alpha1.Database, *corev1.Secret, error) {
+	installationDatabaseName := MattermostRDSDatabaseName(d.installationID)
+
+	logger = logger.WithField("rds-multitenant-database", installationDatabaseName)
+
+	multitenantDatabases, err := store.GetMultitenantDatabases(&model.MultitenantDatabaseFilter{
```
gabrieljackson, line 96: Feel free to use the new method you created on this. `GetMultitenantDatabaseForInstallationID`

jwilander:
@stylianosrigas mind giving this quick look from the infra side?

gsagula:
It's the checking in one of the unit tests that is different for pq. I will fix it.

gsagula:
@jwilander I added as many tests as I could but I will add more in a follow up PR.  Feel free to merge this one. Thanks!

PR #211, opened 2020-05-07
fix missing parameters and default values

cpanato:
#### Summary
add missing default values checker for the utility metadata and fix the wrong Version value

#### Release Note

```release-note
add missing default values for post cluster installation
```

### DIFF
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -81,7 +81,7 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	cluster := model.Cluster{
 		Provider:           createClusterRequest.Provider,
 		Provisioner:        "kops",
-		Version:            "0.0.0",
+		Version:            createClusterRequest.Version,
 		Size:               createClusterRequest.Size,
 		AllowInstallations: createClusterRequest.AllowInstallations,
 		State:              model.ClusterStateCreationRequested,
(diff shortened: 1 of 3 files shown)


internal/api/cluster.go
```diff
@@ -81,7 +81,7 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	cluster := model.Cluster{
 		Provider:           createClusterRequest.Provider,
 		Provisioner:        "kops",
-		Version:            "0.0.0",
+		Version:            createClusterRequest.Version,
```
gabrieljackson, line 84: Hmm, this was actually originally done intentionally to indicate that k8 was not fully set up and verified. This works too, but I sort of prefer the original value. 0/5 though.

cpanato: but then there are differences calling via API or cmd line

if call via API the version will be filled with 0.0.0 and if you call via cmd it will fill the version

gabrieljackson: Hmm, are we sure about that?

The value set here was originally intended to be the version currently running on the k8 cluster. The k8 version in the create cluster request is still used though. It is entered as part of the ProvisionerMetadata: https://github.com/mattermost/mattermost-cloud/pull/211/files#diff-16d6f77671cb631b73f6db649183b5e5R108

The value that is used by kops is the value from the ProvisionerMetadata, so what we set here is just for our own information when we view the cluster object.

cpanato: i will revert this change and after merge this will see. if does not work as expected will push back

cpanato: however this mismatch the behavior we have using the command line and using the API
in the API we cannot set the version in the other hand using the command line we can

PR #212, opened 2020-05-12
update documentation readme

angeloskyratzakos:
#### Summary
Updating documentation with commands for 
* creating/deleting cluster
* creating/deleting installation

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,12 @@ This repository houses the open-source components of Mattermost Private Cloud. O
 
 1. Install [Go](https://golang.org/doc/install)
 2. Install [Terraform](https://learn.hashicorp.com/terraform/getting-started/install.html) version v0.11.14
+   1. Try using [tfswitch](https://warrensbox.github.io/terraform-switcher/) for switching easily between versions
 3. Install [kops](https://github.com/kubernetes/kops/blob/master/docs/install.md) version 1.15.X
 4. Install [Helm](https://helm.sh/docs/using_helm/) version 2.14.X
 5. Install [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-7. Install [mockgen](github.com/golang/mock/mockgen) version 1.4.x
-6. Specify the region in your AWS config, e.g. `~/.aws/config`:
+6. Install [mockgen](github.com/golang/mock/mockgen) version 1.4.x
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,30 +46,122 @@ Simply run the following:
...
+
+- make sure you are running terraform 11.x version
+    ```
+    terraform version
+    ```
+    If not try to use [tfswitch](https://warrensbox.github.io/terraform-switcher/) for switching between versions easily:
+    `brew install warrensbox/tap/tfswitch`
+
+
+- Make sure you have Helm version greater/equal than 2.16.4
```
gabrieljackson, line 74: I am pretty sure we actually have issues with newer versions of helm. That was the case in the past at least. We may just want to change this to be exactly `2.16.4` for now.

README.md
```diff
@@ -46,30 +46,122 @@ Simply run the following:
...
 ### Testing
 
 Run the go tests to test:
 
-```
+```bash
 $ go test ./...
 ```
 
+### Deleting a cluster
```
gabrieljackson, line 155: You may want to call out that you have to delete all installations running on a cluster before the cluster can be deleted.

README.md
```diff
@@ -46,30 +46,122 @@ Simply run the following:
 
 ```
 $ go install ./cmd/cloud
+alias cloud='$HOME/go/bin/cloud'
 ```
 
-### Running
 
+### Pre-requisites
```
gabrieljackson, line 53: This section is nice to have, but it seems like we should try to combine it with the `Environment Setup` section above because they contain the same information. Actually, it looks like there is now a discrepancy on Helm version recommendation with this change so we should definitely update/combine these sections.

README.md
angeloskyratzakos: I will change it to be indifferent from 2.16.4 as I had some issues with nginx with that version, https://stackoverflow.com/questions/60836127/error-validation-failed-serviceaccounts-nginx-ingress-not-found-serviceacc

README.md
angeloskyratzakos: Sorry Gabe I didn't know that one! I will document it

README.md
angeloskyratzakos: you are right Gabe, I have completely missed that. I am updating the setup section.

PR #214, opened 2020-05-13
Remove cert manager and use AWS ACM in public ngix

cpanato:
#### Summary
This PR address two items

1. remove the cert-manager deployment and configuration

Due to a limitation in how many SSL certificates we can generate and having no option for a paid subscription, we decide the remove this feature for now and use the certificate that we generate internally using AWS Certificate Manager.

2. Update the public Nginx deployment to attach the AWS ACM we already have and use that to have a SSL certificate for the installations

#### Ticket Link
 - fixes Jira: https://mattermost.atlassian.net/browse/MM-24954

#### Release Note

```release-note
- remove cert-manager and attach AWS ACM in the public Nginx deployment to use SSL certificates
```


internal/provisioner/public_nginx.go
```diff
@@ -73,11 +81,17 @@ func (n *publicNginx) Destroy() error {
 }
 
 func (n *publicNginx) NewHelmDeployment() *helmDeployment {
+	awsACMCert, err := n.awsClient.GetCertificateSummaryByTag(aws.DefaultCloudDNSTagKey, aws.DefaultPublicCloudDNSTagValue, n.logger)
```
stylianosrigas, line 84: @cpanato how are we planning to create this Cert? I guess with Terraform when we deploy the provisioner in C&C?

helm-charts/public-nginx_values.yaml
```diff
@@ -15,16 +29,24 @@ controller:
...
 
-  config: 
+  config:
     http-snippet: |
       proxy_cache_path /cache/nginx levels=1:2 keys_zone=mattermost_test_cache:10m max_size=10g inactive=120m use_temp_path=off;
       proxy_cache_path /cache/mattermost levels=1:2 keys_zone=mattermost_cache:10m max_size=10g inactive=120m use_temp_path=off;
     keep-alive: "3600"
     proxy-max-temp-file-size: "0"
+    # Will add custom configuration options to Nginx https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/
+    ssl-redirect: "false"
```
stylianosrigas, line 46: @cpanato where are we handling the redirect now?

internal/provisioner/public_nginx.go
cpanato: the certificate is already created
for test is *.test.mattermost.cloud and so on.
we are using the same we used to use before

helm-charts/public-nginx_values.yaml
cpanato: i will make a quick test on that

internal/provisioner/public_nginx.go
stylianosrigas: Athh the one we were using before. We need to make sure we have this everywhere. Cool thanks :)

internal/provisioner/certmanager.go
```diff
@@ -1,164 +0,0 @@
...
-
-func (n *certManager) DesiredVersion() string {
-	return n.desiredVersion
-}
-
-func (n *certManager) ActualVersion() string {
-	return strings.TrimPrefix(n.actualVersion, "cert-manager-")
-}
-
-func (n *certManager) Destroy() error {
```
gabrieljackson, line 76: It would be nice if we performed cleanup of this utility. This is where it normally would happen. The idea here is we could perform a `cluster provision` to cleanup old utilities.

helm-charts/public-nginx_values.yaml
cpanato: tested with those configs and redirect works as well

internal/provisioner/certmanager.go
cpanato: you mean, we clean that and later we remove the cert-manager code? 
I can revert this to make that
i was planning to remove manually since is not many clusters to perform that

gabrieljackson: Yeah, this probably won't be the first time we add and remove a utility, so I think we should get used to having the provisioner cleanup stuff we no longer need. We should be able to run `provision` on the cluster and have the new ACM stuff install first and then remove the cert manager resources.

PR #216, opened 2020-05-18
Add installation scaling

gabrieljackson:
The installation update API endpoint now allows for updating the
installation size value. The cloud server will then update the
Mattermost app values for replica count and resource requests and
limits.

A few notes on installation sizing changes:
- Resource and replica changes are currently targeting the Mattermost
   app pod only. Other pods such as those managed by the MinIO and
   MySQL operator won't be updated with the current Mattermost
   Operator logic. This is intentional as those apps can have replica
   scaling issues.
 - Resizing currently ignores the installation scheduling algorithm.
   There is no good interface to determine if the new installation
   size will safely fit on the cluster. This could, in theory, be done
   when the size request change comes in on the API, but would require
   new scheduling logic. For now, take care when resizing.

https://mattermost.atlassian.net/browse/MM-25094

```release-note
Add installation scaling
```

### DIFF
--- a/internal/api/installation_test.go
+++ b/internal/api/installation_test.go
@@ -525,8 +525,18 @@ func TestUpdateInstallation(t *testing.T) {
 	require.NoError(t, err)
 
 	ensureInstallationMatchesRequest := func(t *testing.T, installation *model.Installation, request *model.PatchInstallationRequest) {
-		require.Equal(t, *request.Version, installation.Version)
-		require.Equal(t, *request.License, installation.License)
+		if request.Version != nil {
+			require.Equal(t, *request.Version, installation.Version)
+		}
+		if request.Version != nil {
+			require.Equal(t, *request.License, installation.License)
+		}
+		if request.Size != nil {
+			require.Equal(t, *request.Size, installation.Size)
+		}
+		if request.MattermostEnv != nil {
+			require.Equal(t, request.MattermostEnv, installation.MattermostEnv)
+		}
 	}
 
 	t.Run("unknown installation", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/api/installation_test.go
```diff
@@ -525,8 +525,18 @@ func TestUpdateInstallation(t *testing.T) {
 	require.NoError(t, err)
 
 	ensureInstallationMatchesRequest := func(t *testing.T, installation *model.Installation, request *model.PatchInstallationRequest) {
-		require.Equal(t, *request.Version, installation.Version)
-		require.Equal(t, *request.License, installation.License)
+		if request.Version != nil {
```
gigawhitlocks, line 528: Test should still pass if `request.X == nil` and `installation.X` does not?

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -194,14 +194,43 @@ func (provisioner *KopsProvisioner) UpdateClusterInstallation(cluster *model.Clu
...
+		logger.Debugf("Cluster installation already on image %s", installation.Image)
+	} else {
+		logger.Debugf("Cluster installation image updated from %s to %s", cr.Spec.Image, installation.Image)
+		cr.Spec.Image = installation.Image
+	}
+
+	if cr.Spec.Size == installation.Size {
+		logger.Debugf("Cluster installation already on size %s", installation.Size)
+	} else {
+		// A few notes on installation sizing changes:
```
gigawhitlocks, line 213: I wish we could surface these notes about the current limitations somewhere more visible than this comment, perhaps in release notes?

internal/api/installation_test.go
gabrieljackson: Yeah, the request logic was moved to pointer patching a while back so nil means there would be no change to the underlying installation. It also means that we can't check that `request.VALUE` == `installation.VALUE` if the first in nil so this check is definitely needed.

The only reason tests were passing up till now is that every subtest was passing in all the patch values that could be nil. My new test with an empty patch payload was how I uncovered that we should be doing the if logic here.

internal/provisioner/kops_provisioner_cluster_installation.go
gabrieljackson: Yeah, totally agree that this needs to be in more places. I often feel like these types of comments can really be helpful especially when trying to work out a fix to something in a pinch. That said, good idea about surfacing it elsewhere.

gabrieljackson:
Thanks for the review @gigawhitlocks!

PR #219, opened 2020-05-21
Add option to change cluster AMI on upgrade

gabrieljackson:
This new functionality allows the cloud provisioner to update the
various instance groups in a cluster to use a new AMI for the nodes.
Upgrades can now handle an update to the kubernetes version, AMI,
or both simultaneously. Passing in an empty value for the AMI will
use the default kops AMI for that kubernetes version.

Tested manually by updating a cluster with 3 master and 2 worker
nodes.

As a follow-up, I may change the wording around `upgrade` as that
can be misleading when only updating the AMI or if we actually want
to downgrade.

https://mattermost.atlassian.net/browse/MM-24641

```release-note
Add option to change cluster AMI on upgrade
```

### DIFF
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -25,7 +25,7 @@ func initCluster(apiRouter *mux.Router, context *Context) {
 	clusterRouter.Handle("", addContext(handleRetryCreateCluster)).Methods("POST")
 	clusterRouter.Handle("", addContext(handleUpdateClusterConfiguration)).Methods("PUT")
 	clusterRouter.Handle("/provision", addContext(handleProvisionCluster)).Methods("POST")
-	clusterRouter.Handle("/kubernetes/{version}", addContext(handleUpgradeKubernetes)).Methods("PUT")
+	clusterRouter.Handle("/kubernetes", addContext(handleUpgradeKubernetes)).Methods("PUT")
 	clusterRouter.Handle("/size/{size}", addContext(handleResizeCluster)).Methods("PUT")
 	clusterRouter.Handle("/utilities", addContext(handleGetAllUtilityMetadata)).Methods("GET")
 	clusterRouter.Handle("", addContext(handleDeleteCluster)).Methods("DELETE")
@@ -327,16 +327,16 @@ func handleUpdateClusterConfiguration(c *Context, w http.ResponseWriter, r *http
 	outputJSON(c, w, cluster)
 }
 
-// handleUpgradeKubernetes responds to PUT /api/cluster/{cluster}/kubernetes/{version},
-// upgrading the cluster to the given Kubernetes version.
+// handleUpgradeKubernetes responds to PUT /api/cluster/{cluster}/kubernetes,
+// upgrading the cluster to the given Kubernetes version and AMI image.
 func handleUpgradeKubernetes(c *Context, w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	clusterID := vars["cluster"]
-	version := vars["version"]
 	c.Logger = c.Logger.WithField("cluster", clusterID)
 
-	if !model.ValidClusterVersion(version) {
-		c.Logger.Warnf("unsupported kubernetes version %s", version)
+	upgradeClusterRequest, err := model.NewUpgradeClusterRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to decode request")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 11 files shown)


gigawhitlocks:
>As a follow-up, I may change the wording around upgrade as that
can be misleading when only updating the AMI or if we actually want
to downgrade.

Throughout the Provisioner? I'm 100% on board with this if you want to change the wording; the wording around upgrade is something that got me a little confused earlier on.

internal/api/cluster.go
```diff
@@ -362,8 +362,9 @@ func handleUpgradeKubernetes(c *Context, w http.ResponseWriter, r *http.Request)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	if kopsMetadata.Version != version {
-		kopsMetadata.Version = version
+	if kopsMetadata.Version != upgradeClusterRequest.Version || kopsMetadata.AMI != upgradeClusterRequest.KopsAMI {
```
gigawhitlocks, line 365: Are we stylistically opposed to putting a newline before or after the `||` in compound boolean statements like this? Personally I think they're a little easier one per line but 0/5 ignore me if you don't like the suggestion :)

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -376,6 +376,55 @@ func (provisioner *KopsProvisioner) UpgradeCluster(cluster *model.Cluster) error
 	}
 	defer kops.Close()
 
+	switch kopsMetadata.Version {
+	case "latest", "":
+		err = kops.UpgradeCluster(kopsMetadata.Name)
+		if err != nil {
+			return err
```
gigawhitlocks, line 383: Make sure to wrap these errors before merge

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -376,6 +376,55 @@ func (provisioner *KopsProvisioner) UpgradeCluster(cluster *model.Cluster) error
...
+		}
+	}
+
+	instanceGroups, err := kops.GetInstanceGroupsJSON(kopsMetadata.Name)
+	if err != nil {
+		return err
+	}
+	for _, ig := range instanceGroups {
+		if ig.Spec.Image != kopsMetadata.AMI {
+			if len(kopsMetadata.AMI) == 0 {
```
gigawhitlocks, line 399: This if statement & everything inside would make a good helper function, which would in turn make this parent function a bit easier to read.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -399,25 +448,6 @@ func (provisioner *KopsProvisioner) UpgradeCluster(cluster *model.Cluster) error
 
 	logger.Info("Upgrading cluster")
```
gigawhitlocks, line 449: Doesn't this log message need to move since the stanza below it moved? Otherwise, unless I'm misunderstanding the code, the Provisioner will print "Upgrading cluster" after kops runs but before Terraform, which seems a little weird to me.

internal/provisioner/kops_utils.go
```diff
@@ -193,3 +193,24 @@ func grossKopsReplaceSize(input, machineType, min, max string) (string, error) {
...
+// kops instance group YAML manifest with a new image value.
+// TODO: remove once new `kops set instancegroup` functionality is available.
+//
+// Example Manifest:
+//
+// apiVersion: kops.k8s.io/v1alpha2
+// kind: InstanceGroup
+// spec:
+//   image: kope.io/k8s-1.15-debian-stretch-amd64-hvm-ebs-2020-01-17
+func grossKopsReplaceImage(input, image string) (string, error) {
```
gigawhitlocks, line 207: well, at least you named it `gross` LOL

internal/provisioner/kops_provisioner_cluster.go
gabrieljackson: It definitely does seem like that may be wrong, but I just double-checked and I think that is still pretty much the correct spot for it. The reasoning is that everything above that line is basically trying to get the kops manifests in the right state for what we want to "upgrade" to and then starting with the terraform plan+deploy we are making actual resource changes that affect the running cluster.

internal/provisioner/kops_provisioner_cluster.go
gabrieljackson: Will do

internal/api/cluster.go
gabrieljackson: This actually reminded me that I had originally intended to use patching logic and have the `Apply` method take care of reporting if the metadata was updated. New commit has this logic!

PR #222, opened 2020-05-25
Add installation and group env var patching logic

gabrieljackson:
This change replaces the original env var update logic from
complete-replacement behavior to flexible patching behavior. The
new logic is as follows:
 - New EnvVarMap is nil: empty patch so no updates.
 - New EnvVarMap is empty: clear existing map completely.
 - New EnvVarMap has new keys and values: add new keys and values
   to existing map.
 - New EnvVarMap has new values for old keys: update old keys with
   new values.
 - NewEnvVarMap has empty values for old keys: delete the old keys.

https://mattermost.atlassian.net/browse/MM-25338

```release-note
Add installation and group env var patching logic
```

### DIFF
--- a/model/env.go
+++ b/model/env.go
@@ -25,6 +25,11 @@ func (e *EnvVar) Validate() error {
 	return nil
 }
 
+// HasValue returns wheather the env var has any value configuration or not.
+func (e *EnvVar) HasValue() bool {
+	return len(e.Value) != 0 || e.ValueFrom != nil
+}
+
 // Validate returns wheather the env var map has valid value configuration.
 func (em *EnvVarMap) Validate() error {
 	for name, env := range *em {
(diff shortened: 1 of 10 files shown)


model/env.go
```diff
@@ -37,6 +42,64 @@ func (em *EnvVarMap) Validate() error {
...
+func (em *EnvVarMap) ClearOrPatch(new *EnvVarMap) bool {
+	if len(*new) == 0 {
+		orginalEmpty := len(*em) != 0
+		*em = nil
+
+		if orginalEmpty {
+			return true
+		}
+
+		return false
```
gigawhitlocks, line 58: `return originalEmpty` ?

model/env.go
```diff
@@ -37,6 +42,64 @@ func (em *EnvVarMap) Validate() error {
 	return nil
 }
 
+// ClearOrPatch takes a new EnvVarMap and patches changes into the existing
+// EnvVarMap with the following logic:
+//  - If the new EnvVarMap is empty, clear the existing EnvVarMap completely.
+//  - If the new EnvVarMap is not empty, apply normal patch logic.
+func (em *EnvVarMap) ClearOrPatch(new *EnvVarMap) bool {
```
gigawhitlocks, line 49: since the input is a pointer, does it need to be checked that it's not `nil` before we try to dereference it on line 50?
```
gore> %                                                                                             ~ » gore
gore version 0.5.0  :help for help
gore> type MyMap map[string]string
gore> var m *MyMap
gore> m == nil
true
gore> len(*m)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4cd87f]

goroutine 1 [running]:
main.main()
+0x1f
exit status 2
gore> 
```

model/env_test.go
```diff
@@ -59,6 +59,211 @@ func TestEnvValid(t *testing.T) {
...
+			model.EnvVarMap{
+				"key1": {Value: "value1"},
+			},
+			model.EnvVarMap{
+				"key1": {},
+			},
+			model.EnvVarMap{},
+		},
+		{
+			"delete nonexistant keys",
```
gigawhitlocks, line 218: nitpick "nonexistent" is the correct spelling

model/group_request.go
```diff
@@ -111,8 +95,14 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 		group.MaxRolling = *p.MaxRolling
 	}
 	if p.MattermostEnv != nil {
-		applied = true
-		group.MattermostEnv = p.MattermostEnv
+		if group.MattermostEnv != nil {
+			if group.MattermostEnv.ClearOrPatch(&p.MattermostEnv) {
```
gigawhitlocks, line 99: 0/5 could remove the conditional and just say `applied = group.MattermostEnv...`

model/env.go
gabrieljackson: That's just crazy enough to work!

model/group_request.go
gabrieljackson: Oh, just remembered that I can't actually do that as applied could be `true` for the other patches and then this would set it back to `false` incorrectly.

model/env.go
gabrieljackson: Thanks for the input. I wasn't originally checking for nil there because I knew the caller would be checking, but by handling the nil check in this method I can clean up the code in the caller so looks like a win/win to me!

cpanato:
how this works if call using the API?

gabrieljackson:
Hey @cpanato, the payload to the API works like I described in the PR description. I modified my local client to show the API output it is sending:

```
# ADD OR UPDATE AN ENV VAR
~/g/s/g/m/mattermost-cloud ❯❯❯ cloud installation update --installation=id1 --mattermost-env=key1=value1
{
    "Version": null,
    "Image": null,
    "Size": null,
    "License": null,
    "MattermostEnv": {
        "key1": {
            "value": "value1"
        }
    }
}

# CLEAR/DELETE AN ENV VAR
~/g/s/g/m/mattermost-cloud ❯❯❯ cloud installation update --installation=id1 --mattermost-env=key1=                                                                                                 env-patching ✭ ✱
{
    "Version": null,
    "Image": null,
    "Size": null,
    "License": null,
    "MattermostEnv": {
        "key1": {}
    }
}

# CLEAR ALL ENV VARS
~/g/s/g/m/mattermost-cloud ❯❯❯ cloud installation update --installation=id1 --mattermost-env-clear
{
    "Version": null,
    "Image": null,
    "Size": null,
    "License": null,
    "MattermostEnv": {}
}

# DON'T UPDATE ENV VARS AT ALL
~/g/s/g/m/mattermost-cloud ❯❯❯ cloud installation update --installation=id1                                                                                                                        
{
    "Version": null,
    "Image": null,
    "Size": null,
    "License": null,
    "MattermostEnv": null
}
```

PR #223, opened 2020-05-26
MM-25556: multi tenant RDS databases should be provisioned to the correct VPC in a multi cluster environment

gsagula:
#### Summary
This PR fixes the issue where the provisioner tries to install a multitenant database in a RDS cluster that belongs to some other VPC.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-25556

#### Release Note
```release-note
- Fixed the issue where the provisioner tries to install a multitenant database in a RDS cluster that does not belong to the acquired VPC.
- Fixed issue where the provisioner may get stuck trying to teardown an installation id that never had a correspondent multitenant database.
```

### DIFF
--- a/internal/store/multitenant_database.go
+++ b/internal/store/multitenant_database.go
@@ -143,7 +143,6 @@ func (sqlStore *SQLStore) UpdateMultitenantDatabase(multitenantDatabase *model.M
 			"RawInstallationIDs": multitenantDatabase.RawInstallationIDs,
 		}).
 		Where(sq.Eq{"ID": multitenantDatabase.ID}).
-		Where(sq.Eq{"VpcID": multitenantDatabase.VpcID}).
 		Where(sq.Eq{"LockAcquiredBy": *multitenantDatabase.LockAcquiredBy}),
 	)
 	if err != nil {
--- a/internal/store/multitenant_database.go
+++ b/internal/store/multitenant_database.go
@@ -14,7 +14,7 @@ var multitenantDatabaseSelect sq.SelectBuilder
 
 func init() {
 	multitenantDatabaseSelect = sq.
-		Select("ID", "RawInstallationIDs", "CreateAt", "DeleteAt", "LockAcquiredBy", "LockAcquiredAt").
+		Select("ID", "VpcID", "RawInstallationIDs", "CreateAt", "DeleteAt", "LockAcquiredBy", "LockAcquiredAt").
 		From("MultitenantDatabase")
 }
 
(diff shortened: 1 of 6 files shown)


internal/store/multitenant_database.go
```diff
@@ -137,6 +143,7 @@ func (sqlStore *SQLStore) UpdateMultitenantDatabase(multitenantDatabase *model.M
 			"RawInstallationIDs": multitenantDatabase.RawInstallationIDs,
 		}).
 		Where(sq.Eq{"ID": multitenantDatabase.ID}).
+		Where(sq.Eq{"VpcID": multitenantDatabase.VpcID}).
```
gabrieljackson, line 146: We don't really need this, right? I actually don't think we need any `Where`s for inserting a new row. They must be ignored.

gsagula: This is not inserting a new row, it's updating existing one. I think the VPC ID is redundant but  LockAcquiredBy I think is important redundancy to have given that we use this table primarily for locking things. We want to make sure that only locked resources get updated. ID is definitely needed.

gsagula:
> No blockers, but one question: this code won't be able to resolve issues with existing multitenant database entries right? I am guessing that some manual updates and cleanup will need to be done for the test account?

Unfortunately, this code won't fix the test accounts. This will need to be a manual change for sure. How many clusters do we have in test? Or how many installations have been created?

gsagula:
I added a logic to only create a multi-tenant database row if the RDS cluster status is available. IRCC @stylianosrigas raised some concerns regarding adding database clusters in creating state. I still have to fix the tests and add another IAM permission for describing RDS endpoints because of this change.

stylianosrigas:
> I added a logic to only create a multi-tenant database row if the RDS cluster status is available. IRCC @stylianosrigas raised some concerns regarding adding database clusters in creating state. I still have to fix the tests and add another IAM permission for describing RDS endpoints because of this change.

@gsagula I think the main concern here is not related to just using a RDS cluster being in creation state but also selecting randomly which cluster to put the installation in. Lets examine the senario that we have a RDS cluster with 5 installations and a new one is being brought up. If we start putting randomly installations in these two we may end up with the case that we have sth like 7 in the old and 5 in the new which will trigger another RDS cluster creation and then we may start using the third one although there is space in the first and the second. The limit of 10 installations is already a limit in the safe zone, so I don't think we should underutilise it even more but maybe @jwilander can give more input here.

@gsagula Regarding the other problem we faced about deletions, will it be fixed as part of another PR?

gsagula:
@stylianosrigas It makes sense, the RDS databases will be fill up in sequence and the provisioner will only bring in a new cluster if there is nothing available in AWS and the RDS cluster is in available status (that means, cluster and endpoints must be available). There is one situation which the RDS databases won't be filled up in order. It is when one RDS cluster is locked and another installation kicks in. If there is another cluster available, provisioner will select it and do the installation. This shouldn't happen often though. Regarding the teardown issue, it should be fixed in this PR.

PR #224, opened 2020-05-27
Stale installation supervisor fix

gabrieljackson:
A devious bug has been uncovered where supervisors can act on
installation data that has become "stale" due to other cloud servers
working on the installation beforehand.

This flow can occur when there are a high amount of installations
that require work and the multiple cloud servers are working on
the same tasks. They could end up getting a slice of installations
that need work at the same time and then processing those installations
independently over time with the outdated information regarding state.

This change forces the installation supervisor to check state on
these installations and abort if the state is outdated. When the
supervisor wakes up later it will get a new set of work and will
proceed.

This also cleans up some multitenant database logging.

https://mattermost.atlassian.net/browse/MM-25593

```release-note
Stale installation supervisor fix
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -116,13 +116,28 @@ func (s *InstallationSupervisor) Supervise(installation *model.Installation) {
 	}
 	defer lock.Unlock()
 
+	// Before working on the installation, it is crucial that we ensure that it
+	// was not updated to a new state by another provisioning server.
+	originalState := installation.State
+	installation, err := s.store.GetInstallation(installation.ID, true, false)
+	if err != nil {
+		logger.WithError(err).Errorf("Failed to get refreshed installation")
+		return
+	}
+	if installation.State != originalState {
+		logger.WithField("oldInstallationState", originalState).
+			WithField("newInstallationState", installation.State).
+			Warn("Another provisioner has worked on this installation; skipping...")
+		return
+	}
+
 	logger.Debugf("Supervising installation in state %s", installation.State)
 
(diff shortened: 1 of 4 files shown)


internal/supervisor/installation.go
```diff
@@ -116,13 +116,28 @@ func (s *InstallationSupervisor) Supervise(installation *model.Installation) {
...
+	// Before working on the installation, it is crucial that we ensure that it
+	// was not updated to a new state by another provisioning server.
+	originalState := installation.State
+	installation, err := s.store.GetInstallation(installation.ID, true, false)
+	if err != nil {
+		logger.WithError(err).Errorf("Failed to get refreshed installation")
+		return
+	}
+	if installation.State != originalState {
+		logger.WithField("oldInstallationState", originalState).
```
gsagula, line 128: nit: you can replace this chain with `logger.WithFields()`

internal/supervisor/installation.go
```diff
@@ -116,13 +116,28 @@ func (s *InstallationSupervisor) Supervise(installation *model.Installation) {
...
+			WithField("newInstallationState", installation.State).
+			Warn("Another provisioner has worked on this installation; skipping...")
```
gsagula, line 130: Is another provisioner working on the installation something normal? What is the reason for warning?

gabrieljackson: Yeah, it is. Multiple provisioners could end up working together to move any cloud objects from the beginning state to an end state.

That's how the `test` account and others are working now.

The warning is there to log that this stale state issue is happening in our accounts so I can verify that the fix is working. We may just silently exit in the future.

PR #225, opened 2020-05-28
Stale cluster and cluster installation supervisor fix

gabrieljackson:
This is a follow-up fix for the cluster and cluster installation
supervisors to prevent them from working on stale cloud resources.
The supervisors will now check if their workload contains a resource
that has already been worked on by another supervisor. If this is
detected, the supervisor will reevaluate all resources on its next
work cycle.

https://mattermost.atlassian.net/browse/MM-25628

```release-note
Stale cluster and cluster installation supervisor fix
```

### DIFF
--- a/internal/supervisor/cluster_installation.go
+++ b/internal/supervisor/cluster_installation.go
@@ -86,11 +86,26 @@ func (s *ClusterInstallationSupervisor) Supervise(clusterInstallation *model.Clu
 	}
 	defer lock.Unlock()
 
+	// Before working on the cluster installation, it is crucial that we ensure
+	// that it was not updated to a new state by another provisioning server.
+	originalState := clusterInstallation.State
+	clusterInstallation, err := s.store.GetClusterInstallation(clusterInstallation.ID)
+	if err != nil {
+		logger.WithError(err).Errorf("Failed to get refreshed cluster installation")
+		return
+	}
+	if clusterInstallation.State != originalState {
+		logger.WithField("oldClusterInstallationState", originalState).
+			WithField("newClusterInstallationState", clusterInstallation.State).
+			Warn("Another provisioner has worked on this cluster installation; skipping...")
+		return
+	}
+
 	logger.Debugf("Supervising cluster installation in state %s", clusterInstallation.State)
 
 	newState := s.transitionClusterInstallation(clusterInstallation, logger)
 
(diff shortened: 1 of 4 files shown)


internal/supervisor/cluster_installation.go
```diff
@@ -86,11 +86,26 @@ func (s *ClusterInstallationSupervisor) Supervise(clusterInstallation *model.Clu
 	}
 	defer lock.Unlock()
 
+	// Before working on the cluster installation, it is crucial that we ensure
```
gigawhitlocks, line 89: Doesn't this still leave open the chance for there to be a race condition? I'm honestly a little confused as to why the locking mechanism isn't the appropriate place for this logic. If we called `TryLock()` on line 84, why doesn't that fail if another Provisioner has started work on this installation?

If we had 2 Provisioners both calling `Supervise` at the same time, isn't it still possible that the state of the clusterInstallation could be changed by another Provisioner between lines 92 and 106?

gigawhitlocks: I recognize that this fix is the same as the one in your PR from yesterday, so I apologize for the belated questioning. I am just trying to fully internalize the issue that was found and the resolution.

gabrieljackson: The lock prevents us from working on the same thing at the same time with two provisioners. The issue appears when one provisioner ends up working on an object and finishes, meaning it locked+unlocked the object.

At a later time, another provisioner will succeed in getting the lock, but the copy of the resource it is working on is not outdated. When this is detected, we play it safe a abort and retry on the next work cycle.

gabrieljackson: To follow up on the `TryLock()` question, the locking mechanism shouldn't care about state of the object imo as there a many reasons something may want to lock and object and in some of those cases state may not matter at all.

In resolving this issue, we lock first, then verify state is not stale and proceed. That seems correct to me, but let me know if you have other thoughts on it.

jwilander:
Also giving my go ahead to merge this without infra review

internal/supervisor/cluster_installation.go
gigawhitlocks: No this clears up my concerns, thank you for taking the time to explain it :)

PR #226, opened 2020-05-28
MM-25633: improved error and log messages produced by the multitenant database

gsagula:
#### Summary
This PR improved error and log messages produced by the multitenant database.

#### Ticket Link
Jira https://mattermost.atlassian.net/browse/MM-25633

#### Release Notes
```release-note
NONE
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -115,6 +115,8 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		}
 	}
 
+	logger = logger.WithField("vpc-id", clusterResources.VpcID)
+
 	err = kops.CreateCluster(
 		kopsMetadata.Name,
 		kopsMetadata.Version,
--- a/internal/tools/aws/database_multitenant.go
+++ b/internal/tools/aws/database_multitenant.go
@@ -27,6 +27,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// MultitenantDatabaseNameLogField is a logger field that is carried accross all methods of a MultitenantDatabase.
+	MultitenantDatabaseNameLogField = "multitenant-database-rds"
+
+	// MultitenantDatabaseIDLogField is a logger field that is carried accross most of the methods of a MultitenantDatabase.
+	MultitenantDatabaseIDLogField = "multitenant-database-id"
+)
+
 // SQLDatabaseManager is an interface that describes operations to query and to
 // close connection with a database. It's used mainly to implement a client that
 // needs to perform non-complex queries in a SQL database instance.
@@ -55,7 +63,7 @@ func NewRDSMultitenantDatabase(instanceID, installationID string, client *Client
 // Teardown removes all AWS resources related to a RDS multitenant database.
 func (d *RDSMultitenantDatabase) Teardown(store model.InstallationDatabaseStoreInterface, keepData bool, logger log.FieldLogger) error {
 	installationDatabaseName := MattermostRDSDatabaseName(d.installationID)
-	logger = logger.WithField("rds-multitenant-database", installationDatabaseName)
+	logger = logger.WithField(MultitenantDatabaseNameLogField, installationDatabaseName)
 
 	logger.Info("Tearing down multitenant database")
 
@@ -79,22 +87,22 @@ func (d *RDSMultitenantDatabase) Teardown(store model.InstallationDatabaseStoreI
 		return nil
 	}
 
+	logger = logger.WithField(MultitenantDatabaseIDLogField, multitenantDatabases[0].ID)
+
 	unlocked, err := d.lockMultitenantDatabase(multitenantDatabases[0].ID, store)
 	if err != nil {
-		return errors.Wrap(err, "failed to teardown multitenant database")
+		return errors.Wrap(err, "failed to teardown multitenant database because provisioner is unable to lock the the resource")
 	}
 	defer unlocked(logger)
 
 	rdsCluster, err := d.getMultitenantDatabaseRDSCluster(multitenantDatabases[0].ID)
 	if err != nil {
-		return errors.Wrap(err, "failed to teardown multitenant database")
+		return errors.Wrap(err, "failed to teardown multitenant database because provisioner is unable to fetch the AWS RDS cluster resource")
 	}
-
-	logger = logger.WithField("rds-cluster-id", *rdsCluster.DBClusterIdentifier)
 
 	err = d.dropDatabaseAndDeleteSecret(multitenantDatabases[0].ID, *rdsCluster.Endpoint, installationDatabaseName, store, logger)
 	if err != nil {
-		return errors.Wrap(err, "failed to teardown multitenant database")
+		return errors.Wrap(err, "failed to teardown multitenant database because provisioner is unable to drop the installation database")
 	}
 
 	err = d.updateTagCounterAndRemoveInstallationID(rdsCluster.DBClusterArn, multitenantDatabases[0], store, logger)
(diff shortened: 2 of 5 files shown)


internal/tools/aws/database_multitenant.go
```diff
@@ -692,19 +697,20 @@ func (d *RDSMultitenantDatabase) updateTagCounterAndRemoveInstallationID(dbClust
...
-		logger.WithError(err).Warnf("Failed to remove multitenant database from datastore. Rolling tag:counter value back to %d", numOfInstallations)
+		logger.WithError(err).Warnf("Failed to remove multitenant database. Reseting %s to %d", RDSMultitenantInstallationCounterTagKey, numOfInstallations)
 		updateTagErr := d.updateCounterTag(dbClusterArn, numOfInstallations)
 		if updateTagErr != nil {
-			logger.WithError(err).Warnf("Failed to roll back tag:counter. Value is still %d", numOfInstallations-1)
+			logger.WithError(err).Warnf("Failed to reset %s. Value is still %d", RDSMultitenantInstallationCounterTagKey, numOfInstallations-1)
 		}
-		return errors.Wrapf(err, "failed to remove installation ID %s from datasore", d.installationID)
+
+		return errors.Wrap(err, "failed to remove multitenant")
```
gabrieljackson, line 713: Looks like this one ended a bit early?

gsagula:
> In the future, I would like to see us try to shorten these up and to use standard logging fields where we can. Those fields are very useful for getting through massive log piles quickly.

100% I'm happy to revisit this one. Should we create logging fields more specific for the multitenant database or databases in general?

internal/tools/aws/database_multitenant.go
```diff
@@ -206,36 +208,36 @@ func (d *RDSMultitenantDatabase) Provision(store model.InstallationDatabaseStore
...
 	databaseInstallationIDs, err := store.AddMultitenantDatabaseInstallationID(*lockedRDSCluster.cluster.DBClusterIdentifier, d.installationID)
 	if err != nil {
-		return errors.Wrapf(err, "failed to add installation ID %s to the multitenant database ID %s", d.installationID, *lockedRDSCluster.cluster.DBClusterIdentifier)
+		return errors.Wrap(err, "failed to provision multitenant database")
 	}
 
 	err = d.updateCounterTag(lockedRDSCluster.cluster.DBClusterArn, len(databaseInstallationIDs))
 	if err != nil {
-		return errors.Wrapf(err, "failed to update tag:counter in RDS cluster ID %s", *lockedRDSCluster.cluster.DBClusterIdentifier)
+		return errors.Wrap(err, "failed to provision multitenant database")
```
jwilander, line 236: Why be less descriptive with all these errors? Won't using `failed to provision multitenant database` in so many places make it difficult for us to find which exact line logged the error when debugging?

gsagula: Because we already have the IDs information in the logging fields. Unless I misunderstood @gabrieljackson [here](https://github.com/mattermost/mattermost-cloud/pull/226#pullrequestreview-420574030), I thought that the goal was to reduce the size of these messages by using fields.

jwilander: I agree with shorter errors and using the logging fields as much as possible but we should do that without losing details on the error. Also for all these error returns, they're not going to get the logger fields added in this function since we're not logging, just returning an error

jwilander:
@gsagula it looks like a few commits got pushed up after @gabrieljackson's review. Any time that happens we should re-request reviews

internal/tools/aws/database_multitenant.go
gsagula: We are logging when they bubble up to the provisioner.

jwilander: Right, I know that but now there's 5-10 log messages that all look the same and it will be hard to track where they originated

gsagula: Right, I don't disagree with having unique as possible error messages but I think we should have a discussion on how standardized the error/logs.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -115,6 +115,8 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		}
 	}
 
+	logger = logger.WithField("vpc-id", clusterResources.VpcID)
```
gsagula, line 118: @gabrieljackson I felt that having a field that tells me the VPC ID helps when debugging multi-tenant database issues, so I added it here.

internal/tools/aws/database_multitenant.go
```diff
@@ -206,36 +216,36 @@ func (d *RDSMultitenantDatabase) Provision(store model.InstallationDatabaseStore
...
 	databaseInstallationIDs, err := store.AddMultitenantDatabaseInstallationID(*lockedRDSCluster.cluster.DBClusterIdentifier, d.installationID)
 	if err != nil {
-		return errors.Wrapf(err, "failed to add installation ID %s to the multitenant database ID %s", d.installationID, *lockedRDSCluster.cluster.DBClusterIdentifier)
+		return errors.Wrap(err, "failed to provision multitenant database because provisioner is unable to add the installation id")
 	}
 
 	err = d.updateCounterTag(lockedRDSCluster.cluster.DBClusterArn, len(databaseInstallationIDs))
 	if err != nil {
-		return errors.Wrapf(err, "failed to update tag:counter in RDS cluster ID %s", *lockedRDSCluster.cluster.DBClusterIdentifier)
+		return errors.Wrap(err, "failed to provision multitenant database")
```
gabrieljackson, line 244: This doesn't explain anything about the counter update having an error.

internal/tools/aws/database_multitenant.go
```diff
@@ -206,36 +216,36 @@ func (d *RDSMultitenantDatabase) Provision(store model.InstallationDatabaseStore
 
 	err = d.createDatabaseIfNotExist(ctx, installationDatabaseName)
 	if err != nil {
-		return errors.Wrapf(err, "unable to create schema in multitenant RDS cluster ID %s", *lockedRDSCluster.cluster.DBClusterIdentifier)
+		return errors.Wrap(err, "failed to provision multitenant database because provisioner is unable to create a installation database")
```
gabrieljackson, line 219: This is an example of one that really should have the error prefix wrapped somewhere up the stack. For example this should be something like `errors.Wrap(err, "failed to create new installation database").

Because this is the `Provision` method, the caller should take any errors it receives and wrap them with something like `errors.Wrap(err, "failed to provision database").

There are quite a few errors around here that could benefit from this type of improvement.

gsagula:
I'm looking into this now. Thanks for the review.

internal/tools/aws/database_multitenant.go
```diff
@@ -141,12 +152,12 @@ func (d *RDSMultitenantDatabase) GenerateDatabaseSpecAndSecret(store model.Insta
 		SecretId: &installationSecretName,
 	})
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "failed to get multitenant database spec and secret from AWS Secrets Manager Service")
+		return nil, nil, errors.Wrap(err, "failed to create a secret and spec for the multitenant database because provisioner is unable to find a secret for this installation")
```
gabrieljackson, line 155: Example:
```
failed to create a secret and spec for the multitenant database because provisioner
```

internal/tools/aws/database_multitenant.go
```diff
@@ -54,53 +62,55 @@ func NewRDSMultitenantDatabase(instanceID, installationID string, client *Client
...
-		return errors.Wrapf(err, "failed to lock multitenant database ID %s when tearing the multitenant database down", multitenantDatabases[0].ID)
+		return errors.Wrap(err, "failed to teardown multitenant database because provisioner is unable to lock the the resource")
 	}
 	defer unlocked(logger)
 
-	rdsCluster, err := d.describeRDSCluster(multitenantDatabases[0].ID)
+	rdsCluster, err := d.getMultitenantDatabaseRDSCluster(multitenantDatabases[0].ID)
 	if err != nil {
-		return errors.Wrapf(err, "failed to describe multitenant database ID %s when tearing the multitenant database down", multitenantDatabases[0].ID)
+		return errors.Wrap(err, "failed to teardown multitenant database because provisioner is unable to fetch the AWS RDS cluster resource")
```
gabrieljackson, line 100: Example:
```
failed to teardown multitenant database because provisioner
```

PR #229, opened 2020-06-01
Creating an additional output field in fluent-bit for security audit logs

angeloskyratzakos:
#### Summary
Creation of an additional output field in fluent-bit for sending security audit logs to SumoLogic

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24417

#### Release Note

```release-note
fluent-bit sending security audit logs 
```


internal/provisioner/fluentbit.go
```diff
@@ -80,12 +79,38 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
 	}
+	domainNameSplitted := strings.Split(privateDomainName, ".")
+	sumoLogic := `[OUTPUT]
+	Name  forward
+	Match  *
+	Host  example.placeholder.mattermost.com
+	Port  %s
+	tls  On
+	tls.verify  Off
+	#tls.ca_file  /etc/certs/ca-chain.pem`
```
gigawhitlocks, line 90: let's delete this commented-out option

internal/provisioner/fluentbit.go
```diff
@@ -80,12 +79,38 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+
+	env := domainNameSplitted[1]
```
gabrieljackson, line 92: This could cause a panic if we don't check the length of the split domain name first. We should do that just in case.

internal/provisioner/fluentbit.go
```diff
@@ -80,12 +79,38 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+	var sumoLogicConf string
+	if env == "prod" {
```
gabrieljackson, line 94: Using a switch statement here would probably clean up the code quite a bit! https://tour.golang.org/flowcontrol/9

internal/provisioner/fluentbit.go
```diff
@@ -80,12 +79,38 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+		sumoLogicConf = fmt.Sprintf(sumoLogic, "24222")
+	} else if env == "staging" {
+		sumoLogicConf = fmt.Sprintf(sumoLogic, "24011")
+	} else if env == "test" {
+		sumoLogicConf = fmt.Sprintf(sumoLogic, "24022")
+	}
```
gabrieljackson, line 100: If we don't get a match should we do something or error, or is this okay?

internal/provisioner/fluentbit.go
```diff
@@ -80,12 +79,38 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
 	if err != nil {
 		logger.WithError(err).Error("unable to lookup private zone name")
 	}
+	domainNameSplitted := strings.Split(privateDomainName, ".")
+	sumoLogic := `[OUTPUT]
+	Name  forward
+	Match  *
+	Host  example.placeholder.mattermost.com
```
angeloskyratzakos, line 86: I need to discuss with @DSchalla if it is ok to have hardcoded the proper URL which is internal

internal/provisioner/fluentbit.go
angeloskyratzakos: Initially I tried with switch, but then it looked messier so I changed with a simple if/else statement

internal/provisioner/fluentbit.go
angeloskyratzakos: we want to be empty as if it is a new environment or dev we don't want to send audit logs on a wrong port, I had an else configuration like that one:
```
else {
		sumoLogicConf = fmt.Sprintf(" ")
	}
```
but I thought it was pointless

gabrieljackson: Okay, sounds good! Yeah, if it is okay if the value is blank then no need to do anything. It's default value will be `""` which is perfect.

internal/provisioner/fluentbit.go
angeloskyratzakos: will do! Thanks Ian!

internal/provisioner/fluentbit.go
gabrieljackson: Yeah, not a big deal at all, but you should be able to do something like this if you want:

```
var sumoLogicConf string
switch domainNameSplitted[1] {
case "prod":
	sumoLogicConf = fmt.Sprintf(sumoLogic, "24222")
case "staging":
	sumoLogicConf = fmt.Sprintf(sumoLogic, "24011")
case "test":
	sumoLogicConf = fmt.Sprintf(sumoLogic, "24022")
}
```

internal/provisioner/fluentbit.go
gabrieljackson: We may want to template that out as well into an env var or something so we can update it easily in the future.

internal/provisioner/fluentbit.go
angeloskyratzakos: Ahhh I see, I did something like:
```
	switch sumoLogicConf := domainNameSplitted[1]; sumoLogicConf {
	case "prod":
		sumoLogicConf := fmt.Sprintf(sumoLogic, "24222")
	case "staging":
		sumoLogicConf := fmt.Sprintf(sumoLogic, "24011")
	case "test":
		sumoLogicConf := fmt.Sprintf(sumoLogic, "24022")
	default:
		sumoLogicConf := " "
	}
```
I followed the wrong example probable
Thanks Gabe for the help!

internal/provisioner/fluentbit.go
DSchalla: Yes, fully agree what @gabrieljackson said.

internal/provisioner/fluentbit.go
```diff
@@ -80,17 +79,53 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+	if err != nil {
+		logger.WithError(err).Error("unable to find Zone ID")
+	}
+
+	tag, err := f.awsClient.GetTagByKeyAndZoneID(aws.DefaultAuditLogsCoreSecurityTagKey, zoneID, logger)
+	if err != nil {
+		logger.WithError(err).Error("failed to find Tag:AuditLogsCoreSecurity")
+	}
+	if tag == nil {
+		logger.Info("unable to find tag")
```
gigawhitlocks, line 93: Should this be a Warning? It seems a little strange that the error message on line 90 is the same as this one. I see the difference between the cases, but it seems a little odd to me that both errors are essentially the same (and handled the same way) but one is INFO level and the other is ERROR level. In my opinion both should be WARNING if they are severe enough to warrant the log message but inconsequential enough that execution continues.

internal/provisioner/fluentbit.go
```diff
@@ -80,17 +79,53 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+		tag = &aws.Tag{}
+	}
+
```
gigawhitlocks, line 96: if `err` was not `nil`, then `tag` is `nil` at this point and line 97 will panic

internal/tools/aws/route53.go
```diff
@@ -47,15 +47,58 @@ func (a *Client) CreatePrivateCNAME(dnsName string, dnsEndpoints []string, logge
...
+	id, err := a.GetPrivateZoneID(logger)
+	if err != nil {
+		return "", err
+	}
+	return a.getZoneDNS(id, logger)
+}
+
+// GetPrivateZoneID returns the R53 hosted zone ID for a given tag or if not provided for the default one
+func (a *Client) GetPrivateZoneID(logger log.FieldLogger, tags ...Tag) (string, error) {
+	if len(tags) > 1 {
```
gigawhitlocks, line 59: If this was changed to `if len(tags != 1)` then the conditional on line 66 could be removed.

angeloskyratzakos: if we do that, function will return empty when we will have `len(tags) == 0` 🤔  right?

internal/provisioner/fluentbit.go
angeloskyratzakos: I think on line 90 it is an error from trying to retrieve the message, while on line 93 the message should be more informative that there wasn't any tag so "skipping setup" probably I should change the message of line 93. Does it make sense?

internal/tools/aws/route53.go
gabrieljackson: Do we maybe just want to split this function up? One function for the default and one for tag(s)? It would probably make more sense when reading the code as well.

angeloskyratzakos: @gabrieljackson I removed the rest of the code because currently is not used so I left the function as is for getting the default tags

internal/provisioner/fluentbit.go
gigawhitlocks: >the message should be more informative that there wasn't any tag so "skipping setup"

Yeah this makes sense to me. Just adding that little extra context would be awesome.

internal/tools/aws/route53.go
gigawhitlocks: > if we do that, function will return empty when we will have len(tags) == 0 thinking right?

That doesn't seem right, if `len(tags)==0` then `tag` is kept as `tag := Tag{...` on line 62. That would be used in the query to `getHostedZoneIDWithTag`.

gigawhitlocks: > if we do that, function will return empty when we will have len(tags) == 0 thinking right?

No, if you provided an empty `tags` list, it would return the error on line 60.
I also made a typo in my suggestion, I meant to suggest `len(tags) != 1`.

gigawhitlocks: > we do that, function will return empty when we will have len(tags) == 0

It'll return `"", err`, on line 60. I can see how this would alter the default behavior, so I think you can disregard this comment, actually. I even made a typo in my original comment so either your code changed or I didn't fully understand the code when I left that comment.

internal/provisioner/fluentbit.go
angeloskyratzakos: You are right! I need to change it with a better context! will add the skipping setup

internal/provisioner/fluentbit.go
```diff
@@ -80,17 +79,53 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
 	}
+
+	var auditLogsConf string
+	zoneID, err := f.awsClient.GetPrivateZoneIDForDefaultTag(logger)
+	if err != nil {
+		logger.WithError(err).Error("unable to get Private Zone ID with the default tag, skipping setup...")
+	} else {
+		tag, err := f.awsClient.GetTagByKeyAndZoneID(aws.DefaultAuditLogsCoreSecurityTagKey, zoneID, logger)
+		if err != nil {
+			logger.WithError(err).Error("unable to find Tag:AuditLogsCoreSecurity")
```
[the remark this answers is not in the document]
angeloskyratzakos: fixed! I haven't thought it at all! thanks! 🎉

internal/provisioner/fluentbit.go
```diff
@@ -80,17 +79,53 @@ func (f *fluentbit) NewHelmDeployment(logger log.FieldLogger) *helmDeployment {
...
+		}
+		if tag == nil {
+			logger.Info("Tag:AuditLogsCoreSecurity is missing, skipping setup...")
```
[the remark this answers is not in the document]
angeloskyratzakos: fixed!

PR #231, opened 2020-06-02
The versioning for kops now includes a `v` prefix

gabrieljackson:
The GitHub tags for kops have changed since we last updated our
kops binary. The downloads now seem to be found under version tags
starting with `v`. Our build tooling currently does not throw an
error when grabbing a kops binary that doesn't exist. This will
be fixed in a future change.

```release-note
The versioning for kops now includes a `v` prefix
```


gabrieljackson:
I manually tested this to make sure an actual linux binary was downloaded.

gabrieljackson:
Important to get this in, so merging now.

PR #236, opened 2020-06-04
Refactor cluster metadata handling

gabrieljackson:
This change alters the behavior of accessing cluster metadata.
Before, it was presented in the cluster object in raw format which
required marshaling and unmarshaling to access and update. This
logic is now handled directly on the store level so the returned
clusters now have metadata in their actual types. When cluster
updates are made, the store will marshal these types back into
their raw format before being stored.

Note: further work and cleanup on some of the utility metadata
logic will be made in a follow-up PR.

https://mattermost.atlassian.net/browse/MM-25793

```release-note
Refactor cluster metadata handling
```

### DIFF
--- a/internal/api/cluster.go
+++ b/internal/api/cluster.go
@@ -81,39 +81,25 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	cluster := model.Cluster{
-		Provider:           createClusterRequest.Provider,
-		Provisioner:        "kops",
+		Provider: createClusterRequest.Provider,
+		ProviderMetadataAWS: &model.AWSMetadata{
+			Zones: createClusterRequest.Zones,
+		},
+		Provisioner: "kops",
+		ProvisionerMetadataKops: &model.KopsMetadata{
+			Version: createClusterRequest.Version,
+			AMI:     createClusterRequest.KopsAMI,
+		},
 		Version:            "0.0.0",
 		Size:               createClusterRequest.Size,
 		AllowInstallations: createClusterRequest.AllowInstallations,
 		State:              model.ClusterStateCreationRequested,
 	}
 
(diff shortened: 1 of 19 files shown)


internal/api/cluster.go
```diff
@@ -575,5 +548,5 @@ func handleGetAllUtilityMetadata(c *Context, w http.ResponseWriter, r *http.Requ
 
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
-	w.Write(cluster.UtilityMetadata)
+	outputJSON(c, w, cluster.UtilityMetadata)
```
gigawhitlocks, line 551: Can we change this to `writeJSON` so that it's consistent with `WriteHeader`? 0/5

internal/store/migrations.go
```diff
@@ -921,6 +921,65 @@ var migrations = []migration{
 			return err
 		}
 
+		return nil
+	}},
+	{semver.MustParse("0.18.0"), semver.MustParse("0.19.0"), func(e execer) error {
+		_, err := e.Exec(`ALTER TABLE Cluster RENAME TO ClusterTemp;`)
```
gigawhitlocks, line 927: I don't think you need to do this. I'm pretty sure you can just do `ALTER TABLE Cluster ADD COLUMN` for each of the columns you need. I could swear I tested this with both Postgres and sqlite and it works with both.

internal/api/cluster.go
gabrieljackson: Sounds good. I just reused the the existing function, but I will update the name on my next pass.

internal/store/migrations.go
gabrieljackson: Yeah, a total of three columns were added and three were removed (basically a rename) so this felt a little more straightforward this time.

PR #239, opened 2020-06-08
Refactor cluster metadata part II

gabrieljackson:
With this change, cluster objects are now stored and presented in
a fully-descriptive way. All metadata now directly references the
active state of the cluster or requests to be made. To accomplish
this, the following changes were made:

 - Added master and worker node types and counts. These values can
   now be individually updated.
 - Removed cluster size value. Due to the increasing need to make
   granular cluster changes, the size value was becoming too
   inflexible to support.
 - Cluster size values refactored into a "cluster dictionary". This
   provides the cloud client with an simple way to lookup basic
   cluster resource values from a simple size keyword.
 - Individual cluster size values on cluster create and cluster
   resize. When these values are specified, they will overwrite any
   values obtained from a cluster dictionary lookup.
 - Cluster upgrades/downgrades and resizes now fully support
   patching logic. Any value can update updated individually.
 - Added a `--dry-run` flag to the `cluster create` and
   `cluster-resize` commands. This outputs the final API payload
   without sending it for verification.
 - Added cluster `refresh-metadata` state and flow. This is called
   after cluster creation, provisioning, upgrade, or resize and
   refreshes the cluster metadata with current kops values.
 - Moved default kops logging level from warning to info. I have
   no idea why I didn't do this earlier.

Reviewer notes:

I tried my best regarding testing. Many manual tests were performed
but without being able to unit test the kops integration yet, there
could be some ghosts in the machine.

The old size and version database columns were left for now. The
thought was these could be used in case a rollback is needed for
any reason and can be removed at a later date.

Also, the `-HA` option was removed from the size keywords as
you can simply use size overwrites to get multiple masters.

Lastly, although you can set a min and max worker node count
now, more work needs to be done to support autoscaling.

https://mattermost.atlassian.net/browse/MM-25793

```release-note
 - Refactor cluster metadata and remove size value
 - Support direct cluster value updates
 - Add --dry-run flag for cluster creation and resize
```


internal/tools/kops/instance_groups.go
```diff
@@ -20,7 +22,80 @@ type InstanceGroupMetadata struct {
...
+
+	var masterIGCount, NodeIGCount, nodeMinCount, nodeMaxCount int64
+	var masterMachineType, nodeInstanceType, AMI string
+	for _, ig := range instanceGroups {
+		switch ig.Spec.Role {
+		case "Master":
+			if AMI == "" {
+				AMI = ig.Spec.Image
+			} else if AMI != ig.Spec.Image {
+				metadata.AddWarning(fmt.Sprintf("Expected all hosts to be running same AMI, but instance group %s is has AMI %s", ig.Metadata.Name, ig.Spec.Image))
```
jwilander, line 58: "is has" typo here and in the warnings below

internal/tools/kops/instance_groups.go
```diff
@@ -20,7 +22,80 @@ type InstanceGroupMetadata struct {
...
+		default:
+			metadata.AddWarning(fmt.Sprintf("Instance group %s has unknown role %s", ig.Metadata.Name, ig.Spec.Role))
+		}
+	}
+
+	if masterIGCount == 0 {
+		metadata.AddWarning("Failed to find any master instance groups")
+	}
+	if NodeIGCount != 1 {
+		metadata.AddWarning(fmt.Sprintf("expected exactly 1 node instance group, but found %d", NodeIGCount))
```
jwilander, line 88: Interesting approach with storing the warnings with the metadata. If I'm reading this right we don't log the warnings any where which would mean the only way we'd notice them is if looked at the cluster itself, right? I imagine you want to store them as they're technically part of the kops state? Should we log them or set up some grafana alerts on the cluster table so we are aware of them?

gabrieljackson: Yeah, that's exactly right: the warnings are stored with the metadata if there are any. I initially started with logging the warnings, but found that having them presented with the cluster metadata worked really well. Basically, you can list your clusters and know that any kops validation errors will be right there if there are any. On the flip side, you know that no errors means validation was clean.

All that said, there is absolutely no reason we can't have our cake and eat it too so I just pushed up a change to log the warnings as well. Great idea on automated alerts!

PR #241, opened 2020-06-12
validate if the request has spaces in some fields

cpanato:
#### Summary
If we create an installation and the dns name have spaces in the end of the string, for example, we cannot create and then cannot complete the deletion of the installation because we receive this error:

```json
{"aws-operation-name":"ListResourceRecordSets","aws-service-id":"Route 53","instance":"pbkueecpu7nb8peosk8j3b41gy","level":"debug","msg":"[aws] GET 400 Bad Request https://route53.amazonaws.com/2013-04-01/hostedzone/Z1KG2H35T59O81/rrset?name=ee-demo.staging.cloud.mattermost.com%20 {\"HostedZoneId\":\"Z1KG2H35T59O81\",\"MaxItems\":null,\"StartRecordIdentifier\":null,\"StartRecordName\":\"ee-demo.staging.cloud.mattermost.com \",\"StartRecordType\":null}","time":"2020-06-11T14:56:47Z","tools-aws":"client"}
{"error":"InvalidInput: FATAL problem: UnsupportedCharacter (Value contains unsupported characters) encountered with ' '\n\tstatus code: 400, request id: 7c4f461b-b27b-404b-9656-c9431f1a4c80","installation":"9ifr3zjx6fyojkriqmtthyfx8y","instance":"pbkueecpu7nb8peosk8j3b41gy","level":"error","msg":"Failed to delete installation DNS","time":"2020-06-11T14:56:47Z"}
```

#### Release Note

```release-note
validate if the request has spaces in some fields
```

### DIFF
--- a/model/installation_request.go
+++ b/model/installation_request.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -81,7 +82,31 @@ func (request *CreateInstallationRequest) Validate() error {
 		return errors.Wrap(err, "invalid env var settings")
 	}
 
+	return checkSpaces(request)
+}
+
+func checkSpaces(request *CreateInstallationRequest) error {
+	if hasWhiteSpace(request.DNS) != -1 {
+		return errors.Errorf("cannot have spaces in dns field. DNS=%s", request.DNS)
+	}
+	if hasWhiteSpace(request.Version) != -1 {
+		return errors.Errorf("cannot have spaces in version field. Version=%s", request.Version)
+	}
+	if hasWhiteSpace(request.Image) != -1 {
+		return errors.Errorf("cannot have spaces in image field. Image=%s", request.Image)
+	}
+	if hasWhiteSpace(request.License) != -1 {
+		return errors.Errorf("cannot have spaces in license field. License=%s", request.License)
+	}
+	if hasWhiteSpace(request.GroupID) != -1 {
+		return errors.Errorf("cannot have spaces in group field. Group=%s", request.GroupID)
+	}
+
 	return nil
+}
+
+func hasWhiteSpace(value string) int {
+	return strings.IndexAny(value, " ")
 }
 
 // NewCreateInstallationRequestFromReader will create a CreateInstallationRequest from an io.Reader with JSON data.
--- a/model/installation_request_test.go
+++ b/model/installation_request_test.go
@@ -92,6 +92,25 @@ func TestCreateInstallationRequestValid(t *testing.T) {
 				},
 			},
 		},
+		{
+			"dns has space",
+			true,
+			&model.CreateInstallationRequest{
+				OwnerID: "owner1",
+				DNS:     "domain.com ",
+			},
+		},
+		{
+			"Group/Database/filestore is blank should not fail validation",
+			false,
+			&model.CreateInstallationRequest{
+				OwnerID:   "owner1",
+				DNS:       "domain.com",
+				GroupID:   "",
+				Filestore: "",
+				Database:  "",
+			},
+		},
 	}
 
 	for _, tc := range testCases {


cpanato:
open to feedback

internal/api/installation.go
```diff
@@ -128,16 +129,16 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
...
-		Size:          createInstallationRequest.Size,
-		Affinity:      createInstallationRequest.Affinity,
+		Version:       strings.TrimSpace(createInstallationRequest.Version),
+		Image:         strings.TrimSpace(createInstallationRequest.Image),
+		DNS:           strings.TrimSpace(createInstallationRequest.DNS),
+		Database:      strings.TrimSpace(createInstallationRequest.Database),
+		Filestore:     strings.TrimSpace(createInstallationRequest.Filestore),
+		License:       strings.TrimSpace(createInstallationRequest.License),
+		Size:          strings.TrimSpace(createInstallationRequest.Size),
+		Affinity:      strings.TrimSpace(createInstallationRequest.Affinity),
```
gabrieljackson, line 141: Hmm, instead of trimming spaces we should handle this in validation: https://github.com/mattermost/mattermost-cloud/blob/master/model/installation_request.go#L56-L61

I think we shouldn't silently succeed in creating a new installation when the dns entry had some bad characters. We should error and let the sender fix it.

internal/supervisor/installation.go
```diff
@@ -688,7 +689,7 @@ func (s *InstallationSupervisor) deleteInstallation(installation *model.Installa
 }
 
 func (s *InstallationSupervisor) finalDeletionCleanup(installation *model.Installation, logger log.FieldLogger) string {
-	err := s.aws.DeletePublicCNAME(installation.DNS, logger)
+	err := s.aws.DeletePublicCNAME(strings.TrimSpace(installation.DNS), logger)
```
gabrieljackson, line 692: I would like to see us not trim the value after we have stored it in the database. It should be valid before we store it.

Do we need this to cleanup a bad record in one of our accounts? I know we really try to avoid manually changing the DB, but I would prefer we do that here if we have a bad record.

cpanato: yes there is one bad thing on staging. but sounds good will refactor

internal/api/installation.go
cpanato: removed

internal/supervisor/installation.go
cpanato: done

model/installation_request.go
```diff
@@ -81,9 +82,45 @@ func (request *CreateInstallationRequest) Validate() error {
...
+	}
+	if hasWhiteSpace(request.Database) != -1 {
+		return errors.Errorf("cannot have spaces in database field. Database=%s", request.Database)
+	}
+	if hasWhiteSpace(request.Filestore) != -1 {
+		return errors.Errorf("cannot have spaces in filestore field. Filestore=%s", request.Filestore)
+	}
+	if hasWhiteSpace(request.Affinity) != -1 {
+		return errors.Errorf("cannot have spaces in affinity field. Affinity=%s", request.Affinity)
+	}
```
gabrieljackson, line 103: For all three of these, we have already checked that they are an exact match to a valid value with the methods such as `IsSupportedDatabase`. They should never have spaces so we can drop these checks.

model/installation_request.go
```diff
@@ -81,9 +82,45 @@ func (request *CreateInstallationRequest) Validate() error {
 		return errors.Wrap(err, "invalid env var settings")
 	}
 
+	return checkSpaces(request)
+}
+
+func checkSpaces(request *CreateInstallationRequest) error {
+	if hasWhiteSpace(request.OwnerID) != -1 {
+		return errors.Errorf("cannot have spaces in owner field. Owner=%s", request.OwnerID)
+	}
```
gabrieljackson, line 91: I think technically it's fine to have spaces in the OwnerID field. It's just metadata.

model/installation_request.go
cpanato: ok

model/installation_request.go
cpanato: right

PR #246, opened 2020-06-16
Add installation hibernation

gabrieljackson:
This change adds functionality to put installations into a new state
called "hibernation". The current effect of this state is to scale
down the Mattermost app deployment so that there are no running
pods. In the future, this could be enhanced to include database
and filestore as well.

Changes:
 - Add hibernation API, states, and flows.
 - Update Mattermost operator to v1.5.0 to support new scale values.
 - Refactor installation supervisor logic around checking if
   cluster installations are stable or not.
 - Correct bad webhook state event.
 - Logging fixes.

Fixes https://mattermost.atlassian.net/browse/MM-25099

```release-note
Add installation hibernation
```

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -32,9 +32,6 @@ func init() {
 	installationUpdateCmd.Flags().Bool("mattermost-env-clear", false, "Clears all env var data.")
 	installationUpdateCmd.MarkFlagRequired("installation")
 
-	installationDeleteCmd.Flags().String("installation", "", "The id of the installation to be deleted.")
-	installationDeleteCmd.MarkFlagRequired("installation")
-
 	installationGetCmd.Flags().String("installation", "", "The id of the installation to be fetched.")
 	installationGetCmd.Flags().Bool("include-group-config", true, "Whether to include group configuration in the installation or not.")
 	installationGetCmd.Flags().Bool("include-group-config-overrides", true, "Whether to include a group configuration override summary in the installation or not.")
@@ -48,9 +45,20 @@ func init() {
 	installationListCmd.Flags().Int("per-page", 100, "The number of installations to fetch per page.")
 	installationListCmd.Flags().Bool("include-deleted", false, "Whether to include deleted installations.")
 
+	installationHibernateCmd.Flags().String("installation", "", "The id of the installation to put into hibernation.")
+	installationHibernateCmd.MarkFlagRequired("installation")
+
+	installationWakeupCmd.Flags().String("installation", "", "The id of the installation to wake up from hibernation.")
+	installationWakeupCmd.MarkFlagRequired("installation")
+
+	installationDeleteCmd.Flags().String("installation", "", "The id of the installation to be deleted.")
+	installationDeleteCmd.MarkFlagRequired("installation")
+
 	installationCmd.AddCommand(installationCreateCmd)
 	installationCmd.AddCommand(installationUpdateCmd)
 	installationCmd.AddCommand(installationDeleteCmd)
+	installationCmd.AddCommand(installationHibernateCmd)
+	installationCmd.AddCommand(installationWakeupCmd)
 	installationCmd.AddCommand(installationGetCmd)
 	installationCmd.AddCommand(installationListCmd)
 	installationCmd.AddCommand(installationShowStateReport)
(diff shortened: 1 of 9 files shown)


cmd/cloud/installation.go
```diff
@@ -164,6 +172,56 @@ var installationDeleteCmd = &cobra.Command{
...
+	Short: "Wake an installation from hibernation.",
+	RunE: func(command *cobra.Command, args []string) error {
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		installationID, _ := command.Flags().GetString("installation")
+
+		installation, err := client.WakupInstallation(installationID)
```
gigawhitlocks, line 211: `WakupInstallation` -- pretty sure you want `WakeupInstallation`

internal/supervisor/installation.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/mattermost/mattermost-cloud/internal/webhook"
 	"github.com/mattermost/mattermost-cloud/model"
 	mmv1alpha1 "github.com/mattermost/mattermost-operator/pkg/apis/mattermost/v1alpha1"
+	"github.com/pkg/errors"
```
gigawhitlocks, line 12: I've been putting third party imports in a different section from the first party ones, e.g.

stdlib imports

third party imports

first party imports

Now, I _thought_ this was already a part of our style guidelines but it's absolutely possible that I imagined it, because this is an old habit of mine anyway. That said, if you're on board with that style, I'd like to see the imports follow that pattern here.

internal/supervisor/installation.go
```diff
@@ -206,7 +208,13 @@ func (s *InstallationSupervisor) transitionInstallation(installation *model.Inst
 		return s.updateInstallation(installation, instanceID, logger)
 
 	case model.InstallationStateUpdateInProgress:
-		return s.waitForUpdateComplete(installation, instanceID, logger)
+		return s.waitForUpdateStable(installation, instanceID, logger)
+
+	case model.InstallationStateHibernationRequested:
```
gigawhitlocks, line 213: What is the opposite state, when a Wakeup is requested?

internal/supervisor/installation.go
```diff
@@ -540,46 +519,109 @@ func (s *InstallationSupervisor) updateInstallation(installation *model.Installa
...
+		InstallationID: installation.ID,
 	})
 	if err != nil {
 		logger.WithError(err).Warn("Failed to find cluster installations")
 		return installation.State
 	}
 
-	var stable, reconciling, failed int
+	var clusterInstallationIDs []string
+	if len(clusterInstallations) > 0 {
```
gigawhitlocks, line 553: nit: could we invert this `if` and `return` early so we could outdent the inner block? I see you'd have to repeat the `return` statement but imho that's worth it. 0/5 though.

internal/supervisor/installation.go
```diff
@@ -721,3 +763,41 @@ func (s *InstallationSupervisor) finalCreationTasks(installation *model.Installa
...
+// Helper funcs
+
+func (s *InstallationSupervisor) checkIfClusterInstallationsAreStable(installation *model.Installation, logger log.FieldLogger) (bool, error) {
+	clusterInstallations, err := s.store.GetClusterInstallations(&model.ClusterInstallationFilter{
+		InstallationID: installation.ID,
+		PerPage:        model.AllPerPage,
+	})
+	if err != nil {
+		logger.WithError(err).Warn("Failed to find cluster installations")
+		return false, nil
```
gigawhitlocks, line 776: Just curious why this is `false, nil` instead of `false, err`? Another way I could phrase the question would be: why is this error merely a warning? The reason is not immediately obvious to me.

cmd/cloud/installation.go
gabrieljackson: But everyone loves `Wakup`!

internal/supervisor/installation.go
gabrieljackson: Ah, I meant to put a comment on that helper func to clear up that very question. Let me do that.

internal/supervisor/installation.go
gabrieljackson: Yeah, that is absolutely what I try to do as well. My editor added this and I didn't review it. I will fix it shortly.

internal/supervisor/installation.go
gabrieljackson: It does the normal update flow. This is to ensure that whatever we have in the object configuration is put into the running state.

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -155,6 +155,53 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
...
+	defer kops.Close()
+
+	err = kops.ExportKubecfg(cluster.ProvisionerMetadataKops.Name)
+	if err != nil {
+		return errors.Wrap(err, "failed to export kubecfg")
+	}
+
+	k8sClient, err := k8s.New(kops.GetKubeConfigPath(), logger)
+	if err != nil {
+		return err
```
stylianosrigas, line 179: maybe wrap the error here?

PR #249, opened 2020-06-18
adding some json map when using the api

cpanato:
#### Summary
Add the json in some structs to have better output when using the API

and also helps my life because sometime I copy the response and need to format to use as input in the next request, with this we make standard.

Also i think is a good practice to have that. not sure.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/model/cluster.go
+++ b/model/cluster.go
@@ -8,18 +8,18 @@ import (
 
 // Cluster represents a Kubernetes cluster.
 type Cluster struct {
-	ID                      string
-	State                   string
-	Provider                string
-	ProviderMetadataAWS     *AWSMetadata
-	Provisioner             string
-	ProvisionerMetadataKops *KopsMetadata
-	UtilityMetadata         *UtilityMetadata
-	AllowInstallations      bool
-	CreateAt                int64
-	DeleteAt                int64
-	LockAcquiredBy          *string
-	LockAcquiredAt          int64
+	ID                      string           `json:"id,omitempty"`
+	State                   string           `json:"state,omitempty"`
+	Provider                string           `json:"provider,omitempty"`
+	ProviderMetadataAWS     *AWSMetadata     `json:"providerMetadataAWS,omitempty"`
+	Provisioner             string           `json:"provisioner,omitempty"`
+	ProvisionerMetadataKops *KopsMetadata    `json:"provisionerMetadataKops,omitempty"`
+	UtilityMetadata         *UtilityMetadata `json:"utilityMetadata,omitempty"`
+	AllowInstallations      bool             `json:"allowInstallations,omitempty"`
+	CreateAt                int64            `json:"createAt,omitempty"`
+	DeleteAt                int64            `json:"deleteAt,omitempty"`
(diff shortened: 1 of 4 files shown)


model/cluster.go
```diff
@@ -8,18 +8,18 @@ import (
...
+	Provider                string           `json:"provider,omitempty"`
+	ProviderMetadataAWS     *AWSMetadata     `json:"providerMetadataAWS,omitempty"`
+	Provisioner             string           `json:"provisioner,omitempty"`
+	ProvisionerMetadataKops *KopsMetadata    `json:"provisionerMetadataKops,omitempty"`
+	UtilityMetadata         *UtilityMetadata `json:"utilityMetadata,omitempty"`
+	AllowInstallations      bool             `json:"allowInstallations,omitempty"`
+	CreateAt                int64            `json:"createAt,omitempty"`
+	DeleteAt                int64            `json:"deleteAt,omitempty"`
+	LockAcquiredBy          *string          `json:"lockAcquiredBy,omitempty"`
+	LockAcquiredAt          int64            `json:"lockAcquiredAt,omitempty"`
```
gabrieljackson, line 22: So, this is sort of subjective, but I like not omitting these values if they are empty. I would like to see that `LockAcquiredBy` is null when it is not locked. Same thing with `LockAcquiredAt` and `DeleteAt`. This gets tricky because it's just personal preference, but I have gotten used to seeing it. Thoughts?

model/cluster_installation.go
```diff
@@ -7,15 +7,15 @@ import (
...
-	LockAcquiredAt int64
+	ID             string  `json:"id,omitempty"`
+	ClusterID      string  `json:"clusterId,omitempty"`
+	InstallationID string  `json:"installationId,omitempty"`
+	Namespace      string  `json:"namespace,omitempty"`
+	State          string  `json:"state,omitempty"`
+	CreateAt       int64   `json:"createAt,omitempty"`
+	DeleteAt       int64   `json:"deleteAt,omitempty"`
+	LockAcquiredBy *string `json:"lockAcquiredBy,omitempty"`
+	LockAcquiredAt int64   `json:"lockAcquiredAt,omitempty"`
```
gabrieljackson, line 18: Same thing here.

model/group.go
```diff
@@ -7,18 +7,18 @@ import (
...
+	Name           string    `json:"name,omitempty"`
+	Description    string    `json:"description,omitempty"`
+	Version        string    `json:"version,omitempty"`
+	Image          string    `json:"image,omitempty"`
+	MaxRolling     int64     `json:"maxRolling,omitempty"`
+	MattermostEnv  EnvVarMap `json:"mattermostEnv,omitempty"`
+	CreateAt       int64     `json:"createAt,omitempty"`
+	DeleteAt       int64     `json:"deleteAt,omitempty"`
+	LockAcquiredBy *string   `json:"lockAcquiredBy,omitempty"`
+	LockAcquiredAt int64     `json:"lockAcquiredAt,omitempty"`
```
gabrieljackson, line 21: Same thing again

cpanato:
@gabrieljackson i will rethink for now i will close this

PR #259, opened 2020-07-01
Add cluster scale up logic

gabrieljackson:
This adds logic for scaling up cluster worker nodes when scheduling
new installations. If the cloud server is configured to do so, worker
nodes will be increased up to the cluster worker node maximum if the
new installation would exceed the cluster resource threshold. The
cluster scaling and installation reconciliation will happen in
parallel.

Notes on using cluster autoscaling:
To leverage this new functionality, two things need to be done. The
server flag `cluster-resource-threshold-scale-value` needs to be set.
A value of `2` or `3` seems like a good starting spot. The clusters that
you would like to enable autoscaling on need to have their max node
count updated to whatever you are comfortable with.

Note that this does not implement scaling down cluster nodes. Scaling
down can be done manually for now.

Fixes https://mattermost.atlassian.net/browse/MM-25093

```release-note
Add cluster scale up logic
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -6,7 +6,6 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"net/http"
 	"os"
@@ -47,6 +46,7 @@ func init() {
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().Int("poll", 30, "The interval in seconds to poll for background work.")
 	serverCmd.PersistentFlags().Int("cluster-resource-threshold", 80, "The percent threshold where new installations won't be scheduled on a multi-tenant cluster.")
+	serverCmd.PersistentFlags().Int("cluster-resource-threshold-scale-value", 0, "The number of worker nodes to scale up by when the threshold is passed. Set to 0 for no scaling. Scaling will never exceed the cluster max worker configuration value.")
 	serverCmd.PersistentFlags().Bool("use-existing-aws-resources", true, "Whether to use existing AWS resources (VPCs, subnets, etc.) or not.")
 	serverCmd.PersistentFlags().Bool("keep-database-data", true, "Whether to preserve database data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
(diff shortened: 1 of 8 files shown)


cmd/cloud/server.go
```diff
@@ -183,7 +189,7 @@ var serverCmd = &cobra.Command{
 			multiDoer = append(multiDoer, supervisor.NewGroupSupervisor(sqlStore, instanceID, logger))
 		}
 		if installationSupervisor {
-			multiDoer = append(multiDoer, supervisor.NewInstallationSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, clusterResourceThreshold, keepDatabaseData, keepFilestoreData, resourceUtil, logger))
+			multiDoer = append(multiDoer, supervisor.NewInstallationSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, clusterResourceThreshold, clusterResourceThresholdScaleValue, keepDatabaseData, keepFilestoreData, resourceUtil, logger))
```
gigawhitlocks, line 192: This is not a suggestion for this PR but we should probably think about refactoring `NewInstallationSupervisor` to take a struct just because 10 positional arguments is a bit unwieldy lol

internal/supervisor/installation.go
```diff
@@ -328,12 +331,12 @@ func (s *InstallationSupervisor) createClusterInstallation(cluster *model.Cluste
 
 	size, err := mmv1alpha1.GetClusterSize(installation.Size)
 	if err != nil {
-		logger.WithError(err).Error("invalid cluster installation size")
+		logger.WithError(err).Error("Invalid cluster installation size")
```
gigawhitlocks, line 334: Nitpick but I don't think these error messages should start with a capital since Go concatenates them together as substrings with colon

cmd/cloud/server.go
gabrieljackson: Totally agree. The two server threshold flags should eventually be moved to be part of the cluster object, but this allows us to see if we like the scaling or not before we do so.

I will create a ticket to tackle this later.

internal/supervisor/installation.go
```diff
@@ -350,8 +353,51 @@ func (s *InstallationSupervisor) createClusterInstallation(cluster *model.Cluste
 		),
 	)
 	if cpuPercent > s.clusterResourceThreshold || memoryPercent > s.clusterResourceThreshold {
-		logger.Debugf("Cluster %s would exceed the cluster load threshold (%d%%): CPU=%d%%, Memory=%d%%", cluster.ID, s.clusterResourceThreshold, cpuPercent, memoryPercent)
-		return nil
+		if s.clusterResourceThresholdScaleValue != 0 &&
```
gigawhitlocks, line 356: I think we should flip this conditional so it's `if s.clusterResourceThresholdScaleValue == 0` and return early (the body of the `else` function) and then the body of the `if` can be outdented and the `else` block eliminated.

cmd/cloud/server.go
gigawhitlocks: Perfect, thanks @gabrieljackson

internal/supervisor/installation.go
gabrieljackson: Actually, the way the logging library works is the `err` is put in structured logging separate from the error string defined. For instance, this code

```
		logger.WithField("webhookURL", hook.URL).WithError(err).Error("Unable to send webhook")
```

produces this error

```
ERRO[2020-07-01T11:20:42-04:00] Unable to send webhook                        error="Post \"http://localhost:8065\": dial tcp [::1]:8065: connect: connection refused" installation=tgh8d8txijbeipqqnudiz3bw7a instance=8eu38nwszbripqjte6nbzd3q8a webhookEvent=stable webhookURL="http://localhost:8065"
```

PR #260, opened 2020-07-07
deploy metric-server

cpanato:
#### Summary
In order to use HPA and VPA, we need to deploy the metrics-server and for that, we need to set two options in the kubelet.

To do that we need to update the KOPS cluster to add two new parameters and then we can deploy the metric server.

This also enables us when troubleshooting a cluster we can get quick metrics using `kubectl top nodes`  or `kubectl top pods` 

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-26763

#### Release Note

```release-note
deploy metric-server
```


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -419,6 +453,20 @@ func (provisioner *KopsProvisioner) UpgradeCluster(cluster *model.Cluster) error
...
+	setValue := "spec.kubelet.authenticationTokenWebhook=true"
+	err = kops.SetCluster(kopsMetadata.Name, setValue)
+	if err != nil {
+		return err
+	}
+	setValue = "spec.kubelet.authorizationMode=Webhook"
+	err = kops.SetCluster(kopsMetadata.Name, setValue)
+	if err != nil {
+		return err
+	}
```
gabrieljackson, line 468: Should this part be done in the `Provision` process? It seems like something we don't need to actually upgrade the cluster for.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -260,6 +273,24 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
...
+		return errors.Wrap(err, "failed to delete service metrics-server")
+	}
+
+	logger.Info("Cleaning up some metrics-server resources to reapply")
+	err = k8sClient.KubeagClientSet.ApiregistrationV1beta1().APIServices().Delete("v1beta1.metrics.k8s.io", &metav1.DeleteOptions{})
+	if k8sErrors.IsNotFound(err) {
+		logger.Info("APIService v1beta1.metrics.k8s.io not found; skipping...")
+	} else if err != nil {
+		return errors.Wrap(err, "failed to delete APIService v1beta1.metrics.k8s.io")
+	}
```
gabrieljackson, line 292: We really shouldn't need to check this on create as the cluster is being created for the first time right? Doesn't hurt to have it, but just thinking that this is probably safe to leave out.

cpanato: only if we update the manifests

gabrieljackson: Okay, that makes sense!

cpanato: this function is called for create/upgrade and provision API calls

PR #261, opened 2020-07-07
Check ID field in webhook listener

gabrieljackson:
This prevents issues when non-cloud-webhooks are received by the
CWL.

```release-note
None
```


gigawhitlocks:
Took me a second to remember what the "CWL" was lol

PR #262, opened 2020-07-09
MM-26670 Add ability to look up an Installation by DNS name

gigawhitlocks:
#### Summary
Adds API endpoint GET `/api/search/name/{FQDN}`
This new endpoint uses GetInstallations with an InstallationFilter to fetch Installations using their FQDN, in order to satisfy MM-26670

Also I felt conflicted about where to actually put this API endpoint. I added it to a new `search` subrouter with the thought that in the future we may want other search methods, and because there was not a great way to add a `search` subrouter to the Installation routes directly since the existing ones assume you have the Installation ID and I think that's the right pattern.

So please provide feedback on that in particular if you have an opinion!

#### Ticket Link

[MM-26670](https://mattermost.atlassian.net/browse/MM-26670)
#### Release Note

```release-note
Adds new endpoint GET /api/search/name/{query} to allow fetching Installations by name
```

### DIFF
--- a/model/client.go
+++ b/model/client.go
@@ -355,31 +355,29 @@ func (c *Client) GetInstallation(installationID string, request *GetInstallation
 }
 
 // GetInstallationByName finds an installation with the given FQDN.
-func (c *Client) GetInstallationByName(installationDNS string, request *GetInstallationRequest) (*Installation, error) {
-	u, err := url.Parse(c.buildURL("/api/search/name/%s", installationDNS))
+func (c *Client) GetInstallationByName(DNS string, request *GetInstallationRequest) (*Installation, error) {
+	if request == nil {
+		request = &GetInstallationRequest{
+			IncludeGroupConfig:          false,
+			IncludeGroupConfigOverrides: false,
+		}
+	}
+	installations, err := c.GetInstallations(&GetInstallationsRequest{
+		IncludeGroupConfig:          request.IncludeGroupConfig,
+		IncludeGroupConfigOverrides: request.IncludeGroupConfigOverrides,
+		IncludeDeleted:              true,
+		PerPage:                     AllPerPage,
+		DNS:                         DNS,
+	})
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "problem getting installation")
 	}
 
-	request.ApplyToURL(u)
-
-	resp, err := c.doGet(u.String())
-	if err != nil {
-		return nil, err
+	if len(installations) != 1 {
+		return nil, errors.Errorf("received ambiguous response (%d Installations) when expecting only one",
+			len(installations))
 	}
-	defer closeBody(resp)
-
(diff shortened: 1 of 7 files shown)


model/installation.go
```diff
@@ -50,6 +50,7 @@ type InstallationFilter struct {
 	Page           int
 	PerPage        int
 	IncludeDeleted bool
+	Name           string
```
jwilander, line 53: Should we use `DNS` over `Name` to be more clear? Because this will be `somename.cloud.mattermost.com` and not just `somename`, right?

gigawhitlocks: Yeah that's fine with me. You're correct that it's the full FQDN.

gigawhitlocks:
Let me sort of work backwards through your thoughts.

>The new endpoint handler logic is very very similar to the installation list handler logic

I noticed this as well, but what I forgot was that we use the `Client` object directly in the Customer Portal, so in my head I was thinking that it needed to be an API endpoint, specifically, to facilitate for that communication. But the logic could definitely live in the Client instead, since all of the consumers of this API use the Client. I don't feel strongly about this either way, but I want to think a little bit more deeply about it before I make a decision. So this leads me to your first point 

>it sort of feels like it should be general search endpoint for multiple types of cloud resources, but I am not sure if/how we could really support that

Yeah, so, especially with the way I named the endpoint, I was definitely looking forward to the possibility of needing other query types like this one, but I'm seeing now how your first question regarding putting this logic in the client applies... `GetInstallations` needs to be modified to support the new filter field, which makes the new endpoint redundant..

alright well this comment is kind of flow-of-consciousness but I'm going to leave it all here for the sake of history and transparency and 

tl;dr I think it's best to put this logic in the Client, update GetInstallations to include the new filter field (which I should have done/noticed to begin with anyway), and eliminate the redundant new endpoint. The Portal can then use the new Client method directly. The next commit to this PR will be those changes.

gabrieljackson:
Your tl;dr sounds good to me and I appreciate your thoughts on perhaps eventually having some endpoints that are more query-oriented. I definitely could envision something more general being useful in certain situations. That said, as we already discussed, the `installation list` API is already sort of a "searcher" for installations so maybe we start by using that and then think through a new set of `search` endpoints if that doesn't work very well.

gigawhitlocks:
>That said, as we already discussed, the installation list API is already sort of a "searcher" for installations

Right, that's what I realized while I was typing out my response to you @gabrieljackson :)

model/client.go
```diff
@@ -354,6 +354,32 @@ func (c *Client) GetInstallation(installationID string, request *GetInstallation
 	}
 }
 
+// GetInstallationByName finds an installation with the given FQDN.
+func (c *Client) GetInstallationByName(DNS string, request *GetInstallationRequest) (*Installation, error) {
```
jwilander, line 358: Should we call this `GetInstallationByDNS` now?

PR #263, opened 2020-07-14
[MΜ 26623] Migrate to single NGINX deployment, enable autoscaling.

stylianosrigas:
### Summary
This PR introduces a couple of NGINX changes:
- Start using k8s official NGINX helm [chart](https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx).
- Enable autoscaling, set limits and requests for NGINX ingress controller.
- Migrate from using separate private and public NGINX deployments to a single deployment that creates both services (new feature of the Helm chart).
- Enable use of TLS certificates for internal Load Balancers (Prometheus traffic).

#### Steps to reprovision existing clusters and to migrate installation to use new NGINX:
- Reprovision the cluster by running ```cloud cluster provision --cluster <cluster_id> --nginx-version 2.11.0```.
- Check that new nginx deployed both internal and public Load Balancers (**nginx-ingress-nginx-controller-internal** and **nginx-ingress-nginx-controller**).
- Manually update Prometheus Route53 record to use the new private Load Balancer (**nginx-ingress-nginx-controller-internal**).
- Manually update cluster installations Route53 records one by one to use the new public Load balancer (**nginx-ingress-nginx-controller**).
    - Update clusterinstallation ingress annotation to use Nginx class **nginx-controller** instead of **nginx**.
    - Manually update network policy to target nginx instead of public-nginx.
- Confirm that all services are up and running.
- Delete old NGINX helm charts.
  - ```helm del --purge public-nginx```
  - ```helm del --purge private-nginx```

### Ticket Link
- https://mattermost.atlassian.net/browse/MM-26623
- https://mattermost.atlassian.net/browse/MM-26858

### Release Note
```release-note
- Start using k8s official NGINX helm [chart](https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx).
- Enable autoscaling, set limits and requests for NGINX ingress controller.
- Migrate from using separate private and public NGINX deployments to a single deployment that creates both services.
- Enable use of TLS certificates for internal Load Balancers (Prometheus traffic).

Folllow deprecation instructions in the README.md for cluster reprovisioning steps.
```


helm-charts/nginx_values.yaml
```diff
@@ -2,14 +2,30 @@
 ## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
 ##
 controller:
-  ingressClass: nginx
+  ingressClass: nginx-controller
+  image:
+    repository: us.gcr.io/k8s-artifacts-prod/ingress-nginx/controller
```
cpanato, line 7: please remove this if we want to have upgrades in the future, with this we hardcode the version/image and when upgrading the chart it will need a change here as well

internal/provisioner/nginx.go
```diff
@@ -30,11 +33,16 @@ func newNginxHandle(desiredVersion string, provisioner *KopsProvisioner, kops *k
 		return nil, errors.New("cannot create a connection to Nginx if the provisioner provided is nil")
 	}
 
+	if awsClient == nil {
+		return nil, errors.New("cannot create a connection to Prometheus if the awsClient provided is nil")
```
gabrieljackson, line 37:
```suggestion
		return nil, errors.New("cannot create a connection to Nginx if the awsClient provided is nil")
```

stylianosrigas: fixed thanks :)

helm-charts/nginx_values.yaml
stylianosrigas: I removed the image version and forgot this, thanks :)

PR #266, opened 2020-07-23
[MM-27201] Move k8s client out of the internal folder for inclusion elsewhere 

nickmisasi:
#### Summary
Move the package out of the internal folder so we can use it elsewhere. Updated references to it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27201

#### Release Note
```release-note
You can now include the Mattermost Kubernetes client in other Golang projects. 
Simply `go get github.com/mattermost/mattermost-cloud` and add the module import `github.com/mattermost/mattermost-cloud/k8s`
```


cpanato:
this is good to have a release note

/kind cleanup

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -19,10 +19,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
-	"github.com/mattermost/mattermost-cloud/internal/tools/k8s"
 	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
 	"github.com/mattermost/mattermost-cloud/internal/tools/terraform"
 	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
+	"github.com/mattermost/mattermost-cloud/k8s"
```
gigawhitlocks, line 25: Is the name `k8s` something that's going to wind up having a namespace conflict when we import it elsewhere? Would we prefer a project-unique name? 0/5 on this but I'd like to hear @gabrieljackson 's thoughts.

nickmisasi: From a quick search for /k8s in the mattermost-cloud repo, I don't think so? It seems most of the Kubernetes packages only use k8s in the domain name (k8s.io), not the package name. Will wait for @gabrieljackson

gabrieljackson: Good question. I think it's safe to use as is, but may cause us some issues in the future. That said, we can always change it then and can import it as `cloud-k8s` or something in other projects to attempt to make it clear. I am fine with this as is for now.

PR #267, opened 2020-07-23
Add Postgres single-tenant database option

gabrieljackson:
This new database option allows for creating installations that are
backed by PostgreSQL databases.

Some general reviewer notes:
 - Postgres 9.6 is the lowest aurora option available. When Mattermost 5.26 is out we will switch to Postgres 10.X as it will be supported
 - We will update the `aws-rds` name to `aws-rds-mysql` in the future to be less confusing. (Database migration required)
 - This database option will at the very least be something that developers can use in the test account, but will allow us to do further validation on if we want to make it our default option for the cloud offering.

Fixes https://mattermost.atlassian.net/browse/MM-27219

```release-note
Add Postgres single-tenant database option
```


grundleborg:
@gabrieljackson might be misunderstanding the PR description here, but I'd recommend just going straight for the latest version (11.7 IIRC) rather than using an ancient Postgres. We did all our load testing on version 11, and it performs a lot better than 9.6!

gabrieljackson:
Oh, good to know @grundleborg. I was following the info found here in regards to the version: https://docs.mattermost.com/administration/important-upgrade-notes.html?highlight=postgres

Regardless, there will be at least one more follow-up to this PR so I can update it then.

grundleborg:
Ah, that document is interestingly (and very conservatively) worded - it's more accurate that our *minimum* required versions are changing there. For Cloud purposes I'd definitely recommend not using anything prior to Postgres 11

gabrieljackson:
Thanks for the clarification. I was surprised to see that as well, and figured I must have misunderstood our postgres version story up till now. I will definitely make sure we get to a new version in the next PR after I can ensure everything works via some local tests.

agnivade:
Yes please, postgres 9.6 and 11 are vastly different. And the latter tests were focussed _solely_ using postgres 11.

PR #269, opened 2020-07-24
Use the AWS_REGION set in the config and not hardcoded ones

cpanato:
#### Summary
This PR change the hardcoded AWS Region to use the one we set in the config, if that is not set then we use a default one.
Also changed the RDS creation to get the availabilities zones available in the region and also not use hardcoded.

This also addresses the mock generation to not have the user path in the mocks, so we can add a check to validate if the mocks are out of sync

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-27183

#### Release Note

```release-note
Use the AWS_REGION set in the config and not hardcoded ones
```


internal/tools/aws/rds.go
```diff
@@ -93,12 +93,13 @@ func (a *Client) rdsEnsureDBClusterCreated(awsID, vpcID, username, password, kms
...
+		return err
+	}
+
 	input := &rds.CreateDBClusterInput{
-		AvailabilityZones: []*string{
-			aws.String("us-east-1a"),
-			aws.String("us-east-1b"),
-			aws.String("us-east-1c"),
-		},
+		AvailabilityZones:     azs,
```
cpanato, line 102: Here, I dont know if we want to cap to 2-3 or want to add all available

gabrieljackson:
Sorry, looks like the Postgres PR changes conflicted with this.

cpanato:
@gabrieljackson thanks :D will rebase later not today 😂

cpanato:
@gabrieljackson @gigawhitlocks PTAL

internal/tools/terraform/plan.go
```diff
@@ -27,11 +30,16 @@ func (c *Cmd) Init(remoteKey string) error {
...
+	awsRegion := os.Getenv("AWS_REGION")
+	if awsRegion == "" {
+		awsRegion = toolsAWS.DefaultAWSRegion
+	}
+
 	_, _, err = c.run(
 		"init",
 		arg("backend-config", fmt.Sprintf("bucket=%s", c.remoteStateBucket)),
 		arg("backend-config", fmt.Sprintf("key=%s/%s", remoteStateDirectory, remoteKey)),
-		arg("backend-config", "region=us-east-1"),
```
angeloskyratzakos, line 34: @cpanato maybe to leave that one at `us-east-1` as the `-backend-config=bucket=mattermost-kops-state-test` exists in us-east-1 or we can have a different approach on using different bucket on other regions

cpanato: i think we need to have a different bucket for other regions as well to not mess up

internal/tools/aws/rds.go
gabrieljackson: This should be fine, I would think.

internal/tools/aws/client_test.go
```diff
@@ -139,7 +139,7 @@ func NewAWSTestSuite(t *testing.T) *AWSTestSuite {
 		RDSParamGroupCluster: "mattermost-provisioner-rds-cluster-pg",
 		RDSParamGroup:        "mattermost-provisioner-rds-pg",
 		RDSClusterID:         "rds-cluster-multitenant-09d44077df9934f96-97670d43",
-		RDSAvailabilityZones: []string{"us-east-1a", "us-east-1b", "us-east-1c"},
+		RDSAvailabilityZones: []string{"us-honk-1a", "us-honk-1b"},
```
gigawhitlocks, line 142: What availability zone is this?

cpanato: /honk

PR #270, opened 2020-07-26
MM-27216 - added MM_CLOUD_INSTALLATION_ID

reflog:
#### Summary

Added env variable to mattermost-server envmap to identify the installation

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27216
#### Release Note

```release-note
Added MM_CLOUD_INSTALLATION_ID env variable to identify mattermost-server installations
```

### DIFF
--- a/internal/store/installation.go
+++ b/internal/store/installation.go
@@ -194,6 +194,8 @@ func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation) e
 	installation.ID = model.NewID()
 	installation.CreateAt = GetMillis()
 
+	// add installation ID to mattermost-server env variables
+	installation.MattermostEnv["MM_CLOUD_INSTALLATION_ID"] = model.EnvVar{Value: installation.ID}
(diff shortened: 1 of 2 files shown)


internal/store/installation.go
```diff
@@ -194,6 +194,11 @@ func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation) e
 	installation.ID = model.NewID()
 	installation.CreateAt = GetMillis()
 
+	// add installation ID to mattermost-server env variables
```
gabrieljackson, line 197: Although this implementation would definitely achieve the goal of getting an env var set with the installation ID, I think it is probably better suited as hardcoded logic when the installation is created in terms of the custom resource: https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/kops_provisioner_cluster_installation.go#L83

The reasoning is that would help protect against the value being accidentally removed or changed when compared to adding it like standard env configuration here.

cpanato:
I think this worth a release note

PR #271, opened 2020-07-27
Add database endpoint init check to single tenant databases

gabrieljackson:
This new database init check is used to prevent the main Mattermost
app container from starting until a basic database endpoint check
is successful.

To accomplish this, the Mattermost Operator has been updates to
version `1.6.0`.

Fixes https://mattermost.atlassian.net/browse/MM-26765

```release-note
Add database endpoint init check to single tenant databases
```

### DIFF
--- a/manifests/operator-manifests/mattermost/operator.yaml
+++ b/manifests/operator-manifests/mattermost/operator.yaml
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:v1.5.0
+          image: mattermost/mattermost-operator:v1.6.0
           imagePullPolicy: IfNotPresent
           command:
           - mattermost-operator
--- a/internal/tools/aws/database.go
+++ b/internal/tools/aws/database.go
@@ -143,18 +143,6 @@ func (d *RDSDatabase) GenerateDatabaseSpecAndSecret(store model.InstallationData
 		"database-type":   d.databaseType,
 	})
 
-	var connTemplate, databaseCheckPort string
-	switch d.databaseType {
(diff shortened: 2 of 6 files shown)


gigawhitlocks:
@gabrieljackson could you refer me to the relevant change in the Operator that corresponds to this? Just curious to have a look at that.

gabrieljackson:
Sure @gigawhitlocks! Here is the important bits in the original PR: https://github.com/mattermost/mattermost-operator/pull/156/files#diff-a0f752049cee581a201be0939de19047R329

manifests/operator-manifests/mattermost/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:v1.5.0
+          image: mattermost/mattermost-operator:v1.6.0
```
cpanato, line 18: @gabrieljackson did you double-check if the crds or any other manifest are up-to-date?

gabrieljackson: Good point! It worked just fine, but I should update them just in case. Thanks for the reminder.

cpanato: @gabrieljackson let me know when you double-check

gabrieljackson:
After being reminded to update the operator manifests, I have done that and have included the new k8s client logic for the new version of custom resources. I also ran into some strange database endpoint check issues with Postgres, so I just removed that for now and will do a follow-up PR to address it. Please re-review when you can @gigawhitlocks and @cpanato

gabrieljackson:
@cpanato, yes I did.

PR #273, opened 2020-07-29
[MM-27278] Adding ability to attach custom IAM Policy to Kops node IAM Instance Profiles

stylianosrigas:
#### Summary
With this PR a custom IAM Policy will be attached to the Instance Profile role that Kops creates for each of the node instances. This IAM policy will be attached either during the creation of a new cluster or the reprovisioning of an existing one.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27278

#### Release Note
```release-note
- Attach custom IAM policy to Kops node IAM Instance Profiles
```


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -26,9 +26,13 @@ import (
...
 
-// DefaultKubernetesVersion is the default value for a kubernetes cluster
-// version value.
-const DefaultKubernetesVersion = "0.0.0"
+const (
+	// DefaultKubernetesVersion is the default value for a kubernetes cluster
+	// version value.
+	DefaultKubernetesVersion = "0.0.0"
+	// CustomNodePolicyName is the name of the custom IAM policy that will be attached in Kops Instance Profile
+	CustomNodePolicyName = "cloud-provisioning-node-policy"
```
stylianosrigas, line 34: @gabrieljackson let me know if you think this makes more sense to add it in the /tools/aws/constants.go file.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -189,6 +193,12 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 	}
 	logger.WithField("name", kopsMetadata.Name).Info("Successfully updated storage class")
 
+	iamRole := fmt.Sprintf("nodes.%s-kops.k8s.local", cluster.ID)
```
gabrieljackson, line 196:
```suggestion
	iamRole := fmt.Sprintf("nodes.%s", kopsMetadata.Name)
```

The kops metadata name value is the full ID plus the `-kops.k8s.local` suffix. We should use that in case we ever decide to update how kops names are generated.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -403,6 +413,12 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 		}
 	}
 
+	iamRole := fmt.Sprintf("nodes.%s-kops.k8s.local", cluster.ID)
```
gabrieljackson, line 416: Same here.

internal/provisioner/kops_provisioner_cluster.go
gabrieljackson: I could see it making sense in either place, but it may fit a bit better there, so yeah I would say we might as well move it.

internal/provisioner/kops_provisioner_cluster.go
stylianosrigas: Awesome, changing that thanks.

internal/mocks/aws-sdk/sts.go
```diff
@@ -0,0 +1,440 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
cpanato, line 1: add the entry in the `doc.go` file similar we did here: https://github.com/mattermost/mattermost-cloud/blob/master/internal/mocks/aws-sdk/doc.go#L14-L15

cpanato: after you add just run `make mocks`  to update

stylianosrigas: thanks done

stylianosrigas:
@gabrieljackson @cpanato can you please have another look. I added the ability to detach the policy from the role before killing the cluster. because kops will error out when it tries to delete the Instance profile if the policy is still attached. I also moved the utility group cleanup in the beginning of the destroy cluster process as it was running after the cluster was deleted.

PR #278, opened 2020-07-31
[MM 27287] Add support for Teleport external storage and HA deployment.

stylianosrigas:
#### Summary
With this PR:
- Support is added for Teleport external storage
- Teleport DynamoDB table and S3 bucket are deleted during cluster tear down
- Teleport 0.3.0 is released

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27287

#### Release Note
```release-note
- Support is added for Teleport external storage
- Teleport DynamoDB table and S3 bucket are deleted during cluster tear down
- Teleport 0.3.0 is released
```


internal/provisioner/teleport.go
```diff
@@ -88,15 +91,35 @@ func (n *teleport) ActualVersion() string {
...
+		return errors.Wrap(err, "unable to delete Teleport dynamodb events table")
+	}
 	return nil
 }
 
 func (n *teleport) NewHelmDeployment() *helmDeployment {
+	awsRegion := os.Getenv("AWS_REGION")
+	if awsRegion == "" {
+		awsRegion = aws.DefaultAWSRegion
+	}
```
gabrieljackson, line 116: I am pretty sure that the awsClient passed in to `teleport` has the region set already. We should be able to just use that. https://github.com/mattermost/mattermost-cloud/blob/3006530cca13f09dc8552dec515710221ebfc79a/cmd/cloud/server.go#L165

internal/tools/aws/dynamodb.go
```diff
@@ -0,0 +1,36 @@
...
+			logger.WithField("dynamodb-table", tableName).Warn("DynamoDB table could not be found; assuming already deleted")
+			return nil
+		}
+	}
+
+	_, err = a.Service().dynamodb.DeleteTable(&dynamodb.DeleteTableInput{
+		TableName: aws.String(tableName),
+	})
+	if err != nil {
+		return errors.Wrap(err, "unable to delete bucket")
```
gabrieljackson, line 32:
```suggestion
		return errors.Wrap(err, "unable to DynamoDB table")
```

internal/provisioner/teleport.go
```diff
@@ -88,15 +91,35 @@ func (n *teleport) ActualVersion() string {
 }
 
 func (n *teleport) Destroy() error {
+	teleportClusterName := fmt.Sprintf("cloud-%s-%s", n.environment, n.cluster.ID)
+	err := n.awsClient.S3EnsureBucketDeleted(teleportClusterName, n.logger)
```
gabrieljackson, line 95: Where does this bucket come from? I took a quick look and didn't see anything related to creating this utility.

stylianosrigas: @gabrieljackson your question makes sense. Both the S3 bucket and the DynamoDB tables are automatically created by the Teleport helm chart deployment.

internal/provisioner/teleport.go
stylianosrigas: @gabrieljackson that's what I thought as well but I couldn't find any reference. Do you have any example?

stylianosrigas:
> How are upgrades handled with this? Can we go straight from the old version to this?

@gabrieljackson Yes we can go straight from the old to the new version. New clusters will automatically use version 0.3.0 and old clusters will be reprovisioned with requested version 0.3.0. The clusters have already the policy in place to handle DynamoDB and S3 and the Teleport helm chart will handle those.

internal/provisioner/teleport.go
gabrieljackson: Oh, that makes sense! Thanks.

I am guessing we can't have helm tear them down for us as well?

gabrieljackson:
Awesome, thanks for confirming.

internal/provisioner/teleport.go
gabrieljackson: After discussing this offline with @stylianosrigas, we will keep this as is and will do a follow-up to clean this up a bit at a later date.

internal/provisioner/teleport.go
stylianosrigas: Nope unfortunately it cannot. I guess the reason for that is that they don't want to delete any data we may want to preserve in an accidental helm chart deletion.

PR #279, opened 2020-08-03
k8s: if pod is not found continue the check and assume that was part of the old replicaset

cpanato:
#### Summary
During the reprovision on the k8s cluster we notice sometimes some deployments get stuck in the `WaitForPodRunning` because we apply all the manifests and right away start checking if the deployment was successful, but to check that we get a list of the pods running for a specific deployment.

In some cases, after we apply the manifest, the deployment can have the pods running for the old version and for the new version. And then, the provisioned get all of those, and we have the provision failed because some pods will get in a delete state and not running (since they are from an old version) and the function does not check that.

test in a cluster and reprovisioned several times and not errors any more

#### Release Note

```release-note
k8s: if pod is not found continue the check and assume that was part of the old replicaset
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -357,11 +357,11 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 			logger.Infof("Waiting up to %d seconds for %q pod %q to start...", wait, deployment, pod.GetName())
 			ctx, cancel := context.WithTimeout(context.Background(), time.Duration(wait)*time.Second)
 			defer cancel()
-			pod, err := k8sClient.WaitForPodRunning(ctx, namespace, pod.GetName())
+			_, err := k8sClient.WaitForPodRunning(ctx, namespace, pod.GetName())
 			if err != nil {
 				return err
 			}
-			logger.Infof("Successfully deployed operator pod %q", pod.Name)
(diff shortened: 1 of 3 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -357,11 +357,11 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 			logger.Infof("Waiting up to %d seconds for %q pod %q to start...", wait, deployment, pod.GetName())
 			ctx, cancel := context.WithTimeout(context.Background(), time.Duration(wait)*time.Second)
 			defer cancel()
-			pod, err := k8sClient.WaitForPodRunning(ctx, namespace, pod.GetName())
+			_, err := k8sClient.WaitForPodRunning(ctx, namespace, pod.GetName())
 			if err != nil {
 				return err
 			}
-			logger.Infof("Successfully deployed operator pod %q", pod.Name)
+			logger.Infof("Successfully deployed service pod %q", pod.Name)
```
stylianosrigas, line 364: Now that pod is removed from line 360, does pod.Name still get a value?

cpanato: the value comes from this `for _, pod := range pods.Items {`

stylianosrigas: It shouldn't be pod.GetName()?

cpanato: fixed

PR #280, opened 2020-08-03
fix region for S3 endpoint

angeloskyratzakos:
#### Summary
Fix region for S3 endpoint as it was using the default one `s3.amazonaws.com`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27474

#### Release Note

```release-note
NONE

```


angeloskyratzakos:
Thanks @gabrieljackson for the suggestion I have updated the code and tested it in E2E and it works

PR #281, opened 2020-08-04
Add cloud resource API security locking

gabrieljackson:
This introduces the concept of cloud resource API security locking
for clusters, installations, cluster installations and groups. When
a resource is locked, no changes can be made to the resource until
it is unlocked again.

Locking and unlocking occurs via a new set of cloud server API
endpoints under /security. By separating the ability to control
resource locks out from the normal endpoints, the security endpoints
can be blocked off from other consumers of the cloud server API.
In the future, these endponts could also be exposed on a separate
port from the normal API if desired.

When creating new cloud resources, it is now possible to create them
in a state where they are immediately locked from API changes. This
opens up the possibility for clients of the cloud server API to
create resources, but not modify or delete which offers greater
control over their basic permissions. This functionality should
be especially helpful when dealing with long-lived resources that
are not changed often.

Fixes https://mattermost.atlassian.net/browse/MM-25542

```release-note
Add cloud resource API security locking
```

### DIFF
--- a/cmd/cloud/security.go
+++ b/cmd/cloud/security.go
@@ -1,0 +1,220 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package main
+
+import (
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	securityCmd.PersistentFlags().String("server", "http://localhost:8075", "The provisioning server whose API will be queried.")
+
+	securityClusterCmd.PersistentFlags().String("cluster", "", "The id of the cluster.")
+	securityClusterCmd.MarkPersistentFlagRequired("cluster")
+
+	securityInstallationCmd.PersistentFlags().String("installation", "", "The id of the installation.")
+	securityInstallationCmd.MarkPersistentFlagRequired("installation")
+
+	securityClusterInstallationCmd.PersistentFlags().String("cluster-installation", "", "The id of the cluster installation.")
+	securityClusterInstallationCmd.MarkPersistentFlagRequired("cluster-installation")
+
+	securityGroupCmd.PersistentFlags().String("group", "", "The id of the group.")
+	securityGroupCmd.MarkPersistentFlagRequired("group")
+
+	securityCmd.AddCommand(securityClusterCmd)
+	securityClusterCmd.AddCommand(securityClusterLockAPICmd)
+	securityClusterCmd.AddCommand(securityClusterUnlockAPICmd)
+
+	securityCmd.AddCommand(securityInstallationCmd)
+	securityInstallationCmd.AddCommand(securityInstallationLockAPICmd)
+	securityInstallationCmd.AddCommand(securityInstallationUnlockAPICmd)
+
+	securityCmd.AddCommand(securityClusterInstallationCmd)
+	securityClusterInstallationCmd.AddCommand(securityClusterInstallationLockAPICmd)
+	securityClusterInstallationCmd.AddCommand(securityClusterInstallationUnlockAPICmd)
+
+	securityCmd.AddCommand(securityGroupCmd)
+	securityGroupCmd.AddCommand(securityGroupLockAPICmd)
+	securityGroupCmd.AddCommand(securityGroupUnlockAPICmd)
+}
+
+var securityCmd = &cobra.Command{
+	Use:   "security",
+	Short: "Manage security locks for different cloud resources.",
+}
+
+var securityClusterCmd = &cobra.Command{
+	Use:   "cluster",
+	Short: "Manage security locks for cluster resources.",
+}
+
+var securityClusterLockAPICmd = &cobra.Command{
+	Use:   "api-lock",
+	Short: "Lock API changes on a given cluster",
+	RunE: func(command *cobra.Command, args []string) error {
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		clusterID, _ := command.Flags().GetString("cluster")
+		err := client.LockAPIForCluster(clusterID)
+		if err != nil {
+			return errors.Wrap(err, "failed to lock cluster API")
+		}
+
+		return nil
+	},
+}
+
+var securityClusterUnlockAPICmd = &cobra.Command{
+	Use:   "api-unlock",
+	Short: "Unlock API changes on a given cluster",
+	RunE: func(command *cobra.Command, args []string) error {
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		clusterID, _ := command.Flags().GetString("cluster")
+		err := client.UnlockAPIForCluster(clusterID)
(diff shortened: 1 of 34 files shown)


cpanato:
to use that we just call the endpoint? there is no secret that we need to set together or use to unlock?

gabrieljackson:
Nope, no secret to set. That could be added later if we want to of course!

internal/api/security.go
```diff
@@ -0,0 +1,276 @@
...
+			return
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// handleClusterUnockAPI responds to POST /api/cluster/{cluster}/api/unlock,
+// unlocking API changes for this cluster.
+func handleClusterUnockAPI(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks, line 70: UnockAPI -> UnlockAPI

internal/api/security.go
```diff
@@ -0,0 +1,276 @@
...
+			return
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// handleInstallationUnockAPI responds to POST /api/installation/{installation}/api/unlock,
+// unlocking API changes for this installation.
+func handleInstallationUnockAPI(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks, line 130: Unock -> Unlock, again

internal/api/helpers.go
```diff
@@ -9,8 +9,13 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
+func logSecurityLockConflict(resourceType string, logger logrus.FieldLogger) {
+	logger.WithField("api-security-lock-conflict", resourceType).Warn("API security lock conflict detected")
```
gigawhitlocks, line 16: What are the implications of this? Does the APILock need to be protected with a mutex?

internal/api/security.go
```diff
@@ -0,0 +1,276 @@
...
+			return
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// handleClusterInstallationUnockAPI responds to POST /api/cluster_installation/{cluster_installation}/api/unlock,
+// unlocking API changes for this cluster installation.
+func handleClusterInstallationUnockAPI(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks, line 190: One more Unock

internal/api/security.go
```diff
@@ -0,0 +1,276 @@
...
+			return
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// handleGroupUnockAPI responds to POST /api/group/{group}/api/unlock,
+// unlocking API changes for this group.
+func handleGroupUnockAPI(c *Context, w http.ResponseWriter, r *http.Request) {
```
gigawhitlocks, line 250: Unock again

internal/api/security.go
gabrieljackson: Nice catch! Who knows what an Unock even is. ;)

internal/store/migrations.go
```diff
@@ -984,6 +984,236 @@ var migrations = []migration{
...
+				LockAcquiredBy,
+				LockAcquiredAt
+			FROM
+				ClusterTemp;
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = e.Exec(`DROP TABLE ClusterTemp;`)
```
gigawhitlocks, line 1044: I thought that both of our database engines supported using ALTER TABLE to add Columns. If this is the case, I would like to see us use that instead of copying data to a temporary table to add columns. These copies are going to really add up when there's a lot of data in this database and since the migrations represent the schema and are therefore a permanent fixture of the process, if we keep using full table copies to add columns the migration process will eventually get very lengthy.

In other words, couldn't lines 989 to 1044 be replaced with
```go
_, err := e.Exec(`ALTER TABLE Cluster ADD COLUMN ApiSecurityLock BOOLEAN NOT NULL DEFAULT 'false';`)
if err != nil {
    return err
}
```
?

model/client.go
```diff
@@ -787,3 +787,60 @@ func (c *Client) DeleteWebhook(webhookID string) error {
 		return errors.Errorf("failed with status code %d", resp.StatusCode)
 	}
 }
+
+// LockAPIForCluster locks API changes for a given cluster.
+func (c *Client) LockAPIForCluster(clusterID string) error {
+	return c.makeSecurityCall("cluster", clusterID, "api", "lock")
```
gigawhitlocks, line 793: Not a suggestion for right now but it occurs to me that it might behoove us to use constants for stuff like `"cluster"` and `"installation"` here and elsewhere in the repo

model/client.go
```diff
@@ -787,3 +787,60 @@ func (c *Client) DeleteWebhook(webhookID string) error {
...
+		return err
+	}
+	defer closeBody(resp)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return nil
+
+	default:
+		return errors.Errorf("failed with status code %d", resp.StatusCode)
```
gigawhitlocks, line 843: Having the output from `c.buildURL` as part of this error message may aid in debugging. 0/5.

internal/api/security.go
gabrieljackson: This pitfalls of copypasta!

internal/store/migrations.go
gigawhitlocks: (also, did we drop support for sqlite? I seem to remember it being discussed awhile back but I've been testing with a local Postgres instance for so long that now I can't recall if we actually dropped the support for SQLite or not. If we did drop support for it, I think Postgres lets you use `ALTER TABLE` like I've described)

internal/api/security.go
gigawhitlocks: it's what happens when you arrive at a friend's door

cmd/cloud/security.go
```diff
@@ -0,0 +1,220 @@
...
+package main
+
+import (
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	securityCmd.PersistentFlags().String("server", "http://localhost:8075", "The provisioning server whose API will be queried.")
```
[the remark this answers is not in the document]
gabrieljackson: Me too, so I added one!

internal/api/helpers.go
gabrieljackson: This is just a logger helper and is only called from an object that the API server has acquired a normal lock on, so it should be safe.

model/client.go
gabrieljackson: Yep! Probably worth doing.

internal/store/migrations.go
gabrieljackson: Nope! We didn't drop SQLite support (yet). It is still somewhat useful so I am dragging it along still, but maybe this should be where we stop supporting it.

Regardless, good point on simplifying the migration. The cluster table still is probably better suited doing the original flow as it also drops two columns in the process and splitting up that logic for mysql and PG seems to be a bit messy, but the other tables are now updated as you recommended.

model/client.go
gabrieljackson: Not a bad idea. We might as well do that for all client error returns if we think that is useful. Probably better as a separate PR.

PR #283, opened 2020-08-11
Add PostgreSQL Multitenant Database option

gabrieljackson:
This new database option allows for multiple installation databases
to be provisioned on a single Postgres RDS Aurora cluster. As part
of this new option, the following improvements were made.

 - Refactor multitenant database locking. Locks are now acquired on
   the root database management methods and are associated with the
   datastore object instead of the RDS cluster resources. This
   decreases the complexity associated with maintaining the locks.
 - Added new DatabaseType field to track which database type a DB
   cluster is providing.
 - Made the VpcID column NOT NULL as it should never be missing.
 - Installation IDs belonging to a multitenant database are now
   marshalled and unmarshalled on directly on the store level. This
   removes the necessity for other packages to manage this process
   and brings the behavior in line with other stored JSON blobs.
 - More tests and logging improvements.

Fixes https://mattermost.atlassian.net/browse/MM-27482 and https://mattermost.atlassian.net/browse/MM-27341

```release-note
Add PostgreSQL Multitenant Database option
```

### DIFF
--- a/internal/tools/aws/constants.go
+++ b/internal/tools/aws/constants.go
@@ -53,17 +53,36 @@ const (
 	// Note: This needs to be manually created before RDS databases can be used.
 	DefaultDBSubnetGroupName = "mattermost-databases"
 
+	// DatabaseTypeMySQLAurora is a MySQL database running on AWS RDS Aurora.
+	DatabaseTypeMySQLAurora = "MYSQL/Aurora"
+
+	// DatabaseTypePostgresSQLAurora is a PostgreSQL database running on AWS
+	// RDS Aurora.
+	DatabaseTypePostgresSQLAurora = "PostgreSQL/Aurora"
+
+	// CloudInstallationDatabaseTagKey is the common tag key for determing
+	// database type.
+	CloudInstallationDatabaseTagKey = "tag:MattermostCloudInstallationDatabase"
+
 	// DefaultDBSecurityGroupTagKey is the default DB security group tag key
 	// that is used to find security groups to use in configuration of the RDS
 	// database.
 	// Note: This needs to be manually created before RDS databases can be used.
 	DefaultDBSecurityGroupTagKey = "tag:MattermostCloudInstallationDatabase"
 
-	// DefaultDBSecurityGroupTagValue is the default DB security group tag value
-	// that is used to find security groups to use in configuration of the RDS
-	// database.
-	// Note: This needs to be manually created before RDS databases can be used.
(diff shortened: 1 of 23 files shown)


internal/tools/aws/constants.go
```diff
@@ -53,17 +53,36 @@ const (
...
+
+	// CloudInstallationDatabaseTagKey is the common tag key for determing
+	// database type.
+	CloudInstallationDatabaseTagKey = "tag:MattermostCloudInstallationDatabase"
+
 	// DefaultDBSecurityGroupTagKey is the default DB security group tag key
 	// that is used to find security groups to use in configuration of the RDS
 	// database.
 	// Note: This needs to be manually created before RDS databases can be used.
 	DefaultDBSecurityGroupTagKey = "tag:MattermostCloudInstallationDatabase"
```
angeloskyratzakos, line 71: those 2 tags are same, do we want to use one of them, or it will cause problems down the line?

internal/tools/aws/database_multitenant.go
```diff
@@ -779,36 +857,73 @@ func (d *RDSMultitenantDatabase) connectRDSCluster(schema, endpoint, username, p
...
+	if d.databaseType == model.DatabaseEngineTypeMySQL {
+		// Query placeholders don't seem to work with argument database.
+		// See https://github.com/mattermost/mattermost-cloud/pull/209#discussion_r422533477
+		query := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s CHARACTER SET ?", databaseName)
+		_, err := d.db.QueryContext(ctx, query, "utf8mb4")
+		if err != nil {
+			return errors.Wrap(err, "failed to run create database SQL command")
+		}
+	} else {
+		query := fmt.Sprintf("DROP DATABASE IF EXISTS %s", databaseName)
```
jwilander, line 870: Do we really want to drop the database if it exists? Wouldn't we just want to check if it exists and if it does then we don't do anything?

Is there any chance of hitting this on a database that has been used for a while?

gabrieljackson: Good point. There shouldn't be a chance that we hit another database, but let's be safe and just check if it already exists instead.

internal/tools/aws/constants.go
gabrieljackson: Yeah, we have a couple duplicate tags already. Instead of cleaning them up here, I was going to do a follow-up ticket to take care of them.

PR #286, opened 2020-08-11
Add API to list multitenant databases

gabrieljackson:
This new endpoint allows the client to request information on all
currently-known multitenant databases. Note that other RDS clusters
could exist that the provisioner has not yet found and added to the
store.

Fixes https://mattermost.atlassian.net/browse/MM-27691

```release-note
Add API to list multitenant databases
```

### DIFF
--- a/internal/api/databases.go
+++ b/internal/api/databases.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package api
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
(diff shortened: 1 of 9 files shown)


internal/api/databases.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	databases, err := c.Store.GetMultitenantDatabases(filter)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to query multitenant databases")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if databases == nil {
+		databases = []*model.MultitenantDatabase{}
```
gigawhitlocks, line 52: this is a real 0/5 nitpick so feel free to ignore me, but the conditional here could be eliminated by just instantiating `databases` like this on line 44, before the call to `GetMultitenantDatabases`

PR #287, opened 2020-08-18
k8s/namespace: when deleting the namespace make sure everything gets deleted as well

cpanato:
#### Summary
Just to make sure we delete everything inside the namespace before removing the namespace, to avoid orphans hanging around in the cluster

#### Ticket Link
n/a

#### Release Note

```release-note
k8s/namespace: when deleting the namespace make sure everything gets deleted as well
```

### DIFF
--- a/k8s/namespaces.go
+++ b/k8s/namespaces.go
@@ -72,8 +72,14 @@ func (kc *KubeClient) GetNamespaces(namespaceNames []string) ([]*corev1.Namespac
 // Any errors will be returned immediately and the remaining namespaces will be
 // skipped.
 func (kc *KubeClient) DeleteNamespaces(namespaceNames []string) error {
+	policy := metav1.DeletePropagationForeground
+	gracePeriod := int64(45)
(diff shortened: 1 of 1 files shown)


k8s/namespaces.go
```diff
@@ -72,8 +72,14 @@ func (kc *KubeClient) GetNamespaces(namespaceNames []string) ([]*corev1.Namespac
 // Any errors will be returned immediately and the remaining namespaces will be
 // skipped.
 func (kc *KubeClient) DeleteNamespaces(namespaceNames []string) error {
+	policy := metav1.DeletePropagationForeground
+	gracePeriod := int64(0)
```
gabrieljackson, line 76: Do we maybe want a small grace period? Would this lead to a lot of containers not being able to shut down properly? Not sure if that is an issue, just wondering though.

cpanato: sounds good, will use 45 seconds

cpanato: done

PR #289, opened 2020-08-19
Log installation size resource requirements

gabrieljackson:
The scheduling algorithm for new installations calculates resource
requirements based on the installation size. It then compares this
requirement to a threshold and logs if it exceeds the threshold.

This change adds additional logging information on the raw resource
requirement of the installation in this case to provide better
debugging.

Fixes https://mattermost.atlassian.net/browse/MM-27970

```release-note
Log installation size resource requirements
```


ethervoid (approved):
LGTM! Nice :)

PR #291, opened 2020-08-20
GetInstallationsByDNS is returning error if length is 0

ethervoid:
#### Summary

This PR just check for ambiguous response if the number of returned installations is greater than 0 otherwise we're always receiving the error `error="received ambiguous response (0 Installations) when expecting only one"`

#### Release Note

```release-note
none
```


gigawhitlocks:
I believe there needs to be a check also to make sure len(Installations) is not zero

ethervoid:
@gigawhitlocks yeah I missed that one. Now is ready :) (sorry for the noise)

PR #292, opened 2020-08-21
Installations count endpoint

ethervoid:
#### Summary

This PR adds a new endpoint to get the number of installations either all or only
the active ones

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27626

#### Release Note

```release-note
Added endpoint that returns the number of installations: /api/installations/count
```

### DIFF
--- a/internal/store/installation.go
+++ b/internal/store/installation.go
@@ -147,7 +147,7 @@ func (sqlStore *SQLStore) GetInstallations(filter *model.InstallationFilter, inc
 // GetInstallationsCount returns the number of installations filtered by the deletedat
 // field
 func (sqlStore *SQLStore) GetInstallationsCount(includeDeleted bool) (int, error) {
-	builder := sq.Select("COUNT(*) as InstallationsNumber").From("Installation")
+	builder := sq.Select("COUNT(*) as InstallationsCount").From("Installation")
 	if !includeDeleted {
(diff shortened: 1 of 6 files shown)


model/installation.go
```diff
@@ -44,6 +44,11 @@ type Installation struct {
 	configMergeGroupSequence int64
 }
 
+// InstallationsCount represents the number of installations
+type InstallationsCount struct {
+	Count int `json:"installations_count"`
```
gabrieljackson, line 49: I would remove the JSON tag here and simply return `Count` to stay similar to the other model structs.

internal/store/installation.go
```diff
@@ -144,6 +144,25 @@ func (sqlStore *SQLStore) GetInstallations(filter *model.InstallationFilter, inc
 	return installations, nil
 }
 
+// GetInstallationsCount returns the number of installations filtered by the deletedat
+// field
+func (sqlStore *SQLStore) GetInstallationsCount(includeDeleted bool) (int, error) {
+	builder := sq.Select("COUNT(*) as InstallationsNumber").From("Installation")
```
gabrieljackson, line 150: nit: maybe this instead?
```suggestion
	builder := sq.Select("COUNT(*) as InstallationsCount").From("Installation")
```

PR #293, opened 2020-08-23
Enable default AWS S3 SSE

gabrieljackson:
This sets the default encryption scheme for new installation S3
buckets to use AES-256 encryption. The Mattermost filestore SSE
configuration option is also set to true to force the behavior
when a license is present.

This also corrects an issue where the cloud installation ID env
var was lost when updating installations.

Fixes https://mattermost.atlassian.net/browse/MM-24783

```release-note
Enable default AWS S3 SSE
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -63,11 +63,7 @@ func (provisioner *KopsProvisioner) CreateClusterInstallation(cluster *model.Clu
 		return errors.Wrapf(err, "failed to create network policy %s", clusterInstallation.Namespace)
 	}
 
-	// add installation ID to mattermost-server env variables
-	if installation.MattermostEnv == nil {
-		installation.MattermostEnv = map[string]model.EnvVar{}
-	}
-	installation.MattermostEnv["MM_CLOUD_INSTALLATION_ID"] = model.EnvVar{Value: installation.ID}
+	mattermostEnv := getMattermostEnvWithOverrides(installation)
 
 	mattermostInstallation := &mmv1alpha1.ClusterInstallation{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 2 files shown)


internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -520,3 +517,22 @@ func (provisioner *KopsProvisioner) execCLI(cluster *model.Cluster, clusterInsta
...
+// or updated.
+func getMattermostEnvWithOverrides(installation *model.Installation) model.EnvVarMap {
+	mattermostEnv := installation.MattermostEnv
+	if mattermostEnv == nil {
+		mattermostEnv = map[string]model.EnvVar{}
+	}
+
+	mattermostEnv["MM_CLOUD_INSTALLATION_ID"] = model.EnvVar{Value: installation.ID}
+
+	if !installation.InternalFilestore() {
```
cpanato, line 533: minio supports the same encryption as AWS

gabrieljackson: I tried enabling encryption with Minio and the Mattermost server started throwing errors and I couldn't connect. Is there maybe something that needs to be set on the Minio side?

cpanato: yes, you need to configure minio to accept that as well, but I don't know if the operator supports that, need to check the options

gabrieljackson: Okay, I figured that may be the case. Either way, I propose that we merge this as is and update it later if we can add the operator option to support it.

internal/tools/aws/s3.go
```diff
@@ -35,6 +35,22 @@ func (a *Client) s3EnsureBucketCreated(bucketName string, logger log.FieldLogger
 		return errors.Wrap(err, "unable to block public bucket access")
 	}
 
+	_, err = a.Service().s3.PutBucketEncryption(&s3.PutBucketEncryptionInput{
```
cpanato, line 38: we might need to add this permission in the provisioner user
```
s3:PutEncryptionConfiguration
```

cpanato: I will check if we need

for now lets /hold this and not merge

gabrieljackson: Thank you for checking that!

cpanato: need to merge and update the envs first: https://github.com/mattermost/mattermost-cloud-monitoring/pull/161

cpanato (approved):

gabrieljackson:
Just talked to @cpanato and the required permissions are now deployed. Please review when you have time @gigawhitlocks.

PR #294, opened 2020-08-25
drop requests to prometheus ingress if is not using the dns name

cpanato:
### Summary
If anyone tries to access the ingresses using the IP of the LB we can drop the request and only accept if is coming from the dns name.

#### Release Note

```release-note
drop requests to prometheus ingress if is not using the dns name
```


stylianosrigas (approved):
LGTM 👍

PR #295, opened 2020-08-26
Add dry-run flag for installation and group CLI commands

gabrieljackson:
This flag behaves in the same way as the cluster commands: the
request API payload is printed for review without sending it to
the API server.

Note that this only applies to commands with complex request
payloads.

Fixes https://mattermost.atlassian.net/browse/MM-28128

```release-note
Add dry-run flag for installation and group CLI commands
```


stylianosrigas (approved):
LGTM 🎉

PR #297, opened 2020-08-27
Fixing ctx timing to make sure that Nginx load balancer services are in place for Prometheus Route53 registration.

stylianosrigas:
#### Summary
Fixing the context for private load balancer service listing. This helps to ensure that Nginx Load Balancer services are in place before attempting to register Prometheus with Route53 private record.

#### Release Note
```release-note
- Fixing context for private load balancer service listing
```


cpanato (approved):

cpanato:
master is broken when creating new clusters

/priority critical-urgent

PR #299, opened 2020-09-02
Add new container exec API endpoint

gabrieljackson:
This new endpoint initially allows for running mmctl commands in
cluster installations. In the future, more commands can be approved
for use and can be managed under this endpoint.

The original mattermost-cli endpoint remains to not introduce a
breaking change and will be refactored later.

Fixes https://mattermost.atlassian.net/browse/MM-28239

```release-note
Add new container exec API endpoint
```

### DIFF
--- a/internal/api/cluster_installation.go
+++ b/internal/api/cluster_installation.go
@@ -5,6 +5,7 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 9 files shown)


internal/api/cluster_installation.go
```diff
@@ -24,6 +25,7 @@ func initClusterInstallation(apiRouter *mux.Router, context *Context) {
 	clusterInstallationRouter.Handle("", addContext(handleGetClusterInstallation)).Methods("GET")
 	clusterInstallationRouter.Handle("/config", addContext(handleGetClusterInstallationConfig)).Methods("GET")
 	clusterInstallationRouter.Handle("/config", addContext(handleSetClusterInstallationConfig)).Methods("PUT")
+	clusterInstallationRouter.Handle("/exec/{command}", addContext(handleRunClusterInstallationExecCommand)).Methods("POST")
 	clusterInstallationRouter.Handle("/mattermost_cli", addContext(handleRunClusterInstallationMattermostCLI)).Methods("POST")
```
jwilander, line 29: Goal is to eventually migrate this to be under `/exec/{command}` too, right?

gabrieljackson: Yes, exactly, just didn't want to break the old endpoint yet.

stylianosrigas (approved):
This is great, thanks @gabrieljackson

PR #300, opened 2020-09-02
add white-list cidrs to internal ingress for prometheus

cpanato:
#### Summary
Block access from public lb to an internal endpoint.

For dev we are allowing all. For other envs we need to set the flag `allow-list-cidr-range` to block the public access

follow up PR to add the flags in test/stg and prod envs:
- TEST: https://gitlab.internal.core.cloud.mattermost.com/cloud/cloud-private/-/merge_requests/140
- STG: https://gitlab.internal.core.cloud.mattermost.com/cloud/cloud-private/-/merge_requests/141
- PROD: https://gitlab.internal.core.cloud.mattermost.com/cloud/cloud-private/-/merge_requests/142

for STG / PROD we need to have the tag and deploy in those envs before merging the gitlab prs.

#### Release Note

```release-note
add white-list cidrs to internal ingress for prometheus
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -80,7 +80,7 @@ var serverCmd = &cobra.Command{
 
 		allowListCIDRRange, _ := command.Flags().GetStringSlice("allow-list-cidr-range")
 		if len(allowListCIDRRange) == 0 {
-			return errors.Error("allow-list-cidr-range must have at least one value")
+			return errors.New("allow-list-cidr-range must have at least one value")
 		}
 
 		logger := logger.WithField("instance", instanceID)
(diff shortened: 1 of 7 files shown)


stylianosrigas:
I did some small changes after testing with multiple clusters. It should be good to go now, @gabrieljackson you can have a review.

stylianosrigas (approved):

cmd/cloud/server.go
```diff
@@ -76,6 +78,11 @@ var serverCmd = &cobra.Command{
 			logger.SetFormatter(&logrus.JSONFormatter{})
 		}
 
+		allowListCIDRRange, _ := command.Flags().GetStringSlice("allow-list-cidr-range")
+		if len(allowListCIDRRange) == 0 {
+			return errors.Errorf("allow-list-cidr-range must have at least one value")
```
gabrieljackson, line 83: nit:
```suggestion
			return errors.Error("allow-list-cidr-range must have at least one value")
```

gabrieljackson (approved):
Looks good! Just one nit.

PR #306, opened 2020-09-16
Add S3 Multitenant Filestore

gabrieljackson:
This new filestore option allows multiple installations to share
a single S3 bucket. When used, IAM bucket policies are created to
limit installation access to a single subdirectory of the bucket
and the Mattermost server configuration is set to use this
subdirectory.

Other changes:
 - Move AWS environment lookup logic to the AWS package.
 - Changed AWS debug logging to use more log fields.

Fixes https://mattermost.atlassian.net/browse/MM-28449

```release-note
Add S3 Multitenant Filestore
```

### DIFF
--- a/internal/tools/aws/client.go
+++ b/internal/tools/aws/client.go
@@ -5,6 +5,7 @@
 package aws
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -32,12 +33,16 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
 	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
 
 // AWS interface for use by other packages.
 type AWS interface {
 	GetCertificateSummaryByTag(key, value string, logger log.FieldLogger) (*acm.CertificateSummary, error)
+
+	GetAccountAliases() (*iam.ListAccountAliasesOutput, error)
+	GetCloudEnvironmentName() (string, error)
 
 	GetAndClaimVpcResources(clusterID, owner string, logger log.FieldLogger) (ClusterResources, error)
 	ReleaseVpc(clusterID string, logger log.FieldLogger) error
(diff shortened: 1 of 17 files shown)


angeloskyratzakos (approved):
LGTM

internal/tools/aws/filestore_multitenant.go
```diff
@@ -0,0 +1,238 @@
...
+		},
+		StringData: map[string]string{
+			"accesskey": iamAccessKey.ID,
+			"secretkey": iamAccessKey.Secret,
+		},
+	}
+
+	S3RegionURL := S3URL
+	awsRegion := *f.awsClient.config.Region
+	if awsRegion != "" && awsRegion != "us-east-1" {
```
gigawhitlocks, line 124: Why the exception for `us-east-1`? Is `s3.us-east-1.amazonaws.com` invalid or something?

internal/tools/aws/iam.go
```diff
@@ -21,12 +21,15 @@ type policyDocument struct {
...
-	Action   []string
-	Resource string
+	Sid       string
+	Effect    string
+	Action    []string
+	Resource  string
+	Condition policyStatementCondition `json:"Condition,omitempty"`
 }
 
+type policyStatementCondition map[string]map[string][]string
```
gigawhitlocks, line 31: `map[string]map[string][]string` hurts me. This type definition would be my suggestion for making it less painful, so, I guess you've already done it. But wow what a type signature. :grimacing:

gigawhitlocks (approved):
Just a couple small questions and comments but this looks great overall @gabrieljackson .

internal/tools/aws/iam.go
gabrieljackson: Yeah, it's a mess and I totally agree. Strangely, the policy doc struct I have here is not something that AWS seems to provide in the SDK. Their own example does the same thing in defining it and then parsing it to JSON. Anyway, yeah, maybe something to improve in the future.

internal/tools/aws/client.go
```diff
@@ -142,3 +146,30 @@ func (c *Client) AddSQLStore(store model.InstallationDatabaseStoreInterface) {
...
+	accountAliases, err := c.GetAccountAliases()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get account aliases")
+	}
+	if len(accountAliases.AccountAliases) < 1 {
+		return "", errors.New("account alias not defined")
+	}
+
+	for _, alias := range accountAliases.AccountAliases {
+		if strings.HasPrefix(*alias, "mattermost-cloud") && len(strings.Split(*alias, "-")) == 3 {
```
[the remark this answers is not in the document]
gabrieljackson: No, shouldn't be `nil` because we are ranging through and so we should be able to trust that the AWS API didn't do that to us.

internal/tools/aws/client.go
```diff
@@ -142,3 +146,30 @@ func (c *Client) AddSQLStore(store model.InstallationDatabaseStoreInterface) {
...
+			envName := strings.Split(*alias, "-")[2]
```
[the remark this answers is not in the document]
gabrieljackson: Yeah, this logic was largely just moved from something in the teleport utility to the aws package. I will make note of that for a future improvement.

PR #308, opened 2020-09-21
MM-27476 Improve DNS hostname validation

gigawhitlocks:
#### Summary

Adds some DNS validation at creation time:
1) Makes sure the provided name is 3 to 64 characters long
2) Checks that the length is within Installation DNS name uniqueness is already enforced at the store layer when trying to store the name, so this is _not_ checked.
2) Checks that the hostname only has the allowed characters as specified in the RFC defining the allowable patterns in hostnames.
3) Tries to resolve the DNS name and fails if it is taken (if it already resolves to something).

I'm happy to add more checks if anything else comes to mind.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27476
#### Release Note

```release-note
Added some extra DNS hostname validation
```


model/installation_request.go
```diff
@@ -90,6 +94,32 @@ func (request *CreateInstallationRequest) Validate() error {
 	return checkSpaces(request)
 }
 
+func isValidDNS(dns string) error {
+	subdomain := strings.SplitN(dns, ".", 2)[0]
+	if len(subdomain) >= 64 || len(subdomain) < 3 {
+		return errors.Errorf("DNS names must be between 3 and 64 characters, but name was %d long. DNS=%s", len(dns), dns)
+	}
```
gabrieljackson, line 101: I believe we need to check the full DNS length here like we did before. That max character count has to apply to the full domain name.

model/installation_request.go
```diff
@@ -90,6 +94,32 @@ func (request *CreateInstallationRequest) Validate() error {
...
+	r := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			return (&net.Dialer{
+				Timeout: time.Second * time.Duration(10000),
+			}).DialContext(ctx, "udp", "1.1.1.1:53")
+		},
+	}
+	_, err := r.LookupHost(context.Background(), dns)
+	if err == nil {
```
gabrieljackson, line 117: So, I guess it is possible that the provided domain exists, but this lookup encounters some other type of error which results in the domain-taken check passing.

This is probably pretty unlikely to happen, but wanted to note it. Can we check the error type perhaps to also fail when it was a network-related error?

model/installation_request_test.go
```diff
@@ -49,12 +49,52 @@ func TestCreateInstallationRequestValid(t *testing.T) {
...
+				DNS:     "joram&gabe.com",
+			},
+		},
+		{
+			"DNS is a real FQDN that is in use already",
+			true,
+			&model.CreateInstallationRequest{
+				OwnerID: "owner1",
+				DNS:     "google.com",
+			},
```
gabrieljackson, line 90: Off-topic comment: if this test ever fails then the internet has gotten very strange. ;)

gabrieljackson (changes requested):

model/installation_request.go
gigawhitlocks: @gabrieljackson is this limitation an internal limitation? According to the RFC:
```
  Each element of the hostname must be from 1 to 63 characters long and
       the entire hostname, including the dots, can be at most 253
       characters long.  Valid characters for hostnames are ASCII(7) letters
       from a to z, the digits from 0 to 9, and the hyphen (-).  A hostname
       may not start with a hyphen.
```

gigawhitlocks: The RFC does make it seem like I do need to add a check to make sure the whole DNS name is <254 characters long, though.

model/installation_request.go
gigawhitlocks: Let me see what I can do

gigawhitlocks: @gabrieljackson, looking at the Go source code, it looks like the only error type that `net.LookupHost` produces is `net.DNSError`.

Here is an example of an error from looking up a nonexistent domain:
```
gore> net.LookupHost("foos.party")
[]string{}
&net.DNSError{
  Err:         "no such host",
  Name:        "foos.party",
  Server:      "10.247.0.2:53",
  IsTimeout:   false,
  IsTemporary: false,
  IsNotFound:  true,
}
```
(off topic, `gore` is a pretty neat Go REPL if you hadn't seen that before)

Here's an example of an error caused by network connectivity (`foo.party` is a registered domain):
```
gore> net.LookupHost("foo.party")
[]string{}
&net.DNSError{
  Err:         "read udp [::1]:60216->[::1]:53: read: connection refused",
  Name:        "foo.party",
  Server:      "[::1]:53",
  IsTimeout:   false,
  IsTemporary: true,
  IsNotFound:  false,
}
```

So, we can't act on the type of error, but I could provide a different error if `IsNotFound` is `false` and either `IsTimeout` or `IsTemporary` is `true`. How does that sound?

Szymongib (commented):
👍  Nothing more to add ;)
+1 for returning a different error when not `IsNotFound`

model/installation_request.go
gabrieljackson: That sounds great!

model/installation_request.go
gabrieljackson: Okay, my recollection of the 64 limit came from what Carlos explained to me when it was first added. If we can have a subdomain with 64 characters then that is awesome!

The 254 character total length check probably won't hurt to have, but it's highly unlikely to occur given we control the rest of the domain name here. That said, I will leave it up to you to implement or not.

model/installation_request.go
gigawhitlocks: Screenshot from 2020-09-22 13-11-12
it'd be a real PITA to write a test for this case, so here's a screenshot of me running the existing tests but with my network disconnected :stuck_out_tongue_closed_eyes: 

If y'all think this warrants a proper test I can figure out how to simulate a network disconnect to add a test case but it sounds like a fair amount of work with little gain.

gabrieljackson: Haha, this is more than fine for now. Good work.

gabrieljackson (approved):
👍

gigawhitlocks:
Still need 2 approvals to merge, right? This shouldn't need any review from the SREs so I re-added Szymon so he can press the approve button before I merge. If I'm wrong about this and you're reading this and your name is Gabe or Joram feel free to merge.

Szymongib (approved):

PR #309, opened 2020-09-23
Add table output option for resource lists

gabrieljackson:
The new table option allows for listing resources in a more compact
and readable way. Not all resource fields are shown so the standard
JSON output should be used when viewing all details is required.

Fixes https://mattermost.atlassian.net/browse/MM-28986

```release-note
Add table output option for resource lists
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -6,9 +6,11 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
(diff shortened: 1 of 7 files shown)


gigawhitlocks (approved):
Awesome! This'll be nice for when I want a nice overview and don't feel like fiddling with `jq`

gabrieljackson:
This is a small client-only change so going to merge it in with one review to keep things moving along.

PR #310, opened 2020-09-23
MM-16248 Add a health check to Provisioner start-up

gigawhitlocks:
#### Summary

This change adds a few health checks to Provisioner startup to end early if basic necessary components are not in place.

1) Checks for the following binaries installed on the system:
    - terraform
    - kubectl
    - kops
    - helm
2) Checks that there is at least one SSH key installed in `~/.ssh`.

    One limitation of this is that AWS doesn't support all SSH key types, so e.g. if you have an ed25519 key, this check will pass but then later operations may fail. I didn't want to bake AWS's limitation surrounding this into our codebase, however, as hopefully they will support ed25519 keys and I am not sure (and couldn't find documentation describing) which other key types they _do_ support, other than RSA.

3) Establishes a connection with AWS and does `GetCloudEnvironmentName()` in order to confirm that the connection to AWS works -- thereby confirming that the user has set up AWS credentials or environment variables.

Am I missing anything else?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-16248

#### Release Note

```release-note
Added a health check to Provisioner startup to ensure the basic requirements are met.
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -182,7 +182,7 @@ var serverCmd = &cobra.Command{
 		}
 		awsClient := toolsAWS.NewAWSClientWithConfig(awsConfig, logger)
 
-		err = healthCheck(awsConfig, s3StateStore)
+		err = checkRequirements(awsConfig, s3StateStore)
 		if err != nil {
 			return errors.Wrap(err, "failed health check")
 		}
@@ -277,7 +277,7 @@ var serverCmd = &cobra.Command{
 	},
 }
 
-func healthCheck(awsConfig *sdkAWS.Config, s3StateStore string) error {
+func checkRequirements(awsConfig *sdkAWS.Config, s3StateStore string) error {
 	for _, requiredUtility := range []string{
 		"terraform",
 		"helm",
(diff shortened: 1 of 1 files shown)


cmd/cloud/server.go
```diff
@@ -272,6 +277,63 @@ var serverCmd = &cobra.Command{
 	},
 }
 
+func healthCheck(awsConfig *sdkAWS.Config, s3StateStore string) error {
```
Szymongib, line 280: Not gonna argue about the semantics, but maybe something like `checkRequirements` would fit this function a little better, what do you think?

Szymongib (approved):

cmd/cloud/server.go
gigawhitlocks: I agree, actually! I didn't put much thought into the name, as I was within the context of the ticket.

cpanato (approved):

gabrieljackson (approved):
This is great!

It was outside the scope of this ticket, but in the future we will want to add an AWS tag test suite to check that all the basic resources like VPCs and DB clusters are available in the account. I will make sure we have a ticket for that.

PR #312, opened 2020-09-29
MM-27129 - Add group status endpoint

Szymongib:
#### Summary

Changes in the PR:
- Add initial version of group status endpoint to API
  - Endpoint path: `api/group/[GROUP_ID]/status`
  - Response data:
    - `InstallationsTotal`
    - `InstallationsUpdated`
    - `InstallationsAwaitingUpdate`
- Add new command to CLI, which targets the new endpoint
  ```
  cloud group status --group=[GROUP_ID]
  ```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27129

#### Release Note

```release-note
Add endpoint returning the Group status
```

### DIFF
--- a/cmd/cloud/group.go
+++ b/cmd/cloud/group.go
@@ -267,7 +267,7 @@ var groupListCmd = &cobra.Command{
 
 var groupGetStatusCmd = &cobra.Command{
 	Use:   "status",
-	Short: "Get a particular groups status.",
+	Short: "Get a particular group's status.",
 	RunE: func(command *cobra.Command, args []string) error {
 		command.SilenceUsage = true
 
--- a/cmd/cloud/group.go
+++ b/cmd/cloud/group.go
@@ -47,6 +47,9 @@ func init() {
 	groupListCmd.Flags().Bool("include-deleted", false, "Whether to include deleted groups.")
 	groupListCmd.Flags().Bool("table", false, "Whether to display the returned group list in a table or not")
 
+	groupGetStatusCmd.Flags().String("group", "", "The id of the group of which the status should be fetched.")
+	groupGetStatusCmd.MarkFlagRequired("group")
+
 	groupJoinCmd.Flags().String("group", "", "The id of the group to which the installation will be added.")
 	groupJoinCmd.Flags().String("installation", "", "The id of the installation to add to the group.")
 	groupJoinCmd.MarkFlagRequired("group")
@@ -61,6 +64,7 @@ func init() {
 	groupCmd.AddCommand(groupDeleteCmd)
 	groupCmd.AddCommand(groupGetCmd)
 	groupCmd.AddCommand(groupListCmd)
+	groupCmd.AddCommand(groupGetStatusCmd)
 	groupCmd.AddCommand(groupJoinCmd)
 	groupCmd.AddCommand(groupLeaveCmd)
 }
@@ -253,6 +257,33 @@ var groupListCmd = &cobra.Command{
 		}
 
 		err = printJSON(groups)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	},
+}
+
+var groupGetStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Get a particular groups status.",
+	RunE: func(command *cobra.Command, args []string) error {
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		groupID, _ := command.Flags().GetString("group")
+		groupStatus, err := client.GetGroupStatus(groupID)
+		if err != nil {
+			return errors.Wrap(err, "failed to query group status")
+		}
+		if groupStatus == nil {
+			return nil
+		}
+
+		err = printJSON(groupStatus)
 		if err != nil {
 			return err
 		}
--- a/internal/api/group.go
+++ b/internal/api/group.go
@@ -25,6 +25,8 @@ func initGroup(apiRouter *mux.Router, context *Context) {
 	groupRouter.Handle("", addContext(handleGetGroup)).Methods("GET")
 	groupRouter.Handle("", addContext(handleUpdateGroup)).Methods("PUT")
 	groupRouter.Handle("", addContext(handleDeleteGroup)).Methods("DELETE")
+	groupRouter.Handle("/status", addContext(handleGetGroupStatus)).Methods("GET")
+
 }
 
 // handleGetGroup responds to GET /api/group/{group}, returning the group in question.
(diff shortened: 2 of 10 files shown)


cmd/cloud/group.go
```diff
@@ -261,6 +265,33 @@ var groupListCmd = &cobra.Command{
...
+		if err != nil {
+			return errors.Wrap(err, "failed to query group status")
+		}
+		if groupStatus == nil {
+			return nil
+		}
+
+		err = printJSON(groupStatus)
+		if err != nil {
+			return err
```
gigawhitlocks, line 288: Could you `Wrap` this `err` with some context?

internal/store/group.go
```diff
@@ -182,6 +169,89 @@ func (sqlStore *SQLStore) GetGroupRollingMetadata(groupID string) (*GroupRolling
...
+	}
+
+	return &model.GroupStatus{
+		InstallationsCount:           totalInstallations,
+		InstallationsRolledOut:       rolledOutInstallations,
+		InstallationsAwaitingRollOut: installationsToBeRolled,
+	}, nil
+}
+
+func (sqlStore *SQLStore) queryInstallationsToBeRolledOut(columns []string, group *model.Group, dest interface{}) error {
```
gigawhitlocks, line 229: :+1: thank you for this refactor

internal/store/group.go
```diff
@@ -182,6 +169,89 @@ func (sqlStore *SQLStore) GetGroupRollingMetadata(groupID string) (*GroupRolling
...
+	}
+	rolledOutInstallations, err := rolledOutResult.value()
+	if err != nil {
+		return nil, err
+	}
+
+	totalInstallations, err := sqlStore.countInstallationsInGroup(group)
+	if err != nil {
+		return nil, err
+	}
```
gigawhitlocks, line 220: we should `Wrap` these errors to provide some more context

internal/store/group.go
```diff
@@ -182,6 +169,89 @@ func (sqlStore *SQLStore) GetGroupRollingMetadata(groupID string) (*GroupRolling
...
+//
+// Note: This function uses the same conditions as GetGroupRollingMetadata to be more accurate
+// with the internal state seen by the Group Supervisor.
+func (sqlStore *SQLStore) GetGroupStatus(groupID string) (*model.GroupStatus, error) {
+	group, err := sqlStore.GetGroup(groupID)
+	if err != nil {
+		return nil, err
+	}
+	if group == nil {
+		return nil, nil
```
gigawhitlocks, line 183: I'm not sure I understand why the case where the group cannot be found isn't an error. It seems to me like this should return `nil, errors.New(`.

I don't feel strongly about this (1/5), however I'm interested in your thinking regarding returning `nil, nil` here.

gigawhitlocks (changes requested):
Overall this looks great! I would just like to see all (or most) of the bare calls to `return nil, err` modified to be `return nil, errors.Wrap(err, "some context about the call site")`, and I'm curious about that one instance returning `nil, nil`, but then after that this will be good to go.

internal/store/group.go
jwilander: Generally we don't treat "not found" as an error in the stores since it's a valid success response. For example, I could be running the command to check if a resource exists or not and I wouldn't want it to error on it not being found.

For this case I'm 0/5 since it's trying to pull a group's status and not get a resource directly

cmd/cloud/group.go
```diff
@@ -261,6 +265,33 @@ var groupListCmd = &cobra.Command{
 	},
 }
 
+var groupGetStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Get a particular groups status.",
```
gabrieljackson, line 270:
```suggestion
	Short: "Get a particular group's status.",
```

internal/api/group.go
```diff
@@ -25,6 +25,8 @@ func initGroup(apiRouter *mux.Router, context *Context) {
 	groupRouter.Handle("", addContext(handleGetGroup)).Methods("GET")
 	groupRouter.Handle("", addContext(handleUpdateGroup)).Methods("PUT")
 	groupRouter.Handle("", addContext(handleDeleteGroup)).Methods("DELETE")
+	groupRouter.Handle("/status", addContext(handleGetGroupStatus)).Methods("GET")
+
```
gabrieljackson, line 29: nit: extra line

internal/store/group.go
gabrieljackson, line 220: We could save ourselves one query here. We can just take the total installation count and subtract installationsToBeRolled to get rolledOutResult for example. If we keep all three queries maybe we want to just double check the counts to make sure they add up and then log a warning if something went wrong?

model/group_status.go
```diff
@@ -0,0 +1,29 @@
...
+	"encoding/json"
+	"io"
+)
+
+// GroupStatus represents the status of a group.
+type GroupStatus struct {
+	InstallationsCount           int64
+	InstallationsRolledOut       int64
+	InstallationsAwaitingRollOut int64
+}
```
gabrieljackson, line 17: Only 1/5 on this, but maybe it would be clearer if this was something like

```suggestion
type GroupStatus struct {
	InstallationsTotal          int64
	InstallationsUpdated        int64
	InstallationsRequiringUpdate int64
}
```
The only reason I suggest it is `rolling` normally just refers to the installations actively being updated at any one moment.

gabrieljackson (changes requested):
Looks good! Just a few comments/thoughts.

internal/store/group.go
Szymongib: Yes, that is what I saw in the code.
Although I am not a fan of the pattern with returning `nil, nil`,  I think we should return `404` if querying for the status of a group that does not exist, but I am also 0/5. 
If no one has objections I will leave it as it is.

internal/store/group.go
Szymongib: I might be wrong on that but I am not sure if that is correct. From what I understood from the existing code:
- rolled out installations are those:
  - with the latest group sequence
  - with stable state
- awaiting roll out are those:
  - not with the latest group sequence
  - with stable state

This leaves room for all other installations that are being currently updated/created/deleted. Those are included in the total count but are included neither in RolledOut nor AwaitingRollOut.
I think we can latter expand data returned by this endpoint to also include a count of installations in other states.

cmd/cloud/group.go
Szymongib: I see those errors from `printJSON` are not wrapped anywhere else in the whole client, so I wonder if it is not better to keep it consistent.
If that is a mistake on the other hand I can add wrap in the whole file.

model/group_status.go
Szymongib: 👍  I would just change `InstallationsRequiringUpdate` to `InstallationsAwaitingUpdate`.

internal/store/group.go
jwilander: I'm good with as is 👍

cmd/cloud/group.go
gigawhitlocks: Eh, let's keep it consistent for now.

internal/store/group.go
gigawhitlocks: Sounds good to me

gigawhitlocks (approved):

internal/store/group.go
gabrieljackson: +1 from me as well

gabrieljackson (approved):
Awesome!

cpanato (approved):
this.is.AMAZING!
/lgtm

PR #313, opened 2020-09-29
Add read replica configuration to installations

gabrieljackson:
New installations created with external databases now have the
correct read replica settings applied. This includes both MySQL
and PostgreSQL database engines for multitenant and single tenant
database types.

Fixes https://mattermost.atlassian.net/browse/MM-28840

```release-note
Add read replica configuration to installations
```

### DIFF
--- a/internal/tools/aws/database.go
+++ b/internal/tools/aws/database.go
@@ -23,7 +23,9 @@ import (
 )
 
 const mysqlConnStringTemplate = "mysql://%s:%s@tcp(%s:3306)/mattermost?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s"
+const mysqlConnReaderStringTemplate = "%s:%s@tcp(%s:3306)/mattermost?readTimeout=30s&writeTimeout=30s"
 const postgresConnStringTemplate = "postgres://%s:%s@%s:5432/mattermost?sslmode=disable&connect_timeout=10"
+const postgresConnReaderStringTemplate = "postgres://%s:%s@%s:5432/mattermost?sslmode=disable&connect_timeout=10"
 
 // RDSDatabase is a database backed by AWS RDS.
 type RDSDatabase struct {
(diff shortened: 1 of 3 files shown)


internal/tools/aws/database.go
```diff
@@ -23,7 +23,9 @@ import (
 )
 
 const mysqlConnStringTemplate = "mysql://%s:%s@tcp(%s:3306)/mattermost?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s"
+const mysqlConnReaderStringTemplate = "%s:%s@tcp(%s:3306)/mattermost?readTimeout=30s&writeTimeout=30s"
```
gabrieljackson, line 26: Note to reviewers that these constants should probably be moved to helper methods like for multitenant databases, but didn't want to refactor that in this PR. Will clean up in a future PR.

internal/tools/aws/helpers.go
```diff
@@ -108,26 +108,37 @@ func MattermostRDSDatabaseName(installationID string) string {
...
-// MattermostPostgresConnString formats the connection string used by Mattermost
+// MattermostPostgresConnStrings formats the connection strings used by Mattermost
 // servers to access a PostgreSQL database.
-func MattermostPostgresConnString(schema, endpoint, username, password string) string {
-	return fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
+func MattermostPostgresConnStrings(schema, endpoint, username, password string) (string, string) {
+	dbConnection := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
 		username, password, endpoint, schema)
+	readReplicas := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
+		username, password, endpoint, schema)
```
angeloskyratzakos, line 135: If those 2 variables return the same thing why we are splitting them?

Szymongib (approved):

internal/tools/aws/helpers.go
gabrieljackson: Good point, Angelos. You are right that I could have used the same variable here, but I wanted it to be a bit more clear that we may want to change the configuration of one or the other without accidentally updating both.

PR #316, opened 2020-10-01
[MM 29112] Migrating to Prometheus operator and enabling Thanos sidecar.

stylianosrigas:
#### Summary
- Migrating to Prometheus operator and enabling Thanos Sidecar.
- Adding clusterID label via the operator to be used by Thanos and Grafana.
- Add a Migrate utility group function that can used when utility group migrations are required.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29112

#### Release Note
```release-note
- Migrating to Prometheus operator and enabling Thanos sidecar.
- Adding clusterID label via the operator to be used by Thanos and Grafana.
```


gabrieljackson (commented):
So I see that we are using a helm chart to deploy the Prometheus operator. Is this something that is bundled with the operator itself or is this a 3rd party solution? I am just wondering if it makes sense to manually create and update the operator like we do with other operators to reduce our dependency on helm some.

stylianosrigas:
> So I see that we are using a helm chart to deploy the Prometheus operator. Is this something that is bundled with the operator itself or is this a 3rd party solution? I am just wondering if it makes sense to manually create and update the operator like we do with other operators to reduce our dependency on helm some.

@gabrieljackson This is using the official Prometheus operator Helm chart, maintained by the Prometheus community. https://github.com/prometheus-community/helm-charts. It used to be under stable but everything now is moved out of there and owned by respective teams.

gabrieljackson:
@stylianosrigas, awesome and thanks for the info. I just wanted to make sure we aren't going to get ourselves in a spot where the helm chart could end up with no future support.

gabrieljackson:
I had a chat with @stylianosrigas and he is going to attempt to put the migration cleanup logic in the code. This should help us reduce the chance of encountering an error when upgrading.

stylianosrigas:
@gabrieljackson @cpanato You can now have another look. I did a modification in the utility group code to remove the Prometheus utility group and put a comment there that this will be reverted as soon as we reprovision existing clusters. I added a Migration function in all utility groups that we can use for automating migrations. In this case it is used to check if the Prometheus chart is deployed and remove it.

internal/provisioner/helm_utils.go
```diff
@@ -198,6 +209,29 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 	return nil
 }
 
+// deleteHelmChart is used to delete Helm charts.
+func deleteHelmChart(chart helmDeployment, configPath string, logger log.FieldLogger) error {
+	arguments := []string{
+		"--debug",
+		"del",
```
cpanato, line 216: can we use the full command name? `delete`

helm-charts/prometheus_operator_values.yaml
```diff
@@ -0,0 +1,2111 @@
...
+kubeTargetVersionOverride: ""
+
+## Provide a name to substitute for the full names of resources
+##
+fullnameOverride: ""
+
+## Labels to apply to all resources
+##
+commonLabels: {}
+# scmhash: abc123
```
cpanato, line 24: can we remove these examples that we are not using?

stylianosrigas: Done thanks :)

internal/provisioner/helm_utils.go
stylianosrigas: Done thanks :)

internal/provisioner/utility_group.go
```diff
@@ -147,17 +157,39 @@ func (group utilityGroup) ProvisionUtilityGroup() error {
...
+	// 	err := utility.CreateOrUpgrade()
+	// 	if err != nil {
+	// 		return errors.Wrap(err, "failed to upgrade one of the cluster utilities")
+	// 	}
+
+	// 	err = group.cluster.SetUtilityActualVersion(utility.Name(), utility.ActualVersion())
+	// 	if err != nil {
+	// 		return err
+	// 	}
+	// }
```
gabrieljackson, line 192: A bit of extra logic here to save for later? ;)

internal/provisioner/utility_group.go
```diff
@@ -147,17 +157,39 @@ func (group utilityGroup) ProvisionUtilityGroup() error {
...
+		} else {
+			err := utility.CreateOrUpgrade()
+			if err != nil {
+				return errors.Wrap(err, "failed to upgrade one of the cluster utilities")
+			}
+
+			err = group.cluster.SetUtilityActualVersion(utility.Name(), utility.ActualVersion())
+			if err != nil {
+				return err
+			}
```
gabrieljackson, line 178: Instead of checking for `prometheus` here, can't we just add the migration to the `CreateOrUpgrade` method for prometheus? That method could just call `Migrate` right away.

gabrieljackson (commented):
I originally envisioned just having the prometheus utility migrate itself to the new operator version, but this works as well with a new utility entry. We will just have to clean up the old entry in the metadata later.

Just left a few comments below.

cpanato (approved):
thank you!

internal/provisioner/utility_group.go
stylianosrigas: Great suggestion thanks, done 👍

gabrieljackson (approved):
👍

PR #320, opened 2020-10-06
MM-29370 Provide for a configurable directory for the Utility Group

gigawhitlocks:
#### Summary

In order to consume values for the Utility Group from a separate repository, synced into the container via [git-sync](https://github.com/kubernetes/git-sync), as configured in [this merge request](https://gitlab.internal.core.cloud.mattermost.com/cloud/cloud-private/-/merge_requests/199), the Provisioner needs a minimally-intrusive way to specify the source directory for these values. This PR provides a command line argument to allow the user to specify, at `cloud server` time, where these values exist. This should provide the necessary flexibility for development, while allowing the new syncing functionality to operate in production.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29370

#### Release Note

```release-note
Added a command line option to allow specification of a non-default location for UtilityGroup charts and values 
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -127,7 +127,11 @@ var serverCmd = &cobra.Command{
 		keepFilestoreData, _ := command.Flags().GetBool("keep-filestore-data")
 		useExistingResources, _ := command.Flags().GetBool("use-existing-aws-resources")
 
-		model.UtilityValuesDirectory, _ = command.Flags().GetString("utility-values-directory")
+		utilityValuesDirectory, _ := command.Flags().GetString("utility-values-directory")
+		if _, err := os.Stat(utilityValuesDirectory); err != nil {
+			return errors.Wrapf(err, "failed to load Utility Group values from directory %s", utilityValuesDirectory)
+		}
+		model.SetUtilityValuesDirectory(utilityValuesDirectory)
 
 		wd, err := os.Getwd()
 		if err != nil {
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -49,7 +49,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("cluster-installation-supervisor", true, "Whether this server will run a cluster installation supervisor or not.")
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
-
+	serverCmd.PersistentFlags().String("utility-values-directory", "./helm-charts", "The location of the Helm charts and other metadata for Cluster Utilities.")
 	serverCmd.PersistentFlags().Int("poll", 30, "The interval in seconds to poll for background work.")
 	serverCmd.PersistentFlags().Int("cluster-resource-threshold", 80, "The percent threshold where new installations won't be scheduled on a multi-tenant cluster.")
 	serverCmd.PersistentFlags().Int("cluster-resource-threshold-scale-value", 0, "The number of worker nodes to scale up by when the threshold is passed. Set to 0 for no scaling. Scaling will never exceed the cluster max worker configuration value.")
@@ -127,6 +127,8 @@ var serverCmd = &cobra.Command{
 		keepFilestoreData, _ := command.Flags().GetBool("keep-filestore-data")
 		useExistingResources, _ := command.Flags().GetBool("use-existing-aws-resources")
 
+		model.UtilityValuesDirectory, _ = command.Flags().GetString("utility-values-directory")
+
 		wd, err := os.Getwd()
 		if err != nil {
 			wd = "error getting working directory"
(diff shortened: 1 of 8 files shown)


cmd/cloud/server.go
```diff
@@ -127,6 +127,8 @@ var serverCmd = &cobra.Command{
 		keepFilestoreData, _ := command.Flags().GetBool("keep-filestore-data")
 		useExistingResources, _ := command.Flags().GetBool("use-existing-aws-resources")
 
+		model.UtilityValuesDirectory, _ = command.Flags().GetString("utility-values-directory")
```
gabrieljackson, line 130: Can we do a directory-exists check here and at least log a warning if it doesn't?

model/cluster_utility.go
```diff
@@ -31,6 +31,10 @@ const (
 	TeleportDefaultVersion = "0.3.0"
 )
 
+// UtilityValuesDirectory is the directory in which Helm charts and
+// other Utility values are stored
+var UtilityValuesDirectory string
```
gabrieljackson, line 36: I am a tiny bit wary about using a shared variables for this, but it should be fine. Perhaps we can have a `Set` method for it and not export it to help ensure we don't update it when we shouldn't? It would also be nice if when retrieving the value we had method like `GetUtilityValues(yaml string)` that would create the complete directory string for you so there was less chance of error there.

gabrieljackson (changes requested):

model/cluster_utility.go
gigawhitlocks: Sure these are both reasonable improvements to the PR as written, can do.

cmd/cloud/server.go
gigawhitlocks: Sure sure

gigawhitlocks: This should actually be an error case, if the directory doesn't exist, so I'm just going to have the process fail to start if it is not missing, or otherwise there will be weird behavior much later in execution.

The only reason I did not include this check to begin with is that I was thinking it would be more appropriate for the PR when the `helm-charts` directory is actually removed, but if I'm going to add this check, I think it should be a fatal error if the directory is not found.

gigawhitlocks: Note that this change essentially deprecates the `helm-charts` directory, but I think it makes sense to deploy this code, see the git-sync sidecar in action, and then remove the `helm-charts` directory as part of the next deployment. Let me know if any of these assumptions are bad ideas.

gigawhitlocks:
I did a small refactor in the most recent commit; for changes more closely related to the code review comments take a look at 
3c09315 @gabrieljackson

gigawhitlocks:
Added do-not-merge because this shouldn't be merged [until after this MR](https://gitlab.internal.core.cloud.mattermost.com/cloud/cloud-private/-/merge_requests/199) which also needs some review.

internal/provisioner/nginx.go
```diff
@@ -87,6 +87,9 @@ func (n *nginx) Destroy() error {
 	return nil
 }
 
+func (n *nginx) ValuesPath() string {
+	return model.UtilityValuesDirectory() + "/nginx_values.yaml"
+}
```
gabrieljackson, line 92: nit: looks like we may want an empty line inserted after this line before the next method.

model/cluster_utility.go
```diff
@@ -65,6 +69,19 @@ func NewUtilityMetadata(metadataBytes []byte) (*UtilityMetadata, error) {
...
+// whole Provisioner, not just a single cluster
+func SetUtilityValuesDirectory(dir string) {
+	utilityValuesDirectory = dir
+}
+
+// UtilityValuesDirectory returns the global location of the values &
+// charts directory
+func UtilityValuesDirectory() string {
+	return utilityValuesDirectory
+}
```
gabrieljackson, line 83: I was thinking that instead of returning just the values path the helper function would return `fmt.Sprintf("%s/%s", utilityValuesDirectory, yamlFileName)` or something so the directory is always built the same way.

0/5 on this, but maybe an idea for future improvement.

gabrieljackson (approved):
Looks good! No blockers.

model/cluster_utility.go
gigawhitlocks: I realize I implemented this a little differently than you originally suggested, so let me explain my thinking a little: I figured that the assembly would be delegated out to the different utilities via `ValuesPath` so that they _don't_ always have to be built the same way. Each Utility then is responsible for tracking its configuration directory and isn't bound to a certain structure beyond the root directory, in the future. Does that make sense @gabrieljackson ?

Szymongib (approved):
LGTM

stylianosrigas (approved):
We need to add Thanos as well, which will be merged tomorrow morning, but other than that it looks good to me @gigawhitlocks

gigawhitlocks:
Awesome, thanks for the heads-up about Thanos @stylianosrigas

internal/provisioner/nginx.go
gigawhitlocks: Doesn't show as Outdated because it's the next line, but I added a line here while handling the merge https://github.com/mattermost/mattermost-cloud/pull/320/commits/70372c2e80b542e7a577eedae7e774d594d3c47e

PR #321, opened 2020-10-08
MM-29236 - Introduce Cluster and Installation Annotations

Szymongib:
#### Summary

This PR introduces the initial version of Cluster and Installation Annotations to the API and Database model:
-  New tables added:
    - `Annotation`
    - `ClusterAnnotation`
    - `InstallationAnnotation`
- Annotations can be added when creating Clusters or Installations with API calls:
  ```
  ...
      "annotations": [
        "abcd", "test", "super-awsome"
    ]
  ...
  ```
- Added `--annotation` flag to `create cluster` and `create installation` commands.
- Introduced `InstllationDTO` and `ClusterDTO` structs as a way to represent Installations and Clusters with data joined from different tables (ex. Annotations)
- Introduced the `CloseConnection` function for some unit tests to close the database connection as we started to hit the Postgress limit when running all unit tests.

API for updating Annotations on existing Clusters and Installation as well as using those for scheduling will be done in different PRs.

P.S.
Sorry for the huge PR, I have underestimated the number of required changes 😞

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29236

#### Release Note

```release-note
Introduce Cluster and Installation Annotations
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -36,7 +36,7 @@ func init() {
 	clusterCreateCmd.Flags().String("fluentbit-version", model.FluentbitDefaultVersion, "The version of Fluentbit to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("nginx-version", model.NginxDefaultVersion, "The version of Nginx to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("teleport-version", model.TeleportDefaultVersion, "The version of Teleport to provision. Use 'stable' to provision the latest stable version published upstream.")
-	clusterCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the cluster. Accepts multiple values.")
+	clusterCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the cluster. Accepts multiple values, for example: '... --annotation abc --annotation def'")
 
 	clusterProvisionCmd.Flags().String("cluster", "", "The id of the cluster to be provisioned.")
 	clusterProvisionCmd.Flags().String("prometheus-version", "", "The version of Prometheus to provision, no change if omitted. Use \"stable\" as an argument to this command to indicate that you wish to remove the pinned version and return the utility to tracking the latest version.")
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -36,6 +36,7 @@ func init() {
 	clusterCreateCmd.Flags().String("fluentbit-version", model.FluentbitDefaultVersion, "The version of Fluentbit to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("nginx-version", model.NginxDefaultVersion, "The version of Nginx to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("teleport-version", model.TeleportDefaultVersion, "The version of Teleport to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the cluster. Accepts multiple values.")
 
 	clusterProvisionCmd.Flags().String("cluster", "", "The id of the cluster to be provisioned.")
 	clusterProvisionCmd.Flags().String("prometheus-version", "", "The version of Prometheus to provision, no change if omitted. Use \"stable\" as an argument to this command to indicate that you wish to remove the pinned version and return the utility to tracking the latest version.")
@@ -113,6 +114,7 @@ var clusterCreateCmd = &cobra.Command{
 		kopsAMI, _ := command.Flags().GetString("kops-ami")
 		zones, _ := command.Flags().GetString("zones")
 		allowInstallations, _ := command.Flags().GetBool("allow-installations")
+		annotations, _ := command.Flags().GetStringArray("annotation")
 
 		request := &model.CreateClusterRequest{
 			Provider:               provider,
@@ -121,6 +123,7 @@ var clusterCreateCmd = &cobra.Command{
 			Zones:                  strings.Split(zones, ","),
 			AllowInstallations:     allowInstallations,
 			DesiredUtilityVersions: processUtilityFlags(command),
+			ExtraAnnotations:       annotations,
 		}
 
 		size, _ := command.Flags().GetString("size")
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -123,7 +123,7 @@ var clusterCreateCmd = &cobra.Command{
 			Zones:                  strings.Split(zones, ","),
 			AllowInstallations:     allowInstallations,
 			DesiredUtilityVersions: processUtilityFlags(command),
-			ExtraAnnotations:       annotations,
+			Annotations:       annotations,
 		}
 
 		size, _ := command.Flags().GetString("size")
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -28,6 +28,7 @@ func init() {
 	installationCreateCmd.Flags().String("database", model.InstallationDatabaseMysqlOperator, "The Mattermost server database type. Accepts mysql-operator, aws-rds, aws-rds-postgres, or aws-multitenant-rds")
 	installationCreateCmd.Flags().String("filestore", model.InstallationFilestoreMinioOperator, "The Mattermost server filestore type. Accepts minio-operator or aws-s3")
 	installationCreateCmd.Flags().StringArray("mattermost-env", []string{}, "Env vars to add to the Mattermost App. Accepts format: KEY_NAME=VALUE. Use the flag multiple times to set multiple env vars.")
+	installationCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the installation. Accepts multiple values.")
 	installationCreateCmd.MarkFlagRequired("owner")
 	installationCreateCmd.MarkFlagRequired("dns")
 
@@ -98,6 +99,7 @@ var installationCreateCmd = &cobra.Command{
 		database, _ := command.Flags().GetString("database")
 		filestore, _ := command.Flags().GetString("filestore")
 		mattermostEnv, _ := command.Flags().GetStringArray("mattermost-env")
+		annotations, _ := command.Flags().GetStringArray("annotation")
 
 		envVarMap, err := parseEnvVarInput(mattermostEnv, false)
 		if err != nil {
@@ -105,17 +107,18 @@ var installationCreateCmd = &cobra.Command{
 		}
 
 		request := &model.CreateInstallationRequest{
-			OwnerID:       ownerID,
-			GroupID:       groupID,
-			Version:       version,
-			Image:         image,
-			Size:          size,
-			DNS:           dns,
-			License:       license,
-			Affinity:      affinity,
-			Database:      database,
-			Filestore:     filestore,
-			MattermostEnv: envVarMap,
+			OwnerID:          ownerID,
+			GroupID:          groupID,
+			Version:          version,
+			Image:            image,
+			Size:             size,
+			DNS:              dns,
+			License:          license,
+			Affinity:         affinity,
+			Database:         database,
+			Filestore:        filestore,
+			MattermostEnv:    envVarMap,
+			ExtraAnnotations: annotations,
 		}
 
 		dryRun, _ := command.Flags().GetBool("dry-run")
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -118,7 +118,7 @@ var installationCreateCmd = &cobra.Command{
 			Database:         database,
 			Filestore:        filestore,
 			MattermostEnv:    envVarMap,
-			ExtraAnnotations: annotations,
+			Annotations: annotations,
 		}
 
 		dryRun, _ := command.Flags().GetBool("dry-run")
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -28,7 +28,7 @@ func init() {
 	installationCreateCmd.Flags().String("database", model.InstallationDatabaseMysqlOperator, "The Mattermost server database type. Accepts mysql-operator, aws-rds, aws-rds-postgres, or aws-multitenant-rds")
 	installationCreateCmd.Flags().String("filestore", model.InstallationFilestoreMinioOperator, "The Mattermost server filestore type. Accepts minio-operator or aws-s3")
 	installationCreateCmd.Flags().StringArray("mattermost-env", []string{}, "Env vars to add to the Mattermost App. Accepts format: KEY_NAME=VALUE. Use the flag multiple times to set multiple env vars.")
-	installationCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the installation. Accepts multiple values.")
+	installationCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the installation. Accepts multiple values, for example: '... --annotation abc --annotation def'")
 	installationCreateCmd.MarkFlagRequired("owner")
 	installationCreateCmd.MarkFlagRequired("dns")
 
--- a/internal/api/installation.go
+++ b/internal/api/installation.go
@@ -172,7 +172,7 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
 		State:           model.InstallationStateCreationRequested,
 	}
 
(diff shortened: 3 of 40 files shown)


gigawhitlocks (commented):
Looks real good overall; just some small things and some questions for now. No approval just yet because I haven't finished reading all the code, but given I shouldn't be doing work today I wanted to submit this partial review so that it'll be here for you Monday.

cmd/cloud/installation.go
```diff
@@ -98,24 +99,26 @@ var installationCreateCmd = &cobra.Command{
...
+			Version:          version,
+			Image:            image,
+			Size:             size,
+			DNS:              dns,
+			License:          license,
+			Affinity:         affinity,
+			Database:         database,
+			Filestore:        filestore,
+			MattermostEnv:    envVarMap,
+			ExtraAnnotations: annotations,
```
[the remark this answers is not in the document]
Szymongib: I like to sort the fields more logically, for example, start with the ID, Name, etc. and the go to more complex ones like for example `ExtraAnnotations` or `MattermostEnv` but that is just something I am used to. I would say I am 1/5 on that.

internal/store/annotation.go
```diff
@@ -0,0 +1,247 @@
...
+func (sqlStore *SQLStore) getOrCreateAnnotations(db dbInterface, annotations []*model.Annotation) ([]*model.Annotation, error) {
+	for i, ann := range annotations {
+		annotation, err := sqlStore.getOrCreateAnnotation(db, ann)
+		if err != nil {
+			return nil, err
+		}
+		annotations[i] = annotation
+	}
+
+	model.SortAnnotations(annotations)
```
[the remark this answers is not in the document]
Szymongib: It is not really important, it is done mainly for convenience and testing.

model/cluster_dto.go
```diff
@@ -0,0 +1,41 @@
...
+//
+
+package model
+
+import (
+	"encoding/json"
+	"io"
+)
+
+// ClusterDTO represents cluster entity with connected data.
```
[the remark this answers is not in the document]
Szymongib: Sorry, my bad. DTO stands for Data Transfer Object. I was considering different names but did not really find anything better. I will definitely add it to the comment.

internal/store/cluster.go
```diff
@@ -17,11 +17,9 @@ var clusterSelect sq.SelectBuilder
 
 func init() {
 	clusterSelect = sq.
-		Select(
-			"ID", "Provider", "Provisioner", "ProviderMetadataRaw", "ProvisionerMetadataRaw",
+		Select("Cluster.ID", "Provider", "Provisioner", "ProviderMetadataRaw", "ProvisionerMetadataRaw",
```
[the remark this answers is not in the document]
Szymongib: Ah, good catch! It is leftover from some other queries I have tried. It is not needed here so I will clean it up.

cmd/cloud/installation.go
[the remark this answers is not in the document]
gabrieljackson: I personally prefer "more logical" field sorting as well. That's how most of the current resources ares sorted that way. 1/5 on this.

internal/store/annotation.go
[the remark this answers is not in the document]
gigawhitlocks: If it doesn't matter, I think we should skip sorting them, as there's always a runtime performance hit for sorting

model/cluster_dto.go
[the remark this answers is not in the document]
gigawhitlocks: Thanks!

cmd/cloud/installation.go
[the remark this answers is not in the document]
gigawhitlocks: I actually agree, I just like to fall back to alphabetical once everything is separated into logical groups. The above struct, for instance, seems to be in no particular order, unless this is ordered somewhat in order of importance and I overlooked that?

internal/store/annotation.go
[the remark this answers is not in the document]
Szymongib: Well, I do not think that it would impact performance in a significant way as it is hard for me to imagine there being more than 10 annotations on a single installation/cluster, but I agree if that is not really needed it can be removed.

model/annotation.go
```diff
@@ -0,0 +1,53 @@
...
+	"sort"
+)
+
+const (
+	annotationMinLen        = 3
+	annotationMaxLen        = 64
+	annotationAllowedFormat = "annotations must start with letter and can contain only lowercase letters, numbers or '_', '-' characters"
+)
+
+var annotationRegex = regexp.MustCompile("^[a-z].[a-z0-9_-]*$")
```
gabrieljackson, line 19: Do we want to support `=` as another valid character so we can have an annotation like `zone=us-west`?

model/annotation.go
```diff
@@ -0,0 +1,53 @@
...
+// AnnotationsFromStringSlice converts list of strings to list of annotations
+func AnnotationsFromStringSlice(names []string) ([]*Annotation, error) {
+	if names == nil {
+		return nil, nil
+	}
+
+	annotations := make([]*Annotation, 0, len(names))
+	for _, n := range names {
+		if len(n) < annotationMinLen || len(n) > annotationMaxLen {
+			return nil, fmt.Errorf("error: annotation '%s' is invalid: annotations must be between %d and %d characters long", n, annotationMinLen, annotationMaxLen)
```
gabrieljackson, line 36: I believe we can leave out the `error:` prefix here and on the error below as this will be returned up the stack.

gabrieljackson (changes requested):
Looks really good! I have a few requests + questions.

model/installation_request.go
```diff
@@ -22,18 +22,19 @@ import (
...
+	Image            string
+	DNS              string
+	License          string
+	Size             string
+	Affinity         string
+	Database         string
+	Filestore        string
+	APISecurityLock  bool
+	MattermostEnv    EnvVarMap
+	ExtraAnnotations []string
```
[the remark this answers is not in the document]
Szymongib: My reasoning for the name `ExtraAnnotations` was that we will probably be adding some annotations automatically in the future like for example `multi-tenant` or `private` based on the Installations affinity. That is why I wanted to highlight that those passed to the API are extra ones.

I am open to change it, just wanted to give you full context, let me know what do you think.

model/annotation.go
Szymongib: Good question. I thought about extending annotations to also contain the value, but to be honest I think it would be better to do it on the database level. We could achieve it quite simply by adding the `Value` column to the `ClusterAnnotations`/`InstallationAnnotations` tables but until then, I think it is better to keep the `=` out of the annotation name so that we do not end up in the situation like having annotation `zone=us-west` with value `us-west` 😄

gigawhitlocks: This regex isn't correct; the `.` in the center means "any character," not a literal period. The period needs to be escaped and a test case needs to be added that includes an incorrect separator character. The regex also allows strings that end in a period with no other characters after, which doesn't seem to be intended. It also forces that the annotation start with exactly one lower case letter, a-z, followed by any character (intended to be a period), followed by an optional string of characters (no periods).

Here are some examples of the current regex behaving in an unintended way:
```
gore> regexp.MustCompile("^[a-z].[a-z0-9_-]*$").MatchString("a?b")
true
gore> regexp.MustCompile("^[a-z].[a-z0-9_-]*$").MatchString("aa.b")
false
gore> regexp.MustCompile("^[a-z].[a-z0-9_-]*$").MatchString("a.")
true
gore> regexp.MustCompile("^[a-z].[a-z0-9_-]*$").MatchString("aaaa.")
false
```

I believe the pattern that we actually want is: at least one lowercase letter, followed by a period, followed by a group of alphanumeric characters with at least one letter, and the string doesn't end in a period. Exactly one period is allowed.

I have written the correct regex for this pattern. Also note that in Go, you want to define regular expression pattern using a literal string with backticks, rather than standard double-quoted strings which allow for special escapes like `\n`, or you will be forced to double-escape special characters that you want to use as a literal, e.g. `\\.` instead of `\.` for a literal period.

The argument to `MustCompile` should be changed to:
```
`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`
```

Example test cases:
```
gore> :import regexp
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("aaa?bb.cc")
false
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("aaabb.cc")
true
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("aaabb.cc(")
false
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("a.cc(")
false
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("a.cc")
true
gore> regexp.MustCompile(`^[a-zA-Z]+(\.[a-z0-9_-]+)?$`).MatchString("banana")
true
```

model/annotation_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestAnnotationsFromStringSlice(t *testing.T) {
+
+	t.Run("valid annotations", func(t *testing.T) {
```
gigawhitlocks, line 15: We need some valid test cases that include the one allowed period by the regex

model/annotation_test.go
```diff
@@ -0,0 +1,98 @@
...
+			{
+				"to long name",
+				[]string{"abcd", "my-annotation-1-with-super-long-name-that-is-not-allowed-but-cool"},
+			},
+			{
+				"upper case letter",
+				[]string{"abcd", "xyz", "Abcd"},
+			},
+			{"to short name", []string{"abcd", "ab"}},
+			{"not allowed character ' '", []string{"ab cd"}},
```
gigawhitlocks, line 56: These tests pass because there's only one allowed character at the beginning of the pattern, not because there is a disallowed character in the invalid input :(

internal/api/lock.go
```diff
@@ -13,8 +13,8 @@ import (
 
 // lockCluster synchronizes access to the given cluster across potentially
 // multiple provisioning servers.
-func lockCluster(c *Context, clusterID string) (*model.Cluster, int, func()) {
-	cluster, err := c.Store.GetCluster(clusterID)
+func lockCluster(c *Context, clusterID string) (*model.ClusterDTO, int, func()) {
```
gigawhitlocks, line 16: For this function and others where the return type has changed to a DTO object, we need to change the name of the variables at the call sites from e.g. `cluster` to `clusterDTO` to avoid the `cluster.Cluster` stutter and also to reflect the new type in those locations.

gigawhitlocks: `go-rename` should make short work of this.

internal/store/annotation_test.go
```diff
@@ -0,0 +1,160 @@
...
+// See LICENSE.txt for license information.
+//
+
+package store
+
+import (
+	"github.com/mattermost/mattermost-cloud/internal/testlib"
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
gigawhitlocks, line 11: We usually separate the imports into groups: stdlib, third party, and first party imports.

internal/api/installation.go
```diff
@@ -294,10 +301,10 @@ func handleUpdateInstallation(c *Context, w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	if patchInstallationRequest.Apply(installation) {
+	if patchInstallationRequest.Apply(installation.Installation) {
```
gigawhitlocks, line 304: With the change of this `installation` variable to a `DTO` object, its name should change as well (as mentioned in another comment on the changing of a function definition for `lockCluster`) to `installationDTO` or something similar, to reflect the change in its type, and also to prevent the seemingly-nonsensical "variable `installation` with value `Installation` reference above, as well as the stutter ("`thing Thing`").

internal/store/installation.go
```diff
@@ -199,7 +204,39 @@ func (sqlStore *SQLStore) GetUnlockedInstallationsPendingWork() ([]*model.Instal
 }
 
 // CreateInstallation records the given installation to the database, assigning it a unique ID.
-func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation) error {
+func (sqlStore *SQLStore) CreateInstallation(installation *model.Installation, annotations []*model.Annotation) error {
```
gigawhitlocks, line 207: Could you help me understand the value that the DTO object brings for holding Annotations versus just storing them directly on the `model.Thing` (`model.Installation` in this case) object directly? Using the new DTO objects to hold this extra metadata requires a great number of changes to call sites and return types, seen throughout this change, including here.

gigawhitlocks (changes requested):
Just a few concerns and questions. The biggest issue that I see is the regular expression pattern for the annotations being incorrect.

model/annotation.go
Szymongib: Thanks for spotting that, regex is in fact incorrect but it is only the period that is wrong. The intention was to allow only `[a-z0-9_-]` characters and force annotation to start with a letter. I have misused `.` instead of a `+` so I believe the correct regex should be just that:
 ```
^[a-z]+[a-z0-9_-]*$
```

File: cmd/cloud/cluster.go [hunk 14]
```diff
@@ -36,6 +36,7 @@ func init() {
 	clusterCreateCmd.Flags().String("fluentbit-version", model.FluentbitDefaultVersion, "The version of Fluentbit to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("nginx-version", model.NginxDefaultVersion, "The version of Nginx to provision. Use 'stable' to provision the latest stable version published upstream.")
 	clusterCreateCmd.Flags().String("teleport-version", model.TeleportDefaultVersion, "The version of Teleport to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().StringArray("annotation", []string{}, "Additional annotations for the cluster. Accepts multiple values.")
```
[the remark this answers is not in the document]
Szymongib: Yes, you use flag multiple time `--annotation thing1 --annotation thing2`. I agree it makes sense to add it to the usage.

model/annotation_test.go
Szymongib: As mentioned in the comment above it was not intended to allow period in annotations.

model/annotation_test.go
Szymongib: As mentioned above regarding the period character, I believe those tests are actually correct.

internal/store/installation.go
Szymongib: As annotations are stored in a different table than the Installation/Cluster I do not think it is good to store them directly in the struct.
For example, if I only need the Installation object I do not need to join other tables containing annotations data and at the same, I avoid confusion caused by pulling only the Installation object with `Annotations` field being `nil` when in fact there might be some annotations assign for this particular Installation.

internal/store/annotation_test.go
Szymongib: I guess it would make sense to enforce the imports order with `goimports` as part of CI check. I will create a ticket for that.

model/annotation.go
gabrieljackson: Adding a `Value` column makes sense to me!

model/installation_request.go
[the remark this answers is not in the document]
gabrieljackson: Hmm, can you clarify how the provisioner would know which annotations should be automatically added? Affinity here would simply be a result of which annotations we provided when creating the installation.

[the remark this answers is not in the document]
Szymongib: Sure, after some discussion in a design doc, we concluded that for Installations it would be good to add the annotation based on the affinity. For example, the multi-tenant installation would automatically get `multi-tenant` annotation, to prevent having installations without any annotations, as such Installation could be scheduled on any Cluster.

model/annotation_test.go
gigawhitlocks: Yep, I understand now

internal/store/annotation_test.go
gigawhitlocks: That's a great idea thank you

internal/store/installation.go
gigawhitlocks: Ok I had a suspicion after thinking about this a bit that it had to do with the underlying data model. Thank you for the explanation

gigawhitlocks (approved):
Looks great!

model/installation_request.go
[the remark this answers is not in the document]
Szymongib: As discussed, changed the `ExtraAnnotations` to `Annotations`.

gabrieljackson (approved):
Great work! This looks awesome.

PR #322, opened 2020-10-09
[MM-28659] Introducing Thanos utility group

stylianosrigas:
#### Summary
- Introducing Thanos utility group.
- Two R53 records created as part of each Thanos deployment. One targeting then normal ingress and the other one the GRPC ingress that will be used by the central querier. 
- Autoscaling, requests and limits set for Thanos Querier.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28659

#### Release Note
```release-note
- Introducing Thanos utility group.
- Two R53 records created as part of each Thanos deployment. One targeting then normal ingress and the other one the GRPC ingress that will be used by the central querier. 
- Autoscaling, requests and limits set for Thanos Querier.
```

### DIFF
--- a/internal/provisioner/thanos.go
+++ b/internal/provisioner/thanos.go
@@ -1,0 +1,218 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package provisioner
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+type thanos struct {
+	awsClient      aws.AWS
+	cluster        *model.Cluster
+	kops           *kops.Cmd
+	logger         log.FieldLogger
+	provisioner    *KopsProvisioner
+	desiredVersion string
+	actualVersion  string
+}
+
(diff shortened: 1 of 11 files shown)


model/cluster_utility.go
```diff
@@ -14,6 +14,8 @@ const (
 	PrometheusCanonicalName = "prometheus"
 	// PrometheusOperatorCanonicalName is the canonical string representation of prometheus operator
 	PrometheusOperatorCanonicalName = "prometheus-operator"
+	// ThanosCanonicalName is the canonical string representation of prometheus
```
gabrieljackson, line 17:
```suggestion
	// ThanosCanonicalName is the canonical string representation of thanos
```

internal/provisioner/thanos.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (t *thanos) Migrate() error {
+	return nil
+}
+
+func (t *thanos) NewHelmDeployment() *helmDeployment {
+	privateDomainName, err := t.awsClient.GetPrivateZoneDomainName(t.logger)
+	if err != nil {
+		t.logger.WithError(err).Error("unable to lookup private zone name")
```
gabrieljackson, line 178: We store this value even if there is an error. I see in the `CreateOrUpgrade` method we look up the privateDomainName again. Do we need to even look it up here?

stylianosrigas: @gabrieljackson Thanks you are right on that. We were doing the same thing in Prometheus, which we shouldn't. I updated both Prometheus and Thanos to pass the DNS as part of the NewHelmDeployment function. This way if there is an error we will catch it in the CreateORUpgrade where the privateDomainname is set.

gabrieljackson (approved):
👍

cpanato (approved):

PR #324, opened 2020-10-14
MM-29005 - Migrate to Helm 3 for provisioning utility group

Szymongib:
#### Summary

As Helm 2 is being deprecated, the Provisioner should use Helm 3 instead.

Changes in the PR:
- Utility group for new clusters installed using `helm3`.
- Utility group migrated from Helm 2 to Helm 3 during reprovisioning using the `2to3` plugin.
- Tiller cleaned up after the migration of all charts in the utility group.
- `helm3` and `2to3` plugin added to the Docker image.

The previous PR #317 was closed as most changes were removed in favor of new ones.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29005

#### Release Note

```release-note
Migrate to Helm 3 for provisioning utility groups
```


gabrieljackson (approved):
Looks good to me! I have a few questions, but no blockers.

1. It looks like we are keeping the helm2 binary only so that we can properly check if tiller exists in clusters. Can you confirm that this is basically correct?
2. I just want to double check that multiple cluster provisions can be run without issue. i.e. no errors are thrown when fully migrated to helm 3 and provision is called again.

Szymongib:
> Looks good to me! I have a few questions, but no blockers.
> 
> 1. It looks like we are keeping the helm2 binary only so that we can properly check if tiller exists in clusters. Can you confirm that this is basically correct?
> 2. I just want to double check that multiple cluster provisions can be run without issue. i.e. no errors are thrown when fully migrated to helm 3 and provision is called again.

1. For checking if Tiller exists as well as checking if a release exists created by Helm 2 [here](https://github.com/mattermost/mattermost-cloud/pull/324/files#diff-15d435c297b7591681f6973d0a718b85743fb8063defd2925f51474c8febb7a1R287) so that we do not try to migrate if the release does not exist. I think it is also used in the cleanup of the old Prometheus chart done by @stylianosrigas.
2. Yes, I have checked the following scenarios:
- Cluster provisioned with Helm 2 -> reprovisioned (updated to Helm 3) -> reprovisioned one more time
- Cluster provisioned with Helm 3 -> reprovisioned

build/Dockerfile
```diff
@@ -26,12 +26,15 @@ LABEL name="Mattermost Cloud" \
 
 ENV CLOUD=/mattermost-cloud/cloud \
     USER_UID=10001 \
-    USER_NAME=cloud
+    USER_NAME=cloud \
+    HELM_DATA_HOME="/helm3/data"
```
gigawhitlocks, line 30: Thank

helm-charts/prometheus_operator_values.yaml
```diff
@@ -1853,7 +1853,7 @@ prometheus:
          resources:
            requests:
              storage: 100Gi
-       selector: {}
+#       selector: {}
```
gigawhitlocks, line 1856: Just want to call this line out and make sure this was intentional since the line is commented-out and not deleted.

gigawhitlocks (approved):

helm-charts/prometheus_operator_values.yaml
Szymongib: Yes, Helm 3 did not like it and it seems like most not used values are commented, so I have followed along.

Szymongib:
Adjusted migrations to the new Thanos chart. Seems like there was an issue in values [here](https://github.com/mattermost/mattermost-cloud/pull/324/files#diff-0f1b433df0dc7849d8648e17c438f46de055209505ac816f697d902dc0a943efR127). @stylianosrigas, please let me know if that is not correct.

stylianosrigas (approved):
LGTM @Szymongib. The Thanos change looks correct 👍

PR #327, opened 2020-10-19
Handle group updates during installation rollouts

gabrieljackson:
This change allows the provisioner to dynamically handle group
configuration updates while installations belonging to that group
are being upgraded. This allows bad group configuration changes to
be corrected and for affected installations to be repaired. To
accomplish this, the following changes were made:

 - Update installation group sequence numbers when resource changes
   are made. This allows the provisioner to then reference the group
   sequence number on the installation to know if the group has
   been changed since the updates began.
 - Check group sequence number multiple times during upgrades. Each
   "tick" now checks that group configuration hasn't changed. When
   the installation goes to the stable state, one final group check
   is performed with the group under lock to confirm that the
   installation update is complete.
 - Rollback to `update-requested` whenever any group changes are
   made before the installation becomes stable. This ensures that
   the installation update happens again in full.
 - Send out more webhooks for important state change events.
 - Provide a cluster installation k8s resource check method. This
   method can be used in the future for additional checks as needed.
   (This is currently unused, but is planned for the future and I don't
   want the code to become lost)

Here is a snippet from manual testing with webhooks captured in `cwl`:

```
2020/10/18 20:44:14 [ INST | 6fir ] stable -> update-requested
2020/10/18 20:44:17 [ CLIN | j6zr ] stable -> reconciling
2020/10/18 20:44:17 [ INST | 6fir ] update-requested -> update-in-progress
2020/10/18 20:44:55 [ INST | 6fir ] update-in-progress -> update-requested
2020/10/18 20:44:59 [ CLIN | j6zr ] reconciling -> stable
2020/10/18 20:45:07 [ CLIN | j6zr ] stable -> reconciling
2020/10/18 20:45:07 [ INST | 6fir ] update-requested -> update-in-progress
2020/10/18 20:47:21 [ CLIN | j6zr ] reconciling -> stable
2020/10/18 20:47:26 [ INST | 6fir ] update-in-progress -> stable
```

Multiple changes to the group object were made resulting in multiple reverts on the installation update process.

Fixes https://mattermost.atlassian.net/browse/MM-29866

```release-note
Handle group updates during installation rollouts
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -42,6 +42,10 @@ type installationStore interface {
 	UnlockClusterInstallations(clusterInstallationID []string, lockerID string, force bool) (bool, error)
 	UpdateClusterInstallation(clusterInstallation *model.ClusterInstallation) error
 
+	GetGroup(id string) (*model.Group, error)
+	LockGroup(groupID, lockerID string) (bool, error)
+	UnlockGroup(groupID, lockerID string, force bool) (bool, error)
+
 	GetMultitenantDatabase(multitenantdatabaseID string) (*model.MultitenantDatabase, error)
 	GetMultitenantDatabases(filter *model.MultitenantDatabaseFilter) ([]*model.MultitenantDatabase, error)
 	GetMultitenantDatabaseForInstallationID(installationID string) (*model.MultitenantDatabase, error)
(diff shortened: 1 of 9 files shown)


angeloskyratzakos (approved):
LGTM

internal/supervisor/installation.go
```diff
@@ -593,6 +647,43 @@ func (s *InstallationSupervisor) waitForUpdateStable(installation *model.Install
 	return model.InstallationStateStable
 }
 
+// Unused stub function
```
gigawhitlocks, line 650: curious, why keep this if it is unused?

gigawhitlocks (approved):

internal/supervisor/installation.go
gabrieljackson: Sorry, I mentioned it in the PR description, but it will be used shortly and don't want it to become stale.

PR #329, opened 2020-10-19
Add cloud dashboard command

gabrieljackson:
The cloud dashboard allows for monitoring all important cloud
resources via auto-refreshing CLI output.

This change also enhances group status functionality by providing
a count for the number of installations currently being updated.

Fixes https://mattermost.atlassian.net/browse/MM-29580

```release-note
Add cloud dashboard command
```

### DIFF
--- a/cmd/cloud/dashboard.go
+++ b/cmd/cloud/dashboard.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -148,5 +149,5 @@ var dashboardCmd = &cobra.Command{
 }
 
 func toStr(i int) string {
-	return fmt.Sprintf("%d", i)
+	return strconv.Itoa(i)
 }
--- a/cmd/cloud/dashboard.go
+++ b/cmd/cloud/dashboard.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package main
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/gosuri/uilive"
+	"github.com/olekukonko/tablewriter"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+func init() {
+	dashboardCmd.PersistentFlags().String("server", defaultLocalServerAPI, "The provisioning server whose API will be queried.")
+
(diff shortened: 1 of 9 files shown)


cmd/cloud/dashboard.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		writer := uilive.New()
+		writer.Start()
+
+		for {
+			tableString := &strings.Builder{}
+			table := tablewriter.NewWriter(tableString)
+			table.SetAlignment(tablewriter.ALIGN_LEFT)
+			table.SetHeader([]string{"TYPE", "TOTAL", "STABLE", "WIP"})
```
Szymongib, line 47: Maybe something useful for the future would be to split `WIP` column into those that are actually being processed and those that are in a failed state, that are waiting for some action.

That is just an idea for the future, not something for this PR.

cmd/cloud/dashboard.go
```diff
@@ -0,0 +1,152 @@
...
+			for i := refreshSeconds; i > 0; i-- {
+				fmt.Fprintf(writer, "%s\nUpdating in %d seconds...\n", renderedDashboard, i)
+				time.Sleep(time.Second)
+			}
+		}
+	},
+}
+
+func toStr(i int) string {
+	return fmt.Sprintf("%d", i)
```
Szymongib, line 151: I believe that a more canonical way of doing it is using `Itoa`
```suggestion
	return strconv.Itoa(i)
```

model/group_status.go
```diff
@@ -13,6 +13,7 @@ import (
 type GroupStatus struct {
 	InstallationsTotal          int64
 	InstallationsUpdated        int64
+	InstallationsBeingUpdated   int64
```
Szymongib, line 16: I like the idea of adding new field but I am not sure about the name, as by subtracting `updatedCount` and `awaitingUpdateCount` from `totalCount` we get all unstable Installations including those in failed states, hibernated or those being created/deleted.

I am afraid that is something that may prove a little confusing in the future. Maybe something like `InstallationsUnstable` would describe it more accurately?

Szymongib (approved):
Very cool feature 👍 
Just one small concern about naming 😉

model/group_status.go
gabrieljackson: Yep, that's a good point. I will change the name here.

cmd/cloud/dashboard.go
gabrieljackson: Yeah, I like this idea for a future PR. Any resources that would fall in the `WIP` column will be listed in detail under the dashboard table so we won't miss that they may have failed, but having this in the table would also be nice.

stylianosrigas (approved):
Awesome 👌

PR #330, opened 2020-10-20
MM-29237 - Schedule Installations based on Annotations

Szymongib:
#### Summary

This PR makes use of Annotations during the scheduling of Installation to the Cluster.
The filtering of clusters is guided by the following principles:
- Installations can be scheduled on any cluster which contains all Annotations present on the Installation.
- Clusters can contain any additional Annotations not present on the Installation.
- Installation without Annotations can be scheduled on any available cluster.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29237

#### Release Note

```release-note
Schedule Installations on Clusters based on the Annotations
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -127,7 +127,7 @@ var clusterCreateCmd = &cobra.Command{
 			Zones:                  strings.Split(zones, ","),
 			AllowInstallations:     allowInstallations,
 			DesiredUtilityVersions: processUtilityFlags(command),
-			Annotations:       annotations,
+			Annotations:            annotations,
 		}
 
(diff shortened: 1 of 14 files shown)


cpanato (approved):
can we have some docs about how to use that?

Szymongib:
> can we have some docs about how to use that?

Yeah, that is a good point. Will the README be sufficient, or is there some better place?

cpanato:
@Szymongib we can create a docs directory and add things there

Szymongib:
@cpanato I have added docs. Please take a look let me know if there is something more you would like to see in them.

cpanato (approved):

gabrieljackson (approved):
Great tests and docs!

PR #332, opened 2020-10-20
Add group no-op update sequence bump option

gabrieljackson:
The group update API now allows for a group's sequence number to
be increased by one with no changes to the group. This allows for
a clean way to ensure all installations in a group have the latest
baseline configuration and resources without making any changes to
the group.

This change also includes a change where GroupID and Sequence
number will be added as manifest labels on the ClusterInstallation
custom resource. These labels can be used for more efficient
troubleshooting.

Fixes https://mattermost.atlassian.net/browse/MM-27272

```release-note
Add group no-op update sequence bump option
```

### DIFF
--- a/model/group_request.go
+++ b/model/group_request.go
@@ -73,6 +73,8 @@ type PatchGroupRequest struct {
 	Version       *string
 	Image         *string
 	MattermostEnv EnvVarMap
+
+	ForceSequenceUpdate bool
 }
 
 // Apply applies the patch to the given group.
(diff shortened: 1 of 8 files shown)


jwilander (approved):
LGTM

model/group_request.go
```diff
@@ -105,6 +107,12 @@ func (p *PatchGroupRequest) Apply(group *Group) bool {
 		}
 	}
 
+	// This special value allows us to bump the group sequence number even when
+	// the patch contains no group modifications.
+	if p.ForceSequenceUpdate {
```
gigawhitlocks, line 112: 0/5 you could simplify this to just `p.ForceSequnceUpdate = applied` couldn't you?

gigawhitlocks (approved):
LGTM

model/group_request.go
gabrieljackson: `applied` is actually the important var that is influenced by `p.ForceSequenceUpdate` and `p.ForceSequenceUpdate` being false doesn't mean `applied` should be false. I think this is about as clean as you can make it, but thanks for the input!

cpanato (approved):

PR #334, opened 2020-10-20
Handle group configuration changes on installation creation

gabrieljackson:
This change corrects an issue to allow group configuration changes
to be handled correctly during installation creation. The creation
flow now mimics the update flow when the created installation is
part of a group.

Fixes https://mattermost.atlassian.net/browse/MM-29883

```release-note
Handle group configuration changes on installation creation
```


jwilander (approved):
LGTM

gabrieljackson:
Doing a few follow-up tests so putting the do not merge label on until that is complete.

gabrieljackson:
The less-important of the two group checks unfortunately can't be done very cleanly without a lot more work. I have removed it for the time being and will create a ticket for adding it in later.

This PR still covers the original issue that was experienced and allows installation creation to handle group changes to some degree.

PR #337, opened 2020-10-22
remove ssh access from 0.0.0.0/0

cpanato:
#### Summary
remove ssh access from 0.0.0.0/0 when creating new clusters and set the CIDRs from the VPN and from the Command and control cluster

two new flags need to be provided:
- env - which env it is running (default `dev`) possible values for us: `test` / `staging` / `prod`
- vpn-list-cidr - the cidrs for the VPN - Mandatory

For the existing clusters, we agree to do the process manually

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-28913

#### Release Note

```release-note
remove ssh access from 0.0.0.0/0 for new kops clusters
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -46,6 +46,7 @@ func init() {
 	serverCmd.PersistentFlags().String("listen", ":8075", "The interface and port on which to listen.")
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
+	serverCmd.PersistentFlags().StringSlice("vpn-list-cidr", []string{"0.0.0.0/0"}, "The list of VPN CIDRs to allow communication with the clusters.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
 	serverCmd.PersistentFlags().Bool("machine-readable-logs", false, "Output the logs in machine readable format.")
 	serverCmd.PersistentFlags().Bool("dev", false, "Set sane defaults for development")
@@ -92,6 +93,11 @@ var serverCmd = &cobra.Command{
 		allowListCIDRRange, _ := command.Flags().GetStringSlice("allow-list-cidr-range")
 		if len(allowListCIDRRange) == 0 {
 			return errors.New("allow-list-cidr-range must have at least one value")
+		}
+
+		vpnListCIDR, _ := command.Flags().GetStringSlice("vpn-list-cidr")
+		if len(vpnListCIDR) == 0 {
+			return errors.New("vpn-list-cidr must have at least one value")
 		}
 
 		logger := logger.WithField("instance", instanceID)
@@ -206,6 +212,7 @@ var serverCmd = &cobra.Command{
 			owner,
 			useExistingResources,
 			allowListCIDRRange,
+			vpnListCIDR,
 			resourceUtil,
 			logger,
 			sqlStore,
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -36,6 +36,7 @@ type KopsProvisioner struct {
 	privateSubnetIds        string
 	publicSubnetIds         string
 	allowCIDRRangeList      []string
+	vpnCIDRList             []string
 	owner                   string
 	useExistingAWSResources bool
 	resourceUtil            *utils.ResourceUtil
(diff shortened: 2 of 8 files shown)


internal/tools/aws/vpc.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+	}
+
+	vpcOut, err := a.Service().ec2.DescribeVpcs(&vpcInput)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to fetch the VPC information using tag %s", vpcTagName)
+	}
+
+	if len(vpcOut.Vpcs) > 1 {
+		return "", errors.New(fmt.Sprintf("expected just one VPC in the list, got %d", len(vpcOut.Vpcs)))
```
gabrieljackson, line 34:
```suggestion
		return "", errors.Errorf("expected just one VPC in the list, got %d", len(vpcOut.Vpcs))
```

gabrieljackson, line 34: If we don't ensure that we have at least one VPC here it will panic on line 37.

```suggestion
	if len(vpcOut.Vpcs) != 1 {
		return "", errors.New(fmt.Sprintf("expected exactly one VPC in the list, got %d", len(vpcOut.Vpcs)))
```

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -86,6 +90,14 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		return errors.Errorf("invalid AWS AMI Image %s", cluster.ProvisionerMetadataKops.AMI)
 	}
 
+	cncVPCName := fmt.Sprintf("mattermost-cloud-%s-command-control", provisioner.environment)
+	cncVPCCIDR, err := awsClient.GetCIDRByVPCTag(cncVPCName, logger)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get the CIDR for the VPC Name %s", cncVPCName)
+	}
```
gabrieljackson, line 97: Does this work in the dev account? I talked to @stylianosrigas and he says there is no VPC name in the dev account that matches this name. He says there is one called `mattermost-cloud-dev`.

cmd/cloud/server.go
```diff
@@ -49,6 +49,8 @@ func init() {
 	serverCmd.PersistentFlags().Bool("cluster-installation-supervisor", true, "Whether this server will run a cluster installation supervisor or not.")
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
+	serverCmd.PersistentFlags().StringSlice("vpn-list-cidr", []string{}, "The list of VPN CIDRs to allow communication with the clusters.")
```
gabrieljackson, line 52: Is it possible to set a default here that is dev friendly like we do above with `allow-list-cidr-range`?

gabrieljackson (changes requested):
Looks good, but I have some comments and questions.

internal/provisioner/kops_provisioner_cluster.go
cpanato: We should add this tag name in dev 
@stylianosrigas wdyt?

cmd/cloud/server.go
```diff
@@ -49,6 +49,8 @@ func init() {
...
+	serverCmd.PersistentFlags().String("env", "dev", "Which envirionment the provisioner is running.")
```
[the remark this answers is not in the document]
cpanato: we dont have set  account aliases

internal/provisioner/kops_provisioner_cluster.go
cpanato: yep confirmed, in dev is called `mattermost-cloud-dev` or we change in dev or add the if here

cpanato: update in dev to follow the same pattern

cmd/cloud/server.go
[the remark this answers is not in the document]
stylianosrigas: @cpanato we do have account aliases set up in each account and we use it in charts, mt database, etc. Example here https://github.com/mattermost/mattermost-cloud/blob/381bc0f5cb267b00f123bfb9820323173c395a02/internal/provisioner/prometheus_operator.go#L74.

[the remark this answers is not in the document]
cpanato: oh nice! thanks for the info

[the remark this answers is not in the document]
stylianosrigas: e.g This is the account alias for Dev:
```json
{
    "AccountAliases": [
        "mattermost-cloud-dev"
    ]
}
```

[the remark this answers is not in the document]
cpanato: and for the other envs all follow the same pattern? @stylianosrigas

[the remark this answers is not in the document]
cpanato: done

cpanato:
@stylianosrigas @gabrieljackson PTAL

internal/tools/aws/vpc.go
```diff
@@ -0,0 +1,36 @@
...
+			},
+		},
+	}
+
+	vpcOut, err := a.Service().ec2.DescribeVpcs(&vpcInput)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to fetch the VPC information using tag %s", vpcTagName)
+	}
+
+	if len(vpcOut.Vpcs) > 1 {
```
gabrieljackson, line 31: I think think we need to ensure that we have exactly one VPC here.
```suggestion
	if len(vpcOut.Vpcs) != 1 {
```

gabrieljackson (changes requested):
Just one last change request. Not sure why my original comment on this got marked as outdated.

internal/tools/aws/vpc.go
cpanato: done

stylianosrigas (approved):
LGTM 👍

gabrieljackson (approved):
👍

PR #340, opened 2020-10-26
Look for .ssh in root directory if not found in $HOME

Szymongib:
#### Summary

As the `$HOME` directory changed in Dockerfile, we need to look for ssh keys in the root directory if `$HOME/.ssh` does not exist.

#### Release Note

```release-note
NONE
```


cpanato (commented):
did you test if this works?

Szymongib:
We decided to move with different approach

PR #348, opened 2020-10-30
MM-30167 Remove unnecessary error-prone DNS check

gigawhitlocks:
#### Summary
We saw a failure to create an instance due to an i/o timeout when looking up a name against 1.1.1.1. We have a number of Workspace availability checks in the Customer Web Server, which render this check unnecessary, and since it can cause creation to fail, it should be removed.

This PR removes this functionality.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30167

#### Release Note

```release-note
Removed DNS check for new Installations
```

### DIFF
--- a/model/installation_request.go
+++ b/model/installation_request.go
@@ -5,15 +5,12 @@
 package model
 
 import (
-	"context"
 	"encoding/json"
 	"io"
-	"net"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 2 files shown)


jwilander (approved):

gabrieljackson (approved):
👍

PR #356, opened 2020-11-06
Use affinity/toleration for services apps in the cluster

cpanato:
#### Summary
Now we have a new instance group in the cluster to hold the utilities services, this PR sets the affinity/toleration to use those nodes.
In case the nodes does not exist it will be scheduled in the "normal" nodes

Also remove the prometheus since we are not using that anymore in favor of prometheus operator

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-30295

#### Release Note

```release-note
Use affinity/toleration for services apps in the cluster
```


helm-charts/prometheus_operator_values.yaml
```diff
@@ -1841,7 +1857,7 @@ prometheus:
       limits:
         cpu: "3"
         memory: 15Gi
```
stylianosrigas, line 1859: @cpanato can we just increase this as well to be able to use the 64GB we will give?

stylianosrigas (approved):
Just one change regarding limit ;)

gabrieljackson (approved):
As long as we have run this on an existing cluster to ensure there are no issues with the removal of the prometheus utility then looks good to me!

helm-charts/prometheus_operator_values.yaml
cpanato: will add 60

cpanato: @stylianosrigas done PTAL

PR #357, opened 2020-11-06
collect only the original node metadata information

cpanato:
#### Summary
Since now we have an external instance group that the provisioner does not know about it we need to get the node that the provisioner know to report the metadata for now

I'm seeing this in the test account

```
"ProvisionerMetadataKops": {
...
        "MasterInstanceType": "t3.large",
        "MasterCount": 3,
        "NodeInstanceType": "r5.xlarge",
        "NodeMinCount": 2,
        "NodeMaxCount": 2,
        "ChangeRequest": {
            "Version": "latest"
        },
        "Warnings": [
            "expected exactly 1 node instance group, but found 2"
        ]
    },
```

which when the provisioner update the metadata if push the last node group

this PR make a workaround to only report the node group the provisioner knows

#### Release Note

```release-note
collect only the original node metadata information
```

### DIFF
--- a/internal/tools/kops/instance_groups.go
+++ b/internal/tools/kops/instance_groups.go
@@ -74,6 +74,12 @@ func (c *Cmd) UpdateMetadata(metadata *model.KopsMetadata) error {
 
 			masterIGCount++
 		case "Node":
+			// TODO: temp fix while we dont support multiple igs
+			if ig.Metadata.Name == "nodes-utilities" {
+				c.logger.Debug("Skipping utility group")
+				continue
+			}
+
 			if AMI == "" {
 				AMI = ig.Spec.Image
 			} else if AMI != ig.Spec.Image {


internal/tools/kops/instance_groups.go
```diff
@@ -83,9 +83,11 @@ func (c *Cmd) UpdateMetadata(metadata *model.KopsMetadata) error {
 			}
 
 			NodeIGCount++
```
gabrieljackson, line 85: Due to this line, we are still going to throw the warning + error message below.

gabrieljackson (changes requested):
I think the logic of this PR needs to be changed so that everything remains the same, but when the role is "Node" we check if the name is whatever the utility group is called and then skip it.

internal/tools/kops/instance_groups.go
cpanato: that is fine, not? at least alert us that we have igs that is not managed by the provisioner

gabrieljackson (approved):
Approving with the idea that this is probably a temp fix and I will review the logic again in a follow-up PR.

cpanato:
yes, this is a temp fix

stylianosrigas (approved):

PR #360, opened 2020-11-10
Add balanced cluster scheduling for installations

gabrieljackson:
A new server startup flag is available to force the installation
supervisor to attempt to schedule installations on clusters with
the lowest average resource utilization. This is done without
cluster locking and should be considered best effort.

This also cleans up some old deprecation warnings.

Fixes https://mattermost.atlassian.net/browse/MM-30044

```release-note
Add balanced cluster scheduling for installations
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -32,7 +32,6 @@ import (
 )
 
 const (
-	clusterRootDir        = "clusters"
 	defaultLocalServerAPI = "http://localhost:8075"
 )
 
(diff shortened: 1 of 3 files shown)


internal/supervisor/installation.go
```diff
@@ -322,6 +342,64 @@ func (s *InstallationSupervisor) createInstallation(installation *model.Installa
...
+func (s *InstallationSupervisor) prioritizeLowerUtilizedClusters(clusters []*model.Cluster, installation *model.Installation, instanceID string, logger log.FieldLogger) []*model.Cluster {
+	lowestResourcePercent := 10000
+	var filteredPrioritizedClusters []*model.Cluster
+
+	for _, cluster := range clusters {
+		if !s.installationCanBeScheduledOnCluster(cluster, installation, logger) {
+			continue
+		}
+
+		clusterResources, err := s.provisioner.GetClusterResources(cluster, true)
```
curtbushko, line 367: GetClusterResources() gives you an accurate used & total CPU/Memory for a cluster but what happens when there are 10 clusters and 100s of nodes? Would you be scanning every cluster on every install?

Even though they say it is less accurate, would it be better to use metrics-server? (not in this PR but for the future)

https://github.com/kubernetes-sigs/metrics-server

gabrieljackson: Yes, you are absolutely correct that this resource calculation is relatively expensive and will take longer with every cluster and node added.

Having a hard toggle via the server startup flag will allow us to use this behavior only when really needed and we fully intend to go with a more-scalable approach when time permits. I will absolutely take a fresh look at the metrics server option to see how much effort it would take to implement.

Thanks for the comment!

jwilander (commented):
Overall looks good - just a couple comments

cmd/cloud/server.go
```diff
@@ -41,24 +40,29 @@ var instanceID string
...
 	serverCmd.PersistentFlags().Bool("group-supervisor", false, "Whether this server will run an installation group supervisor or not.")
 	serverCmd.PersistentFlags().Bool("installation-supervisor", true, "Whether this server will run an installation supervisor or not.")
 	serverCmd.PersistentFlags().Bool("cluster-installation-supervisor", true, "Whether this server will run a cluster installation supervisor or not.")
-	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
-	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
 
-	serverCmd.PersistentFlags().Int("poll", 30, "The interval in seconds to poll for background work.")
+	// Scheduling and installation options
+	serverCmd.PersistentFlags().Bool("balanced-installation-scheduling", false, "Whether to schedule installations on the cluster with the greatest percentage of available resources or not. (slows down scheduling speed as cluster count increases)")
```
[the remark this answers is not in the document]
gabrieljackson: I did think about this option, but was a bit worried about how to handle bad string values on startup. I figured this may be best for now as there are only two options and then it can be changed in the future if we ever have a third option.

That said, 0/5 on this so let me know if you think this should be done now.

internal/supervisor/installation.go
gabrieljackson: Yeah, this definitely could lead to longer scheduling times so we should keep an eye on it and improve when we have other metric options ready.

cmd/cloud/server.go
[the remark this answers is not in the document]
jwilander: Not a big deal - we can always change it in the future

jwilander (approved):

internal/supervisor/installation.go
```diff
@@ -77,31 +77,46 @@ type installationProvisioner interface {
...
+		scheduling:        scheduling,
+		resourceUtil:      resourceUtil,
+		logger:            logger,
+	}
+}
+
+// NewInstallationSupervisorSchedulingOptions creates a new InstallationSupervisorSchedulingOptions.
+func NewInstallationSupervisorSchedulingOptions(balanceInstallations bool, clusterResourceThreshold, clusterResourceThresholdScaleValue int) InstallationSupervisorSchedulingOptions {
+	return InstallationSupervisorSchedulingOptions{
+		balanceInstallations:               balanceInstallations,
```
gigawhitlocks, line 117: Is it possible that we may wish to have other scheduling strategies? 

0/5 but I wonder if this field should be something like `strategy` as an enum-like type (maybe an iota or just typed const strings) with set values to represent the two current strategies (`naive` (instead of a `false` value), and `balanced` (instead of `true)) with an eye towards experimentation with other strategies (like a random scheduler, or round robin, or something more intelligent that we haven't thought of yet)

internal/supervisor/installation.go
```diff
@@ -483,6 +522,60 @@ func (s *InstallationSupervisor) createClusterInstallation(cluster *model.Cluste
...
+
+	////////////////////////////////////////////////////////////////////////////
+	//                              MULTI-TENANCY                             //
+	////////////////////////////////////////////////////////////////////////////
+	// Current model:                                                         //
+	// - isolation=true  | 1 cluster installations                            //
+	// - isolation=false | X cluster installations, where "X" is as many as   //
+	//                     will fit with the given CPU and Memory threshold.  //
+	////////////////////////////////////////////////////////////////////////////
+	if installation.Affinity == model.InstallationAffinityIsolated {
```
gigawhitlocks, line 555: You can outdent the `else` if you combine this boolean with the one below it. 0/5 on this, though, it's just a nitpick.

internal/supervisor/installation_test.go
```diff
@@ -531,7 +535,7 @@ func TestInstallationSupervisor(t *testing.T) {
 	t.Run("creation requested, cluster installations not yet created, no empty clusters", func(t *testing.T) {
 		logger := testlib.MakeLogger(t)
 		sqlStore := store.MakeTestSQLStore(t, logger)
-		supervisor := supervisor.NewInstallationSupervisor(sqlStore, &mockInstallationProvisioner{}, &mockAWS{}, "instanceID", 80, 0, false, false, &utils.ResourceUtil{}, logger)
+		supervisor := supervisor.NewInstallationSupervisor(sqlStore, &mockInstallationProvisioner{}, &mockAWS{}, "instanceID", false, false, standardSchedlingOptions, &utils.ResourceUtil{}, logger)
```
gigawhitlocks, line 538: Did you mean to omit the `u` on this variable name? If you did, I kind of think it's not worth omitting it to save the length and the variable should just be `standardSchedulingOptions`.

gigawhitlocks (approved):
Looks solid; only had some nitpicks.

stylianosrigas (approved):
Nice Gabe, really helpful 💯

internal/supervisor/installation_test.go
gabrieljackson: Nope, didn't mean to omit it and is just a typo. Thanks for catching it.

internal/supervisor/installation.go
gabrieljackson: Yeah, we discussed this idea a bit above in Joram's comment. Definitely something we may want, but will just go with the simpler config for now and will revisit in the future.

PR #362, opened 2020-11-11
MM29235 - Cleanup Helm migration code

Szymongib:
#### Summary

As Helm migration is done, we can remove it from the code.
- Migration code cleaned up
- Helm 2 binary removed
- Helm 3 binary renamed `helm3` -> `helm`

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29235

#### Release Note

```release-note
Cleanup Helm migration code
```

PR #363, opened 2020-11-12
Change required Helm client version to 3.4.x

gigawhitlocks:
#### Summary

I discovered that Helm version 3.3.x fails during `provision` if the `stable` repository already exists on the client machine. This problem is caused by this issue: https://github.com/helm/helm/issues/8771

This shouldn't be an issue running Helm in pods in production, but it does affect dev. This bug was addressed prior to Helm 3.4 and I've updated my dev machine to use Helm 3.4 and I'm able to `provision` without an issue, so this PR just changes the version of Helm that we require as displayed in the README.

#### Ticket Link

No ticket

#### Release Note

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The following is required to properly run the cloud server.
 2. Install [Terraform](https://learn.hashicorp.com/terraform/getting-started/install.html) version v0.11.14
    1. Try using [tfswitch](https://warrensbox.github.io/terraform-switcher/) for switching easily between versions
 3. Install [kops](https://github.com/kubernetes/kops/blob/master/docs/install.md) version 1.17.X
-4. Install [Helm](https://helm.sh/docs/intro/install/) version 2.16.X as `helm` and version 3.3.X as `helm3` with [2to3 plugin](https://github.com/helm/helm-2to3#install)
+4. Install [Helm](https://helm.sh/docs/intro/install/) version 2.16.X as `helm` and version 3.4.X as `helm3` with [2to3 plugin](https://github.com/helm/helm-2to3#install)
 5. Install [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 6. Install [golang/mock](https://github.com/golang/mock#installation) version 1.4.x
 


gabrieljackson (commented):
This is a great find, but should we maybe just change everything to use 3.4.x as part of https://github.com/mattermost/mattermost-cloud/pull/362?

gigawhitlocks:
That's fine with me @gabrieljackson

Szymongib:
I think that the issue was resolved in 3.3.4 and I agree with @gabrieljackson that it would be best to run the same version locally as we run in production just to catch some potential issues earlier. 
I agree with bumping the version everywhere but I would propose doing it in a different PR and not mix it with #362.

gigawhitlocks:
It is not resolved in 3.3.4 @Szymongib . That's the version I had when I observed the problem.

gabrieljackson (approved):

Szymongib (approved):
> It is not resolved in 3.3.4 @Szymongib . That's the version I had when I observed the problem.

Hmm, are you sure? It seems to work fine for me with 3.3.4. Maybe you still had the old stable repo `https://kubernetes-charts.storage.googleapis.com`? It changed recently to `https://charts.helm.sh/stable` and adding repo with the same name and different URL does not seem to work in any version.

gigawhitlocks:
>Maybe you still had the old stable repo https://kubernetes-charts.storage.googleapis.com? It changed recently to https://charts.helm.sh/stable

This is likely, as I don't do anything manually with Helm. However, if this was the case, I don't know why there would be an issue describing the problem I saw, and I don't know why upgrading to 3.4.x would resolve the issue, which it did.

Szymongib:
Thanks for the clarification @gigawhitlocks. I think we can just merge it and create the follow-up ticket for bumping the version in the Docker image, what do you think?

gigawhitlocks:
That sounds perfect. I've opened https://mattermost.atlassian.net/browse/MM-30591 to track the follow-on work.

PR #365, opened 2020-11-13
MM-30277 Add metrics to measure elapsed Installation creation time

gigawhitlocks:
#### Summary

This PR introduces the Prometheus library to the Cloud Provisioner with two metrics being measured to start out with, the time it takes for a Installation to reach `stable` status after being requested and the same for ClusterInstallations, represented as cumulative histograms.

These metrics are exposed on a new endpoint called `/metrics`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30277

#### Release Note

```release-note
Added Prometheus metrics for Installation creation actions.
```


internal/metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+// the Provisioning server
+type CloudMetrics struct {
+	InstallationCreationDurationHist        prometheus.Histogram
+	ClusterInstallationCreationDurationHist prometheus.Histogram
+}
+
+// New creates a new Prometheus-based Metrics object to be used
+// throughout the Provisioner in order to record various performance
+// metrics
+func New() *CloudMetrics {
```
nronas, line 22: I think we can avoid initializing the metrics here and we can define them statically under the metrics package. Something like:

```go
var InstallationCreationDurationHist =  promauto.NewHistogram(
			prometheus.HistogramOpts{
				Name:    "mm_cloud_create_installation_duration_seconds",
				Help:    "The duration of Installation creation tasks",
				Buckets: prometheus.LinearBuckets(0, 30, 20),
			})
```

nronas: Also, the Prometheus client [docs](https://godoc.org/github.com/prometheus/client_golang/prometheus#hdr-A_Basic_Example) suggest to define them as static and check that they have been registered like:

```go
func init() {
    prometheus.MustRegister(InstallationCreationDurationHist)
}
```

gigawhitlocks: Regarding the first comment, the pattern here follows the pattern we use in another internal repo, with the intent being to allow there to be a interface boundary here in the future as necessary and also to avoid package-level variables. If @gabrieljackson agrees that these should be package-level, I'll do that.

Regarding calling `MustRegister` in `init`, the Prometheus metrics are only useful when the Provisioning server is initialized as a server, via the `cloud server` command, so we actually don't need or want these metrics to be initialized at the package level at import time, in case types or methods from this package need to be imported by the `cmd/client` but will not be exposing or recording metrics.

In `cmd/server.go` `Register` is called very close to the beginning of execution and the program will panic if something goes wrong, there, as well, so the practical difference for the `server` command is minimal here.

nronas: Awesome, that makes sense when it comes to the `MustRegister` check. 

When it comes to the first point, my thoughts are that we have currently designed this package to expose public accessors from the struct, which conforms the Prometheus interface. So if in the feature we want to swap Prometheus with something else, the successor has to conform to the same Prometheus interface. What I am trying to say is that I think we are already highly coupled with the Prometheus interface.

I could see better what you are saying if in this package we were implementing a type of an adapter that we could swap backends if needed. Does this make sense?

@gigawhitlocks will ❤️  your thoughts on the above.

gigawhitlocks: >What I am trying to say is that I think we are already highly coupled with the Prometheus interface.

You're right; there are only two benefits that I can foresee with putting an interface here: 1) consistency with other internal repos for the sake of familiarity and 2) potentially having the ability to mock out this implementation for testing.

To be clear: I don't feel strongly about this one way or the other, as both patterns have their advantages, and the only reason I implemented the pattern here instead of just using a package-level variable like the Prometheus docs show (which is certainly simpler and potentially cleaner) was for the consistency with other internal repos, which I inferred to be a requirement, so as for that last part I'd really like feedback from @jwilander or @gabrieljackson before making further changes.

nronas: > I don't feel strongly about this one way or the other

Just to be 100% clear me neither 😄 . Just pretty new to the codebase and just i want to build up context.

Alo 100% 👍 on consistency, if that's how we have other repos set up.

gigawhitlocks: Thanks so much for the review; it really helps me think more rigorously about the choices I've made :)

gabrieljackson: Great ideas covered here and I agree that both proposed implementations probably have strengths and weaknesses. Generally I think it makes sense to only initiate metrics on server startup and to keep things similar to our other codebases unless we have good reasons to change.

All that said, this implementation is good with me and we can change it in the future should we need to.

internal/supervisor/cluster_installation.go
```diff
@@ -244,5 +247,12 @@ func (s *ClusterInstallationSupervisor) checkReconcilingClusterInstallation(clus
 	}
 
 	logger.Info("Cluster installation finished reconciling")
+	s.metrics.ClusterInstallationCreationDurationHist.Observe(
+		elapsedTimeInSeconds(clusterInstallation.CreateAt))
```
gabrieljackson, line 251: Cluster installations end up going into reconciling state when they are updated which means this metric will expose incorrect data in those circumstances. I am not sure of a great way around this issue without creating new states which could have other side effects if we aren't careful.

gabrieljackson: Maybe we could lookup the installation to know whether to send the new metric data point, but that seems somewhat heavy for this. I will keep thinking on this.

gabrieljackson (commented):
This looks good to me other than the issue with cluster installation creation metrics. I will keep thinking on a solution, but that should be changed or removed before we merge this PR.

internal/supervisor/cluster_installation.go
gigawhitlocks: Oh, hmm, I was worried there might be an issue of this variety

internal/metrics/metrics.go
jwilander: Only thing I'll chime in with is that I think consistency across our codebases is important. If there's a better way we want to start doing it I'm ok with starting in one place and then pushing it out to the other codebases as part of paying off technical debt but we should aim towards being consistent as much as possible 👍

internal/supervisor/cluster_installation.go
gigawhitlocks: I decided to go ahead and remove the ClusterInstallation metrics from this PR. They can be added at a later time, when we determine the best approach for gathering them, whether it be adding more metadata to the model and the store, or something else.

internal/metrics/metrics.go
gigawhitlocks: I don't think the way it's done here or elsewhere is necessarily debt, it's just a "six to one, half-dozen the other" situation as best as I can tell. I'm going to err on the side of consistency here.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// CloudMetrics holds all of the metrics needed to properly instrument
+// the Provisioning server
+type CloudMetrics struct {
+	InstallationCreationDurationHist        prometheus.Histogram
+	ClusterInstallationCreationDurationHist prometheus.Histogram
```
gabrieljackson, line 16: nit: this could be removed

gabrieljackson (approved):
👍

internal/metrics/metrics.go
gigawhitlocks: Augh thanks

jwilander (approved):

gigawhitlocks:
Had a discussion with @gabrieljackson about this PR and the future of gathering metrics for Cloud and we think it's best if we gather metrics from a microservice running alongside the Provisioner and not from the Provisioner itself, as proposed in this PR.

I'd like to hold off on merging this PR for the moment until we've had a chance to discuss that idea as a team.

gigawhitlocks:
Per discussion at planning, this PR should be merged as-is and more advanced metrics gathering can be discussed at a later time; this PR adds the default Prometheus metrics as well as the single custom one, and the default Prometheus metrics provide host and Go runtime information that will be useful on a per-Provisioner process basis regardless as to any more advanced metrics gathering/calculation added later based upon webhook events.

gigawhitlocks:
Re: the last comment, this PR will be ready to merge once it has sign-off from SRE, either @angeloskyratzakos or @stylianosrigas

stylianosrigas (approved):
LGTM @gigawhitlocks

PR #350, opened 2020-11-02
Add a command line tool for load testing the Provisioner

gigawhitlocks:
#### Summary

This PR introduces a new tool in the `cmd/tools` directory which can be installed with `go install ./cmd/tools/cloudburst` from the project base directory.

`cloudburst` is a command line tool for load testing the Provisioner. `cloudburst` allows you to very quickly spin up a huge number of Installations and collects timing data as it runs.

```
Run a load test against a provisioning server.

Usage:
  cloudburst [flags]

Flags:
      --batch int       Specify the number of Installations in each batch. Installations in a batch are install serially and batches are installed in parallel. (default 5)
  -h, --help            help for cloudburst
      --runs int        Number of times to repeat the test (default 1)
      --server string   Location of the Provisioning Server to load test (default "http://localhost:8075")
      --total int       Number of Installations to provision (default 20)
```

The nature of this PR is such that this code won't get "stale" the way a more tightly integrated PR might, so feel free to make larger requests in this PR than normal, as my primary goal here is to provide a tool that has an immediate as well as long term value payoff.

#### Ticket Link

No ticket; addresses "explosive scaling testing" OKR goal.

#### Release Note

```release-note
Added cloudburst, a tool for explosive scaling / load testing.
```

PR #367, opened 2020-11-16
MM-30529 Add ability to specify filestore, db type, and Installation size to load testing tool

gigawhitlocks:
#### Summary

This PR adds the ability to specify the configuration of Installations created during a test run to the `cloudburst` load test tool, as requested in a comment from @gabrieljackson in PR #350 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30529

#### Release Note

```release-note
Added ability to control Installation configuration to the load testing tool
```

### DIFF
--- a/cmd/tools/cloudburst/cloudburst.go
+++ b/cmd/tools/cloudburst/cloudburst.go
@@ -411,34 +411,17 @@ func (b *Blaster) createInstallation() (*cloud.Installation, error) {
 }
 
 func isValidInput(database, filestore, installSize string) error {
-	databaseCheck := func() error {
-		if !cloud.IsSupportedDatabase(database) {
-			return errors.Errorf("invalid database requested: unknown database type %s", database)
-		}
-		return nil
+	if !cloud.IsSupportedDatabase(database) {
+		return errors.Errorf("invalid database requested: unknown database type %s", database)
 	}
 
-	filestoreCheck := func() error {
-		if !cloud.IsSupportedFilestore(filestore) {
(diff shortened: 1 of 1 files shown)


cmd/tools/cloudburst/cloudburst.go
```diff
@@ -397,6 +410,55 @@ func (b *Blaster) createInstallation() (*cloud.Installation, error) {
...
+		} {
+			if validDB == database {
+				return nil
+			}
+		}
+		return errors.Errorf("invalid database requested: unknown database type %s", database)
+	}
+
+	filestoreCheck := func() error {
+		for _, validFilestore := range []string{
```
gabrieljackson, line 430: We have an `IsValid()` check for filestores and databases in the cloud model. Using those would clean this up a bit.

```
// IsSupportedFilestore returns true if the given filestore string is supported.
func IsSupportedFilestore(filestore string) bool {
	return filestore == InstallationFilestoreMinioOperator ||
<CUT>
```

gabrieljackson (approved):
Thanks for the new options. Reusing the model checks for filestores and databases will probably save us some time when new options are added in the future and will clean up the code, but this is great.

Szymongib (approved):
Cool! I think it will be useful.
I agree with @gabrieljackson comment to reuse validation logic where possible.

cmd/tools/cloudburst/cloudburst.go
gigawhitlocks: Thanks @gabrieljackson I looked for such a function, but obviously not hard enough :P

gabrieljackson (approved):
👍

PR #371, opened 2020-11-20
split nginx deployments in two

cpanato:
#### Summary
Split Nginx on two deployments 
for the public one we need to have the real ips and other information and using the current solution we don't have those infos

This PR split nginx in two deployments

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-29321

#### Release Note

```release-note
split nginx deployments in two deployments 
```

### DIFF
--- a/internal/provisioner/nginx_internal.go
+++ b/internal/provisioner/nginx_internal.go
@@ -1,0 +1,137 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package provisioner
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+type nginxInternal struct {
+	awsClient      aws.AWS
+	provisioner    *KopsProvisioner
+	kops           *kops.Cmd
+	logger         log.FieldLogger
+	cluster        *model.Cluster
+	actualVersion  *model.HelmUtilityVersion
+	desiredVersion *model.HelmUtilityVersion
+}
+
+func newNginxInternalHandle(version *model.HelmUtilityVersion, cluster *model.Cluster, provisioner *KopsProvisioner, awsClient aws.AWS, kops *kops.Cmd, logger log.FieldLogger) (*nginxInternal, error) {
+	if logger == nil {
+		return nil, errors.New("cannot instantiate NGINX INTERNAL handle with nil logger")
+	}
+
+	if cluster == nil {
+		return nil, errors.New("cannot create a connection to Nginx internal if the cluster provided is nil")
+	}
+
+	if provisioner == nil {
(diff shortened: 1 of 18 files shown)


internal/provisioner/nginx_internal.go
```diff
@@ -0,0 +1,125 @@
...
+	awsClient      aws.AWS
+	provisioner    *KopsProvisioner
+	kops           *kops.Cmd
+	logger         log.FieldLogger
+	cluster        *model.Cluster
+	desiredVersion string
+	actualVersion  string
+}
+
+func newNginxInternalHandle(desiredVersion string, cluster *model.Cluster, provisioner *KopsProvisioner, awsClient aws.AWS, kops *kops.Cmd, logger log.FieldLogger) (*nginxInternal, error) {
```
gabrieljackson, line 28: Most of the utility handler logic seems like it would be identical to the external nginx utility. Is it possible to reuse most of that code or does it just make sense to copy+paste all of it? What do you think @stylianosrigas?

stylianosrigas: @gabe I agree that we can actually reuse things like the structs and just change the helm deployment.

internal/tools/aws/cluster_management.go
```diff
@@ -202,6 +204,35 @@ func (a *Client) GetAndClaimVpcResources(clusterID, owner string, logger log.Fie
...
+			Values: []*string{
+				aws.String(clusterID),
+			},
+		},
+	}
+	clusterAlreadyClaimedVpcs, err := a.GetVpcsWithFilters(clusterAlreadyClaimedFilter)
+	if err != nil {
+		return ClusterResources{}, err
+	}
+	if len(clusterAlreadyClaimedVpcs) == 0 {
```
gabrieljackson, line 229: We may want to check for `len(clusterAlreadyClaimedVpcs) != 1` in case two VPCs were somehow returned. It's probably safe enough to not check this, but just wanted to note it.

internal/provisioner/nginx_internal.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	n.actualVersion = actualVersion
+	return nil
+}
+
+func (n *nginxInternal) CreateOrUpgrade() error {
+	h, err := n.NewHelmDeployment()
+	if err != nil {
+		return errors.Wrap(err, "failed to generate nginx helm deployment")
```
gabrieljackson, line 73: nit:
```suggestion
		return errors.Wrap(err, "failed to generate nginx internal helm deployment")
```

internal/provisioner/nginx_internal.go
gabrieljackson: After discussing this a bit with @cpanato we decided to proceed with this as is and will do some cleanup later.

gabrieljackson (approved):
No blocking comments. Looks good!

internal/tools/aws/cluster_management.go
cpanato: done

cpanato:
@stylianosrigas this ready for your review and whenever you want to merge and plan the rollout

/priority important-soon

cpanato:
@stylianosrigas when you want to merge and propagate to the environments.

stylianosrigas (approved):
@cpanato LGTM. I need to release the current master because we need these changes in production soon. This will happen today and then we can merge your PR and you can start the changes in Test.

cpanato:
@stylianosrigas merge this when you return from your PTO
we already discussed what needs to do here

PR #383, opened 2020-12-04
MM-31043 Allow utility values to be fetched from authenticated sources

gigawhitlocks:
#### Summary

This PR contains 3 changes (and my apologies for a single PR that does several things):

1) Allows the SRE or developer to specify the environment variable `GITLAB_OAUTH_TOKEN` and set that token to an access token generated in the GitLab UI. When the Provisioner detects that a values path is in Gitlab and the OAuth token is present, it will add it on to the end of the URL in order to authenticate with GitLab to fetch the file.

2) Allows the Provisioner to set the `DesiredVersion` struct values to `null` when provisioning is completed. This is the first step towards renaming `ActualVersion` to `Version` and changing `DesiredVersion` to be renamed to `UtilityGroupChangeRequest` (per our discussion @gabrieljackson )

3) Introduces a command line flag and env variable to control Helm debug logging so that it can be toggled on and off at startup or during runtime.

Example of a `stable` cluster object after this change:
```json
    {
        "ID": "s3g97j84w3dqmbk4izifw6g4ma",
        "State": "stable",
        "Provider": "aws",
        "ProviderMetadataAWS": {
            "Zones": [
                "us-east-1a"
            ]
        },
        "Provisioner": "kops",
        "ProvisionerMetadataKops": {
            "Name": "s3g97j84w3dqmbk4izifw6g4ma-kops.k8s.local",
            "Version": "1.17.14",
            "AMI": "kope.io/k8s-1.17-debian-stretch-amd64-hvm-ebs-2020-11-19",
            "MasterInstanceType": "t3.large",
            "MasterCount": 1,
            "NodeInstanceType": "m5.large",
            "NodeMinCount": 4,
            "NodeMaxCount": 4
        },
        "UtilityMetadata": {
            "DesiredVersions": {
                "PrometheusOperator": null,
                "Thanos": null,
                "Nginx": null,
                "Fluentbit": null,
                "Teleport": null
            },
            "ActualVersions": {
                "PrometheusOperator": {
                    "Chart": "9.4.4",
                    "ValuesPath": "helm-charts/prometheus_operator_values.yaml"
                },
                "Thanos": {
                    "Chart": "2.4.3",
                    "ValuesPath": "helm-charts/thanos_values.yaml"
                },
                "Nginx": {
                    "Chart": "2.15.0",
                    "ValuesPath": "helm-charts/nginx_values.yaml"
                },
                "Fluentbit": {
                    "Chart": "2.8.7",
                    "ValuesPath": "helm-charts/fluent-bit_values.yaml"
                },
                "Teleport": {
                    "Chart": "0.3.0",
                    "ValuesPath": "https://gitlab.internal.core.cloud.mattermost.com/api/v4/projects/33/repository/files/test%2Fnginx_values.yaml?ref=master"
                }
            }
        },
        "AllowInstallations": true,
        "CreateAt": 1607040631646,
        "DeleteAt": 0,
        "APISecurityLock": false,
        "LockAcquiredBy": null,
        "LockAcquiredAt": 0
    }
```

Finally, it must be noted that **GitLab URLs need to refer to the API and be URL encoded**. Entering the path to the file, e.g. `test/nginx_values.yaml?ref=master`, has to be done manually, so the `%2F` in the above example was actually entered verbatim on the command line when passed to `provision`. For extra clarity, this are the commands that were executed:
```
$ export GITLAB_OAUTH_TOKEN=mytoken
$ cloud cluster provision --cluster s3g97j84w3dqmbk4izifw6g4ma \
--teleport-version 0.3.0 --teleport-values 'https://gitlab.internal.core.cloud.mattermost.com/api/v4/projects/33/repository/files/test%2Fnginx_values.yaml?ref=master' \
--thanos-version 2.4.3 --thanos-values helm-charts/thanos_values.yaml \
--nginx-version 2.15.0 --nginx-values helm-charts/nginx_values.yaml \
--fluentbit-version 2.8.7 --fluentbit-values helm-charts/fluent-bit_values.yaml \
--prometheus-operator-version 9.4.4 --prometheus-operator-values helm-charts/prometheus_operator_values.yaml
```

If that's a usability issue let me know and I will think harder about how I can make it easier to input the path to the file. The URL must refer to the API because the regularly accessible URLs browseable from the UI do not accept tokens as URL parameters and I could not discover a way to specify that Helm should use a custom HTTP header when fetching the file.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31043

#### Release Note

```release-note
Added ability to use authenticated GitLab instances with utility values
DesiredVersion in UtilitiesMetadata now empties when provisioning or creation completes
Verbose Helm debugging can now be controlled with --debug-helm at startup
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -25,6 +25,7 @@ import (
 	"github.com/mattermost/mattermost-cloud/internal/store"
 	"github.com/mattermost/mattermost-cloud/internal/supervisor"
 	toolsAWS "github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/tools/helm"
 	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/pkg/errors"
@@ -47,6 +48,7 @@ func init() {
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
+	serverCmd.PersistentFlags().Bool("debug-helm", false, "Whether to include Helm output in debug logs.")
 	serverCmd.PersistentFlags().Bool("machine-readable-logs", false, "Output the logs in machine readable format.")
 	serverCmd.PersistentFlags().Bool("dev", false, "Set sane defaults for development")
 
@@ -80,6 +82,9 @@ var serverCmd = &cobra.Command{
 		if debugMode {
 			logger.SetLevel(logrus.DebugLevel)
 		}
+
+		debugHelm, _ := command.Flags().GetBool("debug-helm")
+		helm.SetVerboseHelmLogging(debugHelm)
 
 		machineLogs, _ := command.Flags().GetBool("machine-readable-logs")
 		if machineLogs {
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -67,6 +67,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("keep-database-data", true, "Whether to preserve database data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("require-annotated-installations", false, "Require new installations to have at least one annotation.")
+	serverCmd.PersistentFlags().String("gitlab-oauth", "", "If Helm charts are stored in a Gitlab instance that requires authentication, provide the token here and it will be automatically set in the environment.")
 }
 
 var serverCmd = &cobra.Command{
@@ -85,6 +86,11 @@ var serverCmd = &cobra.Command{
 
 		debugHelm, _ := command.Flags().GetBool("debug-helm")
 		helm.SetVerboseHelmLogging(debugHelm)
+
+		gitlabOAuthToken, _ := command.Flags().GetString("gitlab-oauth")
+		if gitlabOAuthToken != "" {
+			os.Setenv(model.GitlabOAuthTokenKey, gitlabOAuthToken)
+		}
 
 		machineLogs, _ := command.Flags().GetBool("machine-readable-logs")
 		if machineLogs {
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -10,7 +10,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"os/exec"
+	"strings"
 	"time"
+
+	"net/url"
+	"os"
 
 	"github.com/mattermost/mattermost-cloud/internal/tools/helm"
 	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
@@ -130,6 +134,10 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 		}
 		chart.desiredVersion = currentVersion
 	}
+
+	censoredPath := chart.desiredVersion.ValuesPath
+	chart.desiredVersion.ValuesPath = applyGitlabTokenIfPresent(chart.desiredVersion.ValuesPath)
+
 	arguments := []string{
 		"--debug",
 		"upgrade",
@@ -149,6 +157,11 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 	if chart.desiredVersion.Version() != "" {
 		arguments = append(arguments, "--version", chart.desiredVersion.Version())
 	}
+
+	defer func(chart *helmDeployment) {
+		// so that we don't store the GitLab secret in the database
+		chart.desiredVersion.ValuesPath = censoredPath
+	}(&chart)
 
 	helmClient, err := helm.New(logger)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


gigawhitlocks:
Regarding the discussions we had today, b7b1996 solves the problem with the secret being echoed out on the command line and f6792dc contains the general improvements to make Helm quieter @stylianosrigas

stylianosrigas (approved):
LGTM, thanks for the fix Ian. We can make it better I guess in the future to not need to provide all these urls/paths but for now I think its good. Great work 🚀

gigawhitlocks:
Yeah, if we can think of something more ergonomic in the future I'll be happy to implement it, and I'm already thinking about that now. One thing to note is that it's not necessary to specify the values/versions for _all_ of the utilities at once like I did in my example.

internal/tools/helm/run.go
```diff
@@ -5,20 +5,37 @@
...
 func outputLogger(line string, logger log.FieldLogger) {
 	line = strings.TrimSpace(line)
 	if len(line) == 0 {
 		return
 	}
 
-	logger.Debugf("[helm] %s", line)
+	if os.Getenv("MM_CLOUD_VERBOSE_HELM_OUTPUT") != "" {
+		logger.Debugf("[helm] %s", line)
+	}
```
gabrieljackson, line 38: I would like to see if we can change the implementation of this if possible. Although I see how this would work, we are doing a bunch of logic checks in this function for each log line generated when it seems like we just wouldn't want to call this at all.

I believe we should be able to build a helm logger that simply skips calling this function entirely if we aren't logging. As part of this, I think it should be possible to avoid setting a global env var that requires the `expandedArgs = append(expandedArgs, os.ExpandEnv(arg))` logic above.

internal/tools/exechelper/run.go
```diff
@@ -54,6 +55,12 @@ func Run(cmd *exec.Cmd, logger log.FieldLogger, outputLogger OutputLogger) ([]by
 		"args": cmd.Args,
 	}).Info("Invoking command")
 
+	expandedArgs := []string{}
+	for _, arg := range cmd.Args {
+		expandedArgs = append(expandedArgs, os.ExpandEnv(arg))
+	}
+	cmd.Args = expandedArgs
```
gabrieljackson, line 62: I am a bit concerned about this as we discussed in chat. Due to this affecting all commands, it seems slightly risky in regards to introducing unwanted behavior somewhere that would be harder to track down.

I feel like we can change the helm logger logic to avoid this so I would advocate for trying that if possible.

internal/provisioner/helm_utils.go
```diff
@@ -266,3 +282,17 @@ func (d *helmDeployment) Version() (*model.HelmUtilityVersion, error) {
 
 	return nil, errors.Errorf("unable to get version for chart %s", d.chartDeploymentName)
 }
+
+func applyGitlabTokenIfPresent(original string) string {
+	if os.Getenv(model.GitlabOAuthTokenKey) == "" {
```
gabrieljackson, line 287: Is there a reason we don't make this a server startup flag? There is an upcoming change to allow for all server flags to optionally be set with env vars and by setting it as a server flag we get the help text to provide context.

gabrieljackson (changes requested):
Great changes and looking forward to getting the desired version stuff in. Have a few concerns/thoughts below.

cmd/cloud/server.go
```diff
@@ -47,6 +48,7 @@ func init() {
 	serverCmd.PersistentFlags().String("state-store", "dev.cloud.mattermost.com", "The S3 bucket used to store cluster state.")
 	serverCmd.PersistentFlags().StringSlice("allow-list-cidr-range", []string{"0.0.0.0/0"}, "The list of CIDRs to allow communication with the private ingress.")
 	serverCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
+	serverCmd.PersistentFlags().Bool("debug-helm", false, "Whether to include Helm output in debug logs.")
```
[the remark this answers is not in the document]
gigawhitlocks: I believe so, I had a chat with @stylianosrigas and I was under the impression that making Helm quieter by default is desired. I'll let him chime in.

internal/tools/helm/run.go
gigawhitlocks: Uh @gabrieljackson the call to `os.ExpandEnv` is in a different hunk. Did you comment on the wrong one?

cmd/cloud/server.go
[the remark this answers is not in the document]
gabrieljackson: That's good with me if that is what SRE would prefer.

internal/tools/exechelper/run.go
gigawhitlocks: The line that needs to be censored is actually logged out by the regular log handler and not the Helm logger, it's the line that starts with `Invoking command` just above this hunk. But I will look to see if I can move some things around and introduce a `RunWithEnvironmentVariables` that does the same thing as `Run` but expands env vars like we need for this operation, without potentially affecting all of the other call sites for `Run`.

internal/provisioner/helm_utils.go
gigawhitlocks: No good reason not to make it a server flag, other than that usually flags require a little plumbing to pass the value around. I will add a server flag for this.

internal/tools/exechelper/run.go
gigawhitlocks: Potential resolution to this thread in 4f5a42f

internal/tools/helm/run.go
gigawhitlocks: Resolved in dc3817c

internal/provisioner/helm_utils.go
gigawhitlocks: This is potentially resolved in 7084d73. I couldn't think of any reason the server flag shouldn't just set the environment variable; let me know if you can.

gabrieljackson (approved):
Good fixes! LGTM 👍

Szymongib (approved):
LGTM!

PR #389, opened 2020-12-15
Add kops client caching

gabrieljackson:
This change allows the provisioner to cache kops clients and cluster
configuration in order to increase performance when rapidly managing
external resources. Logic is also added for cache invalidation when
an issue is encountered that may be caused by outdated cache data.

Fixes https://mattermost.atlassian.net/browse/MM-31363

```release-note
Add kops client caching
```

### DIFF
--- a/internal/provisioner/kops_provisioner.go
+++ b/internal/provisioner/kops_provisioner.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package provisioner
+
+import (
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// KopsProvisioner provisions clusters using kops+terraform.
+type KopsProvisioner struct {
+	s3StateStore            string
+	allowCIDRRangeList      []string
+	vpnCIDRList             []string
+	owner                   string
+	useExistingAWSResources bool
+	resourceUtil            *utils.ResourceUtil
+	logger                  log.FieldLogger
+	store                   model.InstallationDatabaseStoreInterface
+	kopsCache               map[string]*kops.Cmd
+}
+
+// NewKopsProvisioner creates a new KopsProvisioner.
+func NewKopsProvisioner(s3StateStore, owner string, useExistingAWSResources bool, allowCIDRRangeList, vpnCIDRList []string,
+	resourceUtil *utils.ResourceUtil, logger log.FieldLogger, store model.InstallationDatabaseStoreInterface) *KopsProvisioner {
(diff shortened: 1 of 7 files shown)


internal/provisioner/kops_provisioner.go
```diff
@@ -0,0 +1,113 @@
...
+	logger.Debugf("Kops config cached at %s for %s", kopsClient.GetKubeConfigPath(), name)
+
+	return kopsClient, nil
+}
+
+func (provisioner *KopsProvisioner) invalidateCachedKopsClient(name string, logger log.FieldLogger) error {
+	kopsClient, ok := provisioner.kopsCache[name]
+	if !ok {
+		logger.Errorf("Could not find kops client cache for %s to invalidate", name)
+		return errors.Errorf("could not find kops client cache for %s to invalidate", name)
```
gigawhitlocks, line 94: Is this necessarily an error? I can't think of a case where a caller would want to invalidate the cache and be unable to proceed if the item is missing, as that may indicate that it is already invalidated. Logging the error out makes sense to me but I'm not sure returning an error does.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -226,20 +192,13 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
...
+	kopsClient, err := provisioner.getCachedKopsClient(cluster.ProvisionerMetadataKops.Name, logger)
 	if err != nil {
-		return errors.Wrap(err, "failed to export kubecfg")
+		return errors.Wrap(err, "failed to get kops client from cache")
 	}
-
-	kopsMetadata := cluster.ProvisionerMetadataKops
-
-	logger.Info("Provisioning cluster")
+	defer provisioner.invalidateCachedKopsClientOnError(err, cluster.ProvisionerMetadataKops.Name, logger)
```
gigawhitlocks, line 201: The intent here to only check `err` at the end for a non-nil value as provided by the `err` assignment on line 197, right? If for instance the call on line 211 produces an err, the cache may remain?

gigawhitlocks (approved):
Looks good! I had a couple of questions but if they don't raise flags for you then I think this will be good to go ahead and merge.

internal/provisioner/kops_provisioner.go
gabrieljackson: Good question. The reason I did the log and the return of the error is so that the method could be more concretely tested in unit tests. Also, in the future, I could see us maybe needing to know with certainty that the cache was properly invalidated and this sets us up for that.

internal/provisioner/kops_provisioner_cluster.go
gabrieljackson: This implementation will invalidate the kops cache if any error is encountered before this method returns. Although many of the errors have nothing to do with kops or the cache, I figured it was better to be safe than sorry.

In the future, it may make sense to introduce a second error value in these flows like `k8sErr` that we use any time something related to kops/k8s is done. This would require the code creator to be very careful to use the error correctly otherwise the cache may not be invalidated when it should be. Still, may be good to follow-up on if the caching works well.

angeloskyratzakos (approved):
👍

PR #397, opened 2021-01-13
MM-30849 - Preserve order of Env variables to not to trigger installation rollout

Szymongib:
#### Summary

When iterating through `map` the order of elements is not preserved. This causes the Provisioner to change the order of environment variables on `ClusterInstallation` even if no env changed. This in turn causes rotation of the Installation's pods. Sorting envs prevents unnecesary rollouts.

I decide to do the sorting on Provisioner side not in the Operator as the order of environment variables might matter for the Operator user, but in our case we know, that it does not.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30849

#### Release Note

```release-note
Sort Installation Env vars passed to ClusterInstallation
```

### DIFF
--- a/model/env.go
+++ b/model/env.go
@@ -6,6 +6,7 @@ package model
 
 import (
 	"encoding/json"
+	"sort"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


jwilander (approved):
Nice find 👍

stylianosrigas (approved):

PR #401, opened 2021-01-19
MM-32053 Remove --debug in call to helm update to avoid overflowing Scanner buffer bug

gigawhitlocks:
#### Summary

@stylianosrigas hit an issue while trying to run `cloud cluster provision` that produced the following error:

```
ERRO[2021-01-19T15:18:06-06:00] failed to scan stdout                         cluster=ou9z9cfuz3nk9y8jt7zgnnwmia cluster-utility=prometheus-operator error="bufio.Scanner: token too long" helm-update=prometheus-community/kube-prometheus-stack instance=wxo7stu54ty7ze55bum4uw4ahe provisioner=kops run=pac7t1mqm7gj9gxsjoxgious6r utility-group=create-handle
```

The root cause is a limitation in the default standard library bufio.Scanner, which produces an error after calling Scan() if it gets a token that is too large, which happens with our current Prometheus Chart. This is because Helm outputs (to stdout) the whole computed chart at the end of the operation and some line in it must be larger than the buffer.

The real and correct solution is to update `bufferAndLog` in `internal/tools/exechelper/run.go` to be a more manual implementation that doesn't use `scanner.Scan()` and can handle arbitrarily long lines. [See the docs here for information about this limitation of the Scanner](https://golang.org/pkg/bufio/#Scanner), specifically the docs say "Scanning stops unrecoverably at EOF, the first I/O error, **or a token too large to fit in the buffer.**"

I believe the proper fix will take a little while to implement though, as it's a little bit involved, so this will allow the Provisioner to function normally again in the meantime by simply reducing the amount of output from Helm so that we don't hit the error from the Scanner, and we can discuss the value and priority of making the exec helper a bit more robust once this immediate bug is resolved.

HOWEVER if others feel that it is important and worth the time to go ahead and make the 'correct' change, I can do that instead of merging this one.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32053

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -144,7 +144,6 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 	chart.desiredVersion.ValuesPath = applyGitlabTokenIfPresent(chart.desiredVersion.ValuesPath)
 
 	arguments := []string{
-		"--debug",
 		"upgrade",
 		chart.chartDeploymentName,
 		chart.chartName,
@@ -180,7 +179,6 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 // deleteHelmChart is used to delete Helm charts.
 func deleteHelmChart(chart helmDeployment, configPath string, logger log.FieldLogger) error {
 	arguments := []string{
-		"--debug",
 		"delete",
 		"--kubeconfig", configPath,
 		"--namespace", chart.namespace,


stylianosrigas (approved):
LGTM as long as we have a ticket in place to track the work for the proper solution.

gabrieljackson (approved):
This is good as a temporary fix, but I do think the proper fix should be prioritized as a follow-up.

gigawhitlocks:
Sounds good, I'll try to do the proper implementation today since you think it should be a priority @gabrieljackson . I knew I was running out of time to get it done yesterday (the root cause analysis took me a little while) so I thought I'd go ahead and post this to see what y'all thought.

gigawhitlocks:
This root cause analysis is incorrect. I have a bugfix in the works for the _actual_ problem.

PR #402, opened 2021-01-21
MM-32053 Properly fetch and decode values file from Gitlab API

gigawhitlocks:
#### Summary

Late in the development of this feature I switched the routine that fetches the values file from GitLab from fetching the raw permalink of the file directly to using the files API and missed a few steps with regard to how that API actually serves the file, namely, it serves the files base64-encoded as part of a JSON object.

This base64-encoded blob is what was blowing up golang's `Scanner`, causing the actual output and behavior we witnessed in test.

This change adds the necessary and missing steps to unmarshal the JSON and decode the YAML file to disk so that it can be used by Helm.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32053

#### Release Note

```release-note
Fixed bug in Gitlab utility values paths fetching logic
```

### DIFF
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -359,7 +359,7 @@ func fetchFromGitlabIfNecessary(path string) (string, func(string), error) {
 
 	return temporaryValuesFile.Name(), func(path string) {
 		if strings.HasPrefix(path, os.TempDir()) {
-			defer os.Remove(path)
+			os.Remove(path)
 		}
 	}, nil
 }
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -7,14 +7,16 @@ package provisioner
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
-
-	"net/url"
-	"os"
 
 	"github.com/mattermost/mattermost-cloud/internal/tools/helm"
 	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
@@ -141,7 +143,11 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 		chart.desiredVersion.ValuesPath = censoredPath
 	}(&chart, censoredPath)
 
-	chart.desiredVersion.ValuesPath = applyGitlabTokenIfPresent(chart.desiredVersion.ValuesPath)
+	var err error
+	chart.desiredVersion.ValuesPath, err = fetchFromGitlabIfNecessary(chart.desiredVersion.ValuesPath)
+	if err != nil {
+		return errors.Wrap(err, "failed to get values file")
+	}
 
 	arguments := []string{
 		"--debug",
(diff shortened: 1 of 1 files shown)


stylianosrigas (approved):
LGTM Ian, thanks 🚀

internal/provisioner/helm_utils.go
```diff
@@ -283,16 +289,66 @@ func (d *helmDeployment) Version() (*model.HelmUtilityVersion, error) {
...
 }
 
-func applyGitlabTokenIfPresent(original string) string {
-	if os.Getenv(model.GitlabOAuthTokenKey) == "" {
-		return original
+type gitlabValuesFileResponse struct {
+	Content string `json:"content"`
+}
+
+const temporaryValuesFile string = "/tmp/helm-values.yaml"
```
gabrieljackson, line 296: Just checking, this works because we never run two different threads that invoke helm at the same time right? I think that is a safe assumption to make for now, but just wanted to double-check my assumption.

internal/provisioner/helm_utils.go
```diff
@@ -283,16 +289,66 @@ func (d *helmDeployment) Version() (*model.HelmUtilityVersion, error) {
...
-			original,
-			model.GitlabOAuthTokenKey)
+
+	content, err := base64.StdEncoding.DecodeString(valuesFileBytes.Content)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to decode base64-encoded YAML file")
 	}
-	return original
+
+	err = ioutil.WriteFile(temporaryValuesFile, []byte(content), 0777)
```
gabrieljackson, line 348: nit: the file permission could be something like `0600` instead right? Considering we are downloading something from the internet in certain cases it may not hurt to limit that a bit.

gabrieljackson (approved):
LGTM. Just a few non-blocking comments.

internal/provisioner/helm_utils.go
gigawhitlocks: I was going to ask you to confirm that 😅

internal/provisioner/helm_utils.go
gigawhitlocks: Since this runs in a single user context in a container it really shouldn't matter but I'm happy to change it to something more restrictive

internal/provisioner/helm_utils.go
gigawhitlocks: Since neither of seem certain about the safety of using the same static file over and over I changed this to use the standard library temp file creator and added a `defer` to delete it in the calling function after the operation is complete @gabrieljackson

internal/provisioner/helm_utils.go
```diff
@@ -148,6 +148,9 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 	if err != nil {
 		return errors.Wrap(err, "failed to get values file")
 	}
+	if strings.HasPrefix(chart.desiredVersion.ValuesPath, os.TempDir()) {
+		defer os.Remove(chart.desiredVersion.ValuesPath)
+	}
```
gabrieljackson, line 153: Instead of this check, you could modify `fetchFromGitlabIfNecessary` to return a cleanup function as well and then defer that. May be a less brittle, but this should work fine regardless.

gabrieljackson (approved):
👍

internal/provisioner/helm_utils.go
gabrieljackson: Yeah, I think this is much better and we won't have to worry about file change collisions anymore which is great.

internal/provisioner/helm_utils.go
gigawhitlocks: Sure! Oh I like that idea a lot actually thanks @gabrieljackson

gabrieljackson (approved):
Nice!

PR #403, opened 2021-01-25
MM-32264 - Switch Provisioner to use Operator beta CR by default

Szymongib:
#### Summary
Switch Provisioner to use `Mattermost` resource instead of `ClusterInstallation` by default.
This change will make all new Installations to be created with `Mattermost` CR but will not migrate existing Installations which will continue to operate.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-32264

#### Release Note

```release-note
Use Mattermost CR instead of ClusterInstallation for new Installation
```


stylianosrigas (approved):
LGTM

gabrieljackson (approved):
Such a big change for us from such a small PR. ;)

PR #406, opened 2021-01-25
Fix v1alpha1 hibernation replica count check

gabrieljackson:
This corrects a small regression for checking replica counts for
v1alpha1 resources when installation hibernation is used.

Fixes https://mattermost.atlassian.net/browse/MM-32305

```release-note
Fix v1alpha1 hibernation replica count check
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -21,6 +21,8 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+const hibernationReplicaCount = -1
+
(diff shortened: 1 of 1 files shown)


Szymongib (approved):
Good catch!
It it the regression tho? We did the exact same check before https://github.com/mattermost/mattermost-cloud/blob/bca3e68eef10fbc1caf68077d5ff6fd5c31df347/internal/supervisor/cluster_installation.go#L245-L247
and I do not see that `spec.Replicas` was modified somewhere before that 🤔

angeloskyratzakos (approved):
LGTM

gabrieljackson:
> It it the regression tho? We did the exact same check before

Yeah, it's probably not worked for a while. Basically since we started checking replicas along with status. It probably broke shortly after I merged the hibernation code in initially.

PR #409, opened 2021-01-31
Improved server startup check

gabrieljackson:
This change includes the following:
 - Removes duplicate AWS environment check on startup.
 - Uses client wrappers on the startup check for all available
   binaries to ensure that there is no mismatch when the clients
   are later rebuilt and used.
 - Logs terraform, kops, and helm versions on startup.

Fixes https://mattermost.atlassian.net/browse/MM-32473

```release-note
Improved server startup check
```


angeloskyratzakos (approved):

Szymongib (approved):
LGTM

PR #410, opened 2021-02-01
Upgrade to kops 1.18.3, terraform 0.12.29 and kubectl version 1.18.15

angeloskyratzakos:
#### Summary
Updating kops version to 1.18.3, terraform 0.12.29 and kubectl version 1.18.15
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28911

#### Release Note

```release-note
updates to use terraform 0.12.29, kops 1.18.3 and kubectl version 1.18.15
```


stylianosrigas (approved):
Yesss finally, LGTM 👌

README.md
```diff
@@ -29,7 +29,7 @@ The following is required to properly run the cloud server.
 1. Install [Go](https://golang.org/doc/install)
 2. Install [Terraform](https://learn.hashicorp.com/terraform/getting-started/install.html) version v0.11.14
```
gabrieljackson, line 30: Looks like we need to update the version recommendation here as well.

gabrieljackson (commented):
Looks good, but there are two binary files checked in with this commit that probably shouldn't be. Let's remove them and then we should be good to go!

README.md
angeloskyratzakos: You are absolutely right! thanks for the suggestion!

angeloskyratzakos:
> Looks good, but there are two binary files checked in with this commit that probably shouldn't be. Let's remove them and then we should be good to go!

ahhh my bad I was checking if the makefile works fine with new versions. I will update the gitignore

gabrieljackson (approved):
👍

PR #411, opened 2021-02-05
Add hibernated domain redirection to nginx ingress

gabrieljackson:
This new redirection ensures that requests for installations that
are hibernated are forwarded to another configurable target. This
new endpoint can perform further processing including waking up
the installation in question.

This new target can be configured per nginx ingress via helm values.

This change also refactors some route 53 logic and contains some
other cleanup.

Fixes https://mattermost.atlassian.net/browse/MM-32530

```release-note
Add hibernated domain redirection to nginx ingress
```

### DIFF
--- a/internal/tools/aws/constants.go
+++ b/internal/tools/aws/constants.go
@@ -132,6 +132,10 @@ const (
 	// zone ID in AWS Route53.
 	DefaultPublicCloudDNSTagValue = "public"
 
+	// HibernatingInstallationResourceRecordIDPrefix is a prefix given to AWS
+	// route53 resource records when the installation it points to is hibernating.
+	HibernatingInstallationResourceRecordIDPrefix = "[hybernating] "
+
 	// CustomNodePolicyName is the name of the custom IAM policy that will be
 	// attached in Kops Instance Profile.
 	CustomNodePolicyName = "cloud-provisioning-node-policy"
--- a/internal/tools/aws/constants.go
+++ b/internal/tools/aws/constants.go
@@ -134,7 +134,7 @@ const (
 
 	// HibernatingInstallationResourceRecordIDPrefix is a prefix given to AWS
 	// route53 resource records when the installation it points to is hibernating.
-	HibernatingInstallationResourceRecordIDPrefix = "[hybernating] "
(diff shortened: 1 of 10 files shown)


internal/tools/aws/constants.go
```diff
@@ -132,6 +132,10 @@ const (
 	// zone ID in AWS Route53.
 	DefaultPublicCloudDNSTagValue = "public"
 
+	// HibernatingInstallationResourceRecordIDPrefix is a prefix given to AWS
+	// route53 resource records when the installation it points to is hibernating.
+	HibernatingInstallationResourceRecordIDPrefix = "[hybernating] "
```
gigawhitlocks, line 137: typo: "hybernating" instead of "hibernating"

gigawhitlocks (approved):

internal/tools/aws/constants.go
gabrieljackson: lol, time for `[hypernating]` installations I guess.

gabrieljackson:
Sorry for the extra change here. I found one spot that we should have been checking resource change counts on and wanted to cover it with a fix before I forgot.

internal/tools/aws/constants.go
```diff
@@ -132,6 +132,10 @@ const (
 	// zone ID in AWS Route53.
 	DefaultPublicCloudDNSTagValue = "public"
 
+	// HibernatingInstallationResourceRecordIDPrefix is a prefix given to AWS
+	// route53 resource records when the installation it points to is hibernating.
+	HibernatingInstallationResourceRecordIDPrefix = "[hibernating] "
```
stylianosrigas, line 137: @gabrieljackson I guess we need this gap here?

stylianosrigas (approved):
Just a nit question.

internal/tools/aws/constants.go
gabrieljackson: I added the gap to the prefix so it would always include a space before the domain name part. You can safely check for `[hibernating]` with or without the space for blackbox.

stylianosrigas: Make sense, thanks :)

PR #412, opened 2021-02-05
Allow installation updates on OwnerID

gabrieljackson:
This change allows for the installation OwnerID value to be
updated in case it is no longer accurate.

This will be used by the cloud plugin import functionality.

Fixes https://mattermost.atlassian.net/browse/MM-32650

```release-note
Allow installation updates on OwnerID
```

### DIFF
--- a/model/installation.go
+++ b/model/installation.go
@@ -35,6 +35,7 @@ type Installation struct {
 	Size                       string
 	Affinity                   string
 	State                      string
+	CRVersion                  string
 	CreateAt                   int64
 	DeleteAt                   int64
 	APISecurityLock            bool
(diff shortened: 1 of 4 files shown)


model/installation.go
```diff
@@ -35,15 +35,14 @@ type Installation struct {
 	Size                       string
 	Affinity                   string
 	State                      string
+	CRVersion                  string
```
gigawhitlocks, line 38: Did you mean to drop the comment & JSON annotation?

gigawhitlocks (approved):
One non-blocking comment (just wanted to draw your attention to that spot just in case I had spotted a mistake)

as long as it's intentional, LGTM

model/installation.go
gabrieljackson: Yep, the migration should have not left any installations with an empty value so best not to hide this in case we get a bug one day. Thanks for checking though!

jwilander (approved):

PR #413, opened 2021-02-09
Improve hibernation logic

gabrieljackson:
This change includes the following:
 - Improve group supervisor logic related to installation updates
   when the group contains hibernating installations.
 - Corrects group status information provided via the API when
   the group contains hibernating installations.
 - Updates the dashboard cloud CLI command to properly present
   installations stats when hibernating installations are present.

Fixes https://mattermost.atlassian.net/browse/MM-32740

```release-note
Improve hibernation logic
```

### DIFF
--- a/internal/store/group.go
+++ b/internal/store/group.go
@@ -81,6 +81,7 @@ func (sqlStore *SQLStore) GetUnlockedGroupsPendingWork() ([]*model.Group, error)
 			Select("ID").
 			From("Installation").
 			Where("GroupID = ?", rawGroup.ID).
+			Where("State = ?", model.InstallationStateStable).
 			Where("(GroupSequence IS NULL OR GroupSequence != ?)", rawGroup.Sequence).
 			Where("DeleteAt = 0").
 			Limit(1)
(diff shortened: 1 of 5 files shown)


internal/store/group.go
```diff
@@ -81,6 +81,7 @@ func (sqlStore *SQLStore) GetUnlockedGroupsPendingWork() ([]*model.Group, error)
 			Select("ID").
 			From("Installation").
 			Where("GroupID = ?", rawGroup.ID).
+			Where("State = ?", model.InstallationStateStable).
```
gigawhitlocks, line 84: Will this preclude marking an Installation as needing work when it is in the e.g. `provisioning-failed` state?

gigawhitlocks (approved):
Just the one question, as long as the answer is "no," LGTM

internal/store/group.go
gabrieljackson: Yeah, this means that installations in other states won't mark the group as needing work. The supervisors should continue to push those installations to stable eventually meaning it will then get picked up by this check at that time.

jwilander (approved):
LGTM

PR #417, opened 2021-02-17
MM-31628 - Implement migration of ClusterInstallations to Mattermost resource 

Szymongib:
#### Summary

The PR implements possibility of automatically migrating `ClusterInstallation` resources to `Mattermost`.
For Provisioner to run the migration `--force-cr-upgrade` flag needs to be set to `true`. The migration will be run if Installation's `CRVersion` does not match the latest version.

During the migration there will be extra pods for each Installation, therefore we should be careful with the load on the clusters and databases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31628

#### Release Note

```release-note
Implement migration from ClusterInstallation to Mattermost resources
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -69,6 +69,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("keep-filestore-data", true, "Whether to preserve filestore data after installation deletion or not.")
 	serverCmd.PersistentFlags().Bool("require-annotated-installations", false, "Require new installations to have at least one annotation.")
 	serverCmd.PersistentFlags().String("gitlab-oauth", "", "If Helm charts are stored in a Gitlab instance that requires authentication, provide the token here and it will be automatically set in the environment.")
+	serverCmd.PersistentFlags().Bool("force-cr-upgrade", false, "If specified installation CRVersions will be updated to the latest version when supervised.")
 }
 
 var serverCmd = &cobra.Command{
@@ -100,6 +101,8 @@ var serverCmd = &cobra.Command{
 
 		requireAnnotatedInstallations, _ := command.Flags().GetBool("require-annotated-installations")
 		model.SetRequireAnnotatedInstallations(requireAnnotatedInstallations)
+
+		forceCRUpgrade, _ := command.Flags().GetBool("force-cr-upgrade")
 
 		allowListCIDRRange, _ := command.Flags().GetStringSlice("allow-list-cidr-range")
 		if len(allowListCIDRRange) == 0 {
@@ -184,6 +187,7 @@ var serverCmd = &cobra.Command{
 			"use-existing-aws-resources":             useExistingResources,
 			"keep-database-data":                     keepDatabaseData,
 			"keep-filestore-data":                    keepFilestoreData,
+			"force-cr-upgrade":                       forceCRUpgrade,
 			"debug":                                  debugMode,
 			"dev-mode":                               devMode,
 		}).Info("Starting Mattermost Provisioning Server")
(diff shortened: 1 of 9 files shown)


internal/provisioner/kops_provisioner_cluster_installation_beta.go
```diff
@@ -414,6 +423,91 @@ func (provisioner *kopsCIBeta) getMattermostCustomResource(cluster *model.Cluste
...
+	cr.Spec.Migrate = true
+	cr, err = ciClient.Update(ctx, cr, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to start migration of the ClusterInstallation")
+	}
+
+	return provisioner.waitForMigration(name, mmClient, logger)
+}
+
+func (provisioner *kopsCIBeta) waitForMigration(name string, mmClient v1beta1.MattermostInterface, logger log.FieldLogger) error {
```
gabrieljackson, line 463: Unfortunately, I think this method of migration may not work well for us. Despite the logic of the migration working perfectly, the wait loop contained here will tie up the ClusterInstallation Supervisor during this period. Because the supervisors all tick and run sequentially this basically ties up the provisioner entirely from processing other work during this time.

I believe we will need to alter this to perform a check and to quickly exit and continue on to other resources if the migration is still underway.

I know that we use similar wait loops elsewhere like in some cluster flows, but the reason we can get away with it there is simply because we separate out our cluster supervisors to only run on certain provisioners and we also don't make cluster changes all that often.

gabrieljackson (changes requested):
This is awesome! I have one main concern with the wait logic which I outline below.

internal/provisioner/kops_provisioner_cluster_installation_beta.go
```diff
@@ -414,6 +423,91 @@ func (provisioner *kopsCIBeta) getMattermostCustomResource(cluster *model.Cluste
...
+		}
+
+		return errors.Wrapf(err, "failed to get cluster installation %s", clusterInstallation.ID)
+	}
+
+	if cr.Spec.Migrate {
+		if cr.Status.Migration.Error != "" {
+			err = errors.Errorf("error while migrating ClusterInstallation: %s", cr.Status.Migration.Error)
+			logger.Error(err)
+			return err
```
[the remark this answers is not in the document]
Szymongib: No, left it here by mistake.

internal/provisioner/kops_provisioner_cluster_installation_beta.go
Szymongib: Yes, I was on the fence on that. I was thinking that running separate instances will give us enough space for other things as we will probably want to run migration slowly anyway, to not to kill databases. Also the migration is one time thing and the alternative implementation that I had in mind was kind of messy as it would introduce more potential states.

I see your point though, and after rethinking that I came to similar conclusion. As you said it would be better to just do the check.

gabrieljackson: I also thought about running separate migration-only provisioners to slowly migrate, but unfortunately due to how work is picked up by provisioners randomly I could see issues with ensuring all installations get migrated. That said, your new commit looks good in that it will allow us to let every provisioner do migrations now. Also, good job doing the that without new installation states.

Regarding overloading databases, I was also concerned about that, but I think we are okay here because we can allow the group rolling update behavior to ensure that we don't go too quickly when we do the migrations.

gabrieljackson (approved):
Great work!

cmd/cloud/server.go
```diff
@@ -246,7 +250,7 @@ var serverCmd = &cobra.Command{
 		}
 		if installationSupervisor {
 			scheduling := supervisor.NewInstallationSupervisorSchedulingOptions(balancedInstallationScheduling, clusterResourceThreshold, clusterResourceThresholdScaleValue)
-			multiDoer = append(multiDoer, supervisor.NewInstallationSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, keepDatabaseData, keepFilestoreData, scheduling, resourceUtil, logger, cloudMetrics))
+			multiDoer = append(multiDoer, supervisor.NewInstallationSupervisor(sqlStore, kopsProvisioner, awsClient, instanceID, keepDatabaseData, keepFilestoreData, scheduling, resourceUtil, logger, cloudMetrics, forceCRUpgrade))
```
gigawhitlocks, line 253: This isn't a comment about this PR, but in general we might want to think about moving all these positional arguments to a struct, lol, 11 positional arguments to `NewInstallationSupervisor` is cumbersome to say the least. It didn't start out this bad but we've been slowly adding arguments to that function and it's getting a bit out of control lol

gigawhitlocks (approved):
This looks great Szymon, it's actually very straightforward. Thanks!

cmd/cloud/server.go
Szymongib: Yeah, that is fair point.

stylianosrigas (approved):

PR #419, opened 2021-02-19
Optimize route53 record lookup

gabrieljackson:
This change simplifies the route53 record lookup logic to no longer
fully check all record pages. Instead, it relies on record ordering
to be correct to more efficiently return results.

Fixes https://mattermost.atlassian.net/browse/MM-33105

```release-note
Optimize route53 record lookup
```


internal/tools/aws/route53.go
```diff
@@ -324,32 +324,40 @@ func (a *Client) deleteCNAME(hostedZoneID, dnsName string, logger log.FieldLogge
 	return nil
 }
 
+// getRecordSetsForDNS returns up to 10 record sets for a given DNS name and
+// relies on API payload ordering behavior to function correctly. Why is this
+// the case, you ask? Well...
+// "Stay Awhile and Listen" - Deckard Cain
```
jwilander, line 330: 🤣

jwilander (approved):

internal/tools/aws/route53.go
stylianosrigas: 🤣 🤣 🤣

stylianosrigas (approved):

internal/tools/aws/route53.go
```diff
@@ -324,32 +324,40 @@ func (a *Client) deleteCNAME(hostedZoneID, dnsName string, logger log.FieldLogge
...
-
-		for _, recordSet := range recordList.ResourceRecordSets {
-			if strings.TrimRight(*recordSet.Name, ".") == dnsName {
-				recordSets = append(recordSets, recordSet)
-			}
-		}
+	recordList, err := a.Service().route53.ListResourceRecordSets(
+		&route53.ListResourceRecordSetsInput{
+			HostedZoneId:    &hostedZoneID,
+			StartRecordName: &dnsName,
```
stylianosrigas, line 346: @gabrieljackson This StartRecordName will change every time with the last of the previous run right?

gabrieljackson: `StartRecordName` used to change multiple times per DNS record lookup to get through all of the record pages. This was the primary reason we hit the rate limit for the API.

No the start record is simply the DNS value we are looking for so if it exists it will be the first entry.

PR #420, opened 2021-02-21
Enhance AWS client with a resource cache

gabrieljackson:
The AWS client used throughout the provisioner now contains a
resource cache to enhance performance and reduce the number of API
calls needed for many tasks. Three AWS resource values are now
stored: environment name, private hosted zone ID, and public hosted
zone ID.

These three values are some of the most commonly used by the AWS
client so caching them does the following:
 - Reduces the number of API calls made when creating or updating
   CNAME records by roughly 80 percent.
 - Eliminates all API calls made when sending out installation
   and cluster installation webhooks.
 - Greatly reduces the number of API calls made during cluster
   provisioning.

These three values should always remain constant in a given AWS
account so they are safe to cache for the lifetime of the client.
This also prepares the provisioner to be able to better support
managing multiple AWS accounts in a single server.

Fixes https://mattermost.atlassian.net/browse/MM-33119

```release-note
Enhance AWS client with a resource cache
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -208,14 +208,12 @@ var serverCmd = &cobra.Command{
 			// https://github.com/aws/aws-sdk-go/blob/99cd35c8c7d369ba8c32c46ed306f6c88d24cfd7/aws/request/retryer.go#L20
 			MaxRetries: sdkAWS.Int(toolsAWS.DefaultAWSClientRetries),
 		}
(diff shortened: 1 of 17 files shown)


stafot (approved):
lgtm

angeloskyratzakos (approved):
Nice one Gabe!

gigawhitlocks (approved):
I love it when a change removes more than double the code it adds. Looks great, Gabe

PR #422, opened 2021-02-24
Removing Helm stable chart - add fluent bit official chart

angeloskyratzakos:
#### Summary
Adding official Fluent-bit helm chart repo 
Removing Helm stable chart

#### Release Note

```release-note
Removing Helm stable chart - replace it with fluent bit official chart
```


internal/provisioner/utility_group.go
```diff
@@ -65,6 +65,7 @@ var helmRepos = map[string]string{
 	"ingress-nginx":        "https://kubernetes.github.io/ingress-nginx",
 	"prometheus-community": "https://prometheus-community.github.io/helm-charts",
 	"bitnami":              "https://charts.bitnami.com/bitnami",
+	"fluent":               "https://fluent.github.io/helm-charts",
```
stylianosrigas, line 68: @angeloskyratzakos maybe remove stable if we don't need it anymore? Can you check if its needed anywhere else?

stylianosrigas (changes requested):

internal/provisioner/utility_group.go
angeloskyratzakos: You are right! removing it as it is not needed any more.

stylianosrigas (approved):
LGTM 🚀

PR #426, opened 2021-03-09
Installation database weighting

gabrieljackson:
This change introduces the concept of database weights to
installations. Installations that are hibernating now reserve less
space in a shared database than other states. This allows for
the spare capacity of multiple hibernating installations to be
allocated to new installations while still offering some protection
if these installations wake up.

Fixes https://mattermost.atlassian.net/browse/MM-33653

```release-note
Installation database weighting
```

### DIFF
--- a/internal/tools/aws/database_multitenant.go
+++ b/internal/tools/aws/database_multitenant.go
@@ -116,6 +116,10 @@ func (d *RDSMultitenantDatabase) updateMultitenantDatabase(store model.Installat
 		return errors.Wrapf(err, "failed to describe the multitenant RDS cluster ID %s", database.ID)
 	}
 
+	return d.updateCounterTagWithCurrentWeight(database, rdsCluster, store, logger)
+}
+
+func (d *RDSMultitenantDatabase) updateCounterTagWithCurrentWeight(database *model.MultitenantDatabase, rdsCluster *rds.DBCluster, store model.InstallationDatabaseStoreInterface, logger log.FieldLogger) error {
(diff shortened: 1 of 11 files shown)


stylianosrigas (approved):
LGTM

internal/tools/aws/database_multitenant.go
```diff
@@ -92,6 +93,45 @@ func (d *RDSMultitenantDatabase) MaxSupportedDatabases() int {
...
+		return errors.Wrap(err, "failed to calculate total database weight")
+	}
+	roundedUpWeight := int(math.Ceil(weight))
+
+	err = d.updateCounterTag(rdsCluster.DBClusterArn, roundedUpWeight)
+	if err != nil {
+		return errors.Wrapf(err, "failed to update tag:counter in RDS cluster ID %s", *rdsCluster.DBClusterIdentifier)
+	}
+
+	logger.Debugf("Multitenant database %s counter value updated to %d", database.ID, roundedUpWeight)
```
Szymongib, line 130: I believe we use the exact same logic in 3 places now. I would consider making it a separate function.

Szymongib (approved):
Looks great!
I have one minor suggestion, but no blocking comments.

internal/tools/aws/database_multitenant.go
gabrieljackson: Thanks for the suggestion. I originally didn't make a helper function for this due to the differences in locking, but I just realized I can at least split out the tag part. Thanks!

PR #427, opened 2021-03-10
[MM-33713] Add validation for rotator when using rotator for cluster upgrades via API

stylianosrigas:
#### Summary
- Add validation for rotator when using rotator for cluster upgrades via API

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33713

#### Release Note
```release-note
- Add validation for rotator when using rotator for cluster upgrades via API
```

### DIFF
--- a/model/cluster_request.go
+++ b/model/cluster_request.go
@@ -167,24 +167,24 @@ func (p *PatchUpgradeClusterRequest) Validate() error {
 
 	if p.RotatorConfig != nil {
 		if p.RotatorConfig.UseRotator == nil {
-			return errors.Errorf("Rotator config use rotator should be set")
+			return errors.Errorf("rotator config use rotator should be set")
 		}
 
 		if *p.RotatorConfig.UseRotator {
 			if p.RotatorConfig.EvictGracePeriod == nil {
-				return errors.Errorf("Rotator config evict grace period should be set")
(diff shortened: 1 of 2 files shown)


stafot (approved):
LGTM. thanks!

model/cluster_request.go
```diff
@@ -165,6 +165,30 @@ func (p *PatchUpgradeClusterRequest) Validate() error {
 		return errors.Errorf("unsupported cluster version %s", *p.Version)
 	}
 
+	if p.RotatorConfig != nil {
+		if p.RotatorConfig.UseRotator == nil {
+			return errors.Errorf("Rotator config use rotator should be set")
```
gabrieljackson, line 170: nit: lowecase `rotatator config use....` for this return error and the ones below.

gabrieljackson (approved):
Good fix! No blockers on my end.

It may make sense to set defaults for some of these values in the future. We do this for other API requests such as creating installations. Either way, good to make sure the validation is correct.

stylianosrigas:
> Good fix! No blockers on my end.
> 
> It may make sense to set defaults for some of these values in the future. We do this for other API requests such as creating installations. Either way, good to make sure the validation is correct.

Yeap I was thinking about and was not sure if we want to enforce it. For now I just set these checks to avoid any upgrade fails.

PR #429, opened 2021-03-15
Add bifrost metric monitoring

gabrieljackson:
This update exposes bifrost metrics for prometheus to scrape and
export for monitoring purposes.

Fixes https://mattermost.atlassian.net/browse/MM-32149

```release-note
Add bifrost metric monitoring
```

### DIFF
--- a/manifests/bifrost/bifrost.yaml
+++ b/manifests/bifrost/bifrost.yaml
@@ -17,12 +17,12 @@ spec:
       - name: bifrost
         command:
         - /bifrost
-        image: mattermost/bifrost:cloud-1
+        image: mattermost/bifrost:v1.0.1
         imagePullPolicy: IfNotPresent
         env:
         - name: BIFROST_SERVICESETTINGS_HOST
           value: "0.0.0.0:8087"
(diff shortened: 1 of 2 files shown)


stylianosrigas (approved):
Niceeee 🚀

manifests/bifrost/bifrost.yaml
```diff
@@ -65,8 +65,13 @@ spec:
     name: bifrost
   ports:
   - port: 80
+    name: bifrost
     protocol: TCP
     targetPort: 8087
+  - port: 8099
+    name: service
```
stylianosrigas, line 72: @gabrieljackson can we rename service here and in the service monitor to be called metrics?

gabrieljackson: I called it service because that is what it is called internally in bifrost as it is used for metrics and health endpoints (and maybe more in the future).

Still want me to change it?

stylianosrigas: Ah if its not dedicated to only metrics lets leave it like that ;)

Szymongib (approved):
LGTM

PR #431, opened 2021-03-16
Update kOps and other tooling

gabrieljackson:
This change includes:
 - Update kOps to 1.19.1
 - Update builds to use Go 1.16.2 and Alpine 3.13.2
 - Update helm to 3.5.3
 - Update kubectl to 1.19.8
 - Add provisioner logic to handle kOps clusters with multiple
   instance groups.

Fixes https://mattermost.atlassian.net/browse/MM-33906 and https://mattermost.atlassian.net/browse/MM-32329

```release-note
Update kOps and other tooling
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,14 +6,14 @@
 ################################################################################
 
 ## Docker Build Versions
-DOCKER_BUILD_IMAGE = golang:1.14.6
-DOCKER_BASE_IMAGE = alpine:3.12
+DOCKER_BUILD_IMAGE = golang:1.16.2
+DOCKER_BASE_IMAGE = alpine:3.13.2
 
 ## Tool Versions
 TERRAFORM_VERSION=0.12.29
-KOPS_VERSION=v1.18.3
-HELM_VERSION=v3.4.1
-KUBECTL_VERSION=v1.18.15
+KOPS_VERSION=v1.19.1
+HELM_VERSION=v3.5.3
+KUBECTL_VERSION=v1.19.8
 
 ################################################################################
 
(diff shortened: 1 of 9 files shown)


gigawhitlocks (approved):

jwilander (approved):

gabrieljackson:
The PR has been updated with logic to handle the new default kops instance group behavior. All cluster flows were tested at least once locally.

Note to SREs: there appears to be some logic in kops that doesn't allow us to set an instance group node count back to 0 when it is greater than 1. This means we will always have a minimum node count of the number of AZs the cluster is in. This shouldn't be an issue, but wanted to note it in case we see this in our accounts. I will see if there are ways around this in a future PR.

Makefile
```diff
@@ -6,14 +6,14 @@
 ################################################################################
 
 ## Docker Build Versions
-DOCKER_BUILD_IMAGE = golang:1.14.6
-DOCKER_BASE_IMAGE = alpine:3.12
+DOCKER_BUILD_IMAGE = golang:1.16.2
+DOCKER_BASE_IMAGE = alpine:3.13.2
 
 ## Tool Versions
 TERRAFORM_VERSION=0.12.29
```
stylianosrigas, line 13: Theoretically we can support terraform 0.13.5 right? Should we give it a go or not in this PR?

stylianosrigas (approved):

Makefile
gabrieljackson: Yep, that right. I was planning on doing that as a follow-up due to this already having a bunch of changes.

PR #432, opened 2021-03-18
MM-33398 - Installation backup API

Szymongib:
#### Summary

The last part of backup functionality for Installations. This PR contains:
- API to manage backup operations
- CLI methods
- Delete functionality for Backup Supervisor
- Minor improvements of other parts

Some notes on deleting backup:
- When backup deletion is requested the backuped data is removed from the file store.
- When deleting installation, deletion of backups will be dependant on `keepFilestoreData` flag, if it is set to `true` backups will be preserved, otherwise they will be deleted.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33398

#### Release Note

```release-note
Installation backup API
```

### DIFF
--- a/internal/tools/aws/s3.go
+++ b/internal/tools/aws/s3.go
@@ -103,3 +103,16 @@ func (a *Client) S3EnsureBucketDirectoryDeleted(bucketName, directory string, lo
 
 	return nil
 }
+
+// S3EnsureObjectDeleted is used to ensure that the file is deleted.
+func (a *Client) S3EnsureObjectDeleted(bucketName, path string) error {
+	_, err := a.Service().s3.DeleteObject(&s3.DeleteObjectInput{
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(path),
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to delete object")
+	}
+
+	return nil
+}
--- a/cmd/cloud/backup.go
+++ b/cmd/cloud/backup.go
@@ -1,0 +1,169 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package main
+
+import (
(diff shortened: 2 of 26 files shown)


internal/tools/aws/s3.go
```diff
@@ -103,3 +103,16 @@ func (a *Client) S3EnsureBucketDirectoryDeleted(bucketName, directory string, lo
...
 	return nil
 }
+
+// S3EnsureObjectDeleted is used to ensure that the file is deleted.
+func (a *Client) S3EnsureObjectDeleted(bucketName, path string) error {
+	_, err := a.Service().s3.DeleteObject(&s3.DeleteObjectInput{
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(path),
+	})
+	if err != nil {
```
gigawhitlocks, line 113: is this error `nil` when the object was previously deleted? I just want to make sure this method does not return an error if you try to use it to delete an object that has already been deleted. 'Ensure' in the name makes it sound like it should not return an error if you try to delete something that has already been deleted.

gigawhitlocks (approved):
Looks great to me; Szymon at this point I'm taking it as a personal challenge to ever find anything in your code that stands out as a problem :joy:

I had only one question which might be highlighting a bug, and it might not be. I suspect it is not, hence the approval.

internal/tools/aws/s3.go
Szymongib: Thanks Ian. I have double checked that and it does not return error if object was already deleted or never existed.

gabrieljackson (approved):
This looks great!

As we discussed earlier, I am still thinking on putting installations being backed up into a new state, but I realize that this should work without it and having the other state could cause issues with things like auto-wakeup from the CWS. Just want to note it though in case we run into an edge case down the road.

stylianosrigas (approved):
Great stuff Szymon 🚀

PR #433, opened 2021-03-24
MM-34229 - Unify paging logic across different models

Szymongib:
#### Summary

As paging logic is the same for every model we can unify it to common struct and remove some duplicated logic in creating and parsing url queries, applying database filters and registering command line arguments.

It will make it easier to make any changes and adding new models.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-34229

#### Release Note

```release-note
Unify paging logic across different models
```


gabrieljackson (approved):
More lines deleted than created. 👍

gigawhitlocks (approved):
This is great and I/we can make use of it in the AWAT as well :tada:

PR #434, opened 2021-03-24
Handle case when migration not started yet

Szymongib:
#### Summary

If the Mattermost Operator is lagging behind the Provisioner migration might not start between Provisioner runs.
This PR provides a way to handle this case.

#### Release Note

```release-note
Handle case when migration not started yet
```


gabrieljackson (approved):

angeloskyratzakos (approved):
LGTM

PR #437, opened 2021-03-26
Respect installation weights when querying multi tenant databases

Szymongib:
#### Summary

Get installation weights when querying for multitenant databases. 

#### Release Note

```release-note
Respect installation weights when querying multi tenant databases
```


gabrieljackson (approved):
👍

PR #441, opened 2021-04-01
Refactor state transitions validation

Szymongib:
#### Summary
This PR is a proposal of small refactoring of states validation logic together with some tests.

Let me know if you like this approach and find it any better than the previous one. If that is not the case, we can close this PR.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/model/cluster_states.go
+++ b/model/cluster_states.go
@@ -83,80 +83,47 @@ var AllClusterRequestStates = []string{
 // ValidTransitionState returns whether a cluster can be transitioned into the
 // new state or not based on its current state.
 func (c *Cluster) ValidTransitionState(newState string) bool {
-	switch newState {
-	case ClusterStateCreationRequested:
-		return validTransitionToClusterStateCreationRequested(c.State)
-	case ClusterStateProvisioningRequested:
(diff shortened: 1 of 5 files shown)


gabrieljackson (approved):
Nice cleanup! I think this is better as well so LGTM.

model/cluster_states.go
```diff
@@ -83,80 +83,47 @@ var AllClusterRequestStates = []string{
...
-		return validTransitionToClusterStateCreationRequested(c.State)
-	case ClusterStateProvisioningRequested:
-		return validTransitionToClusterStateProvisioningRequested(c.State)
-	case ClusterStateUpgradeRequested:
-		return validTransitionToClusterStateUpgradeRequested(c.State)
-	case ClusterStateResizeRequested:
-		return validTransitionToClusterStateResizeRequested(c.State)
-	case ClusterStateDeletionRequested:
-		return validTransitionToClusterStateDeletionRequested(c.State)
+	validStates, found := validClusterTransitions[newState]
```
gigawhitlocks, line 86: :heart_eyes:

gigawhitlocks (approved):
holy crap I love this change I can't believe I didn't approve it already

PR #444, opened 2021-04-19
Deploy pgbouncer helm chart as utility

angeloskyratzakos:
#### Summary
Deploy pgbouncer helm chart as utility.
Deploy pgbouncer configmap and secret with userlist as manifests during the provision of the cluster if they don't exist.
Scrape pgbouncer metrics with prometheus

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34732
https://mattermost.atlassian.net/browse/MM-34733

#### Release Note

```release-note
New pgbouncer deployment as utility 
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -186,6 +186,17 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		return errors.Wrap(err, "unable to attach custom node policy")
 	}
 
+	files := []k8s.ManifestFile{
+		{
+			Path:            "manifests/pgbouncer-manifests/pgbouncer-secret.yaml",
+			DeployNamespace: "pgbouncer",
+		},
+	}
+	err = k8sClient.CreateFromFiles(files)
+	if err != nil {
+		return err
+	}
+
 	ugh, err := newUtilityGroupHandle(kops, provisioner, cluster, awsClient, logger)
 	if err != nil {
 		return err
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -348,19 +348,19 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 
 	_, err = k8sClient.Clientset.CoreV1().Namespaces().Get(ctx, "pgbouncer", metav1.GetOptions{})
 	if k8sErrors.IsNotFound(err) {
-		logger.Info("namespace resource does not exist, will create it...")
-		files = append(files, k8s.ManifestFile{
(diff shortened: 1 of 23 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -345,6 +345,43 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 			DeployNamespace: "kube-system",
 		},
 	}
+
+	_, err = k8sClient.Clientset.CoreV1().Namespaces().Get(ctx, "pgbouncer", metav1.GetOptions{})
+	if k8sErrors.IsNotFound(err) {
+		logger.Info("namespace resource does not exist, will create it...")
```
stylianosrigas, line 351: I think it makes more sense to create the namespace with a API call instead of deploying a manifest. We already do this [here](https://github.com/mattermost/mattermost-cloud/blob/600a375c7076d9f0a17280413d23f8b276783873/k8s/namespaces.go#L33) 1/5 on that.

stylianosrigas (approved):
LGTM, only one comment about namespace creation.
Great work 🚀

internal/provisioner/kops_provisioner_cluster.go
angeloskyratzakos: Thanks @stylianosrigas for your suggestion, I have updated it.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -345,6 +345,43 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 			DeployNamespace: "kube-system",
 		},
 	}
+
+	_, err = k8sClient.Clientset.CoreV1().Namespaces().Get(ctx, "pgbouncer", metav1.GetOptions{})
+	if k8sErrors.IsNotFound(err) {
```
gabrieljackson, line 350: Is there a reason you didn't want to use `CreateOrUpdateNamespace` for this? I has the check in it to see if the namespace exists.

gabrieljackson: A follow up question I have is why is there some pgbouncer provisioning logic hardcoded here? I see that you were able to add pgbouncer as a utility which is awesome, but I am curious why the following change isn't done within that utility code instead?

gabrieljackson (commented):
Looks good, but I have a question.

internal/provisioner/kops_provisioner_cluster.go
angeloskyratzakos: Thank you, @gabrieljackson, for both of your suggestions you were absolutely right, I have updated the code, so the pgbouncer manifests are created alongside with utility and I have used the `CreateOrUpdateNamespace`

gabrieljackson (approved):
Very nice! Thanks for the improvements.

PR #445, opened 2021-04-22
MM 34873 shared vpc feature for kops clusters

Mshahidtaj:
Summary

Shared VPC feature for KOPS clusters. Introduce changes:

- Creating a secondary kops cluster inside the same VPC used by the Primary / existing cluster.
- Deleting a secondary cluster & untagging resources where required.
- Represent VPC information with cloud cluster list command.

Ticket Link

https://mattermost.atlassian.net/browse/MM-34077

Release Note

`Introducing Shared VPC feature in provisioner for KOPS clusters creation.`

Test cases:

`[Testing Shared VPC feature -Dev.pdf](https://github.com/mattermost/mattermost-cloud/files/6355460/Testing.Shared.VPC.feature.-Dev.pdf)
`

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -84,7 +84,12 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 	defer kops.Close()
 
 	var clusterResources aws.ClusterResources
-	if provisioner.params.UseExistingAWSResources {
+	if kopsMetadata.ChangeRequest.VPC != "" && provisioner.params.UseExistingAWSResources {
+		clusterResources, err = awsClient.GetVpcResourcesByVpcID(kopsMetadata.ChangeRequest.VPC, logger)
+		if err != nil {
+			return err
+		}
+	} else if provisioner.params.UseExistingAWSResources {
 		clusterResources, err = awsClient.GetAndClaimVpcResources(cluster.ID, provisioner.params.Owner, logger)
 		if err != nil {
 			return err
@@ -102,7 +107,8 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		clusterResources.WorkerSecurityGroupIDs,
 		allowSSHCIDRS,
 	)
-	if err != nil {
+	// Only release VPC resources, if create cluster request was for the primary cluster.
+	if err != nil && kopsMetadata.ChangeRequest.VPC == "" {
 		releaseErr := awsClient.ReleaseVpc(cluster.ID, logger)
 		if releaseErr != nil {
 			logger.WithError(releaseErr).Error("Unable to release VPC")
@@ -110,7 +116,13 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 
 		return errors.Wrap(err, "unable to create kops cluster")
 	}
-
+	// Tag Public subnets & respective VPC for the secondary cluster.
+	if err == nil && kopsMetadata.ChangeRequest.VPC != "" {
+		err = awsClient.TagResourcesByCluster(clusterResources, cluster.ID, provisioner.params.Owner, logger)
+		if err != nil {
+			return err
+		}
+	}
 	terraformClient, err := terraform.New(kops.GetOutputDirectory(), provisioner.params.S3StateStore, logger)
 	if err != nil {
 		return err
@@ -210,7 +222,7 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 
 	// Start by gathering resources that will be needed later. If any of this
 	// fails then no cluster changes have been made which reduces risk.
-	bifrostSecret, err := awsClient.GenerateBifrostUtilitySecret(cluster.ID, logger)
+	bifrostSecret, err := awsClient.GenerateBifrostUtilitySecret(cluster.ID, cluster.ProvisionerMetadataKops.ChangeRequest.VPC, logger)
 	if err != nil {
 		return errors.Wrap(err, "failed to generate bifrost secret")
 	}
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -107,8 +107,8 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		clusterResources.WorkerSecurityGroupIDs,
 		allowSSHCIDRS,
 	)
-	// Only release VPC resources, if create cluster request was for the primary cluster.
-	if err != nil && kopsMetadata.ChangeRequest.VPC == "" {
+	// release VPC resources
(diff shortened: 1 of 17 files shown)


model/cluster.go
```diff
@@ -26,6 +26,7 @@ type Cluster struct {
 	LockAcquiredBy          *string
 	LockAcquiredAt          int64
 	Networking              string
+	VPC                     string
```
gabrieljackson, line 29: So, I believe we shouldn't set the VPC value here in the cluster object as well as the KopsMetadata. We should only need it in the metadata where the other relevant values are.

I realize that Networking is set in the cluster object as well from your previous PR, but that was a bit of an oversight on my part. It should also probably just be set a single time in the KopsMetadata. Not a big deal, but we should avoid adding duplicates of the same values as it just leads to additional complexity and more change to have the values get out of sync.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -102,15 +107,22 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 		clusterResources.WorkerSecurityGroupIDs,
 		allowSSHCIDRS,
 	)
-	if err != nil {
+	// Only release VPC resources, if create cluster request was for the primary cluster.
+	if err != nil && kopsMetadata.ChangeRequest.VPC == "" {
```
gabrieljackson, line 111: There is an error in this logic. We no longer are just checking for just `err != nil` which means if `err != nil` and `kopsMetadata.ChangeRequest.VPC != ""` then the code just proceeds as if there was no error.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -102,15 +107,22 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
...
 		releaseErr := awsClient.ReleaseVpc(cluster.ID, logger)
 		if releaseErr != nil {
 			logger.WithError(releaseErr).Error("Unable to release VPC")
 		}
 
 		return errors.Wrap(err, "unable to create kops cluster")
 	}
-
+	// Tag Public subnets & respective VPC for the secondary cluster.
+	if err == nil && kopsMetadata.ChangeRequest.VPC != "" {
```
gabrieljackson, line 120: Same here I think.

gabrieljackson (changes requested):
Nice work, but I have outlined some concerns below.

internal/tools/kops/instance_groups.go
```diff
@@ -129,13 +129,17 @@ func (c *Cmd) UpdateMetadata(metadata *model.KopsMetadata) error {
...
+		return err
+	}
 	metadata.AMI = AMI
 	metadata.MasterInstanceType = masterMachineType
 	metadata.MasterCount = masterIGCount
 	metadata.NodeMinCount = nodeMinCount
 	metadata.NodeMaxCount = nodeMaxCount
-	metadata.Networking = networking
+	metadata.Networking = GetCurrentCni(networking)
+	metadata.VPC = strings.Trim(vpc, "\"")
```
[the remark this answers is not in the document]
Mshahidtaj: Yeah , The VPC value was  like this before trim `"VPC": "\"vpc-094c270d675886cbb\""`.
The `\` is the escape char in the above trim, we r just removing double quotes which eventually remove the `\` char as well and the resultant value will be like following `"VPC": "vpc-094c270d675886cbb"`

internal/store/cluster.go
```diff
@@ -73,6 +73,14 @@ func (r *rawCluster) toCluster() (*model.Cluster, error) {
 	if err != nil {
 		return nil, err
 	}
+	if r.Cluster.ProvisionerMetadataKops != nil && r.Cluster.ProvisionerMetadataKops.Networking != "" {
+
```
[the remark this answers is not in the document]
Mshahidtaj: Done. Thanks

internal/store/cluster.go
```diff
@@ -73,6 +73,14 @@ func (r *rawCluster) toCluster() (*model.Cluster, error) {
...
+		r.Cluster.Networking = r.Cluster.ProvisionerMetadataKops.Networking
+	}
+	if r.Cluster.ProvisionerMetadataKops != nil && r.Cluster.ProvisionerMetadataKops.VPC != "" {
+
```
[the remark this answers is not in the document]
Mshahidtaj: Removed. Thanks

model/cluster.go
Mshahidtaj: As per your sugestion , I've removed the VPC field from the cluster object but idea was to represents it in the output of `cloud cluster list` e.g

` "CustomInstanceGroups": {},
     "Networking": "weave",
      "VPC": "vpc-094c270d675886cbb"`

internal/provisioner/kops_provisioner_cluster.go
Mshahidtaj: Sorry, forgot to remove the additional VPC condition here, it is not required any more because `release_vpc` handle the resource release for both primary & secondary cluster. 
modify the condition and did a quick cluster/ delete test. looks fine to me.

internal/provisioner/kops_provisioner_cluster.go
Mshahidtaj: Same err check not required. 
modify the condition and did a quick cluster/ delete test. looks fine to me.

internal/tools/aws/cluster_management.go
```diff
@@ -299,22 +299,43 @@ func (a *Client) claimVpc(clusterResources ClusterResources, clusterID string, o
...
-
-	vpcs, err := a.GetVpcsWithFilters(vpcFilters)
+	vpcs, err := a.GetVpcsWithFilters(secondaryVpcFilters)
 	if err != nil {
 		return err
 	}
+	if len(vpcs) != 0 {
+		//untag keys for secondary cluster
+		isSecondaryCluster = true
+
```
gabrieljackson, line 320: nit: extra newline

gabrieljackson (approved):
Looks good! A few minor issues noted below.

internal/provisioner/nginx.go
```diff
@@ -118,11 +118,16 @@ func (n *nginx) NewHelmDeployment() (*helmDeployment, error) {
...
+	if n.cluster.ProvisionerMetadataKops.ChangeRequest.VPC != "" {
+		clusterResources, err = n.awsClient.GetVpcResourcesByVpcID(n.cluster.ProvisionerMetadataKops.ChangeRequest.VPC, n.logger)
+	} else {
+		clusterResources, err = n.awsClient.GetVpcResources(n.cluster.ID, n.logger)
+	}
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to retrive the VPC information")
 	}
-
+
```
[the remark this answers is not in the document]
Mshahidtaj: Removed.

internal/tools/aws/cluster_management.go
Mshahidtaj: Removed.

stylianosrigas (approved):
LGTM

PR #449, opened 2021-04-28
Add installation database migration model and store

Szymongib:
#### Summary
Another PR for Installation restoration and migration story. This PR introduces:
- Model for `InstallationDBMigrationOperation`.
- Store methods for new model.
- Adds `MigratedInstallations` to `MultitenantDatabase` - The new field will store installations migrated from this database until the logical database still exists on this DB cluster.
- Adds store method to `InstallationBackup` verifying if backup is used by any operation.

#### Release Note

```release-note
Add installation database migration model and store
```

### DIFF
--- a/internal/store/installation_backup.go
+++ b/internal/store/installation_backup.go
@@ -7,6 +7,7 @@ package store
 import (
 	"database/sql"
 	"encoding/json"
+	"fmt"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/mattermost/mattermost-cloud/model"
@@ -94,6 +95,51 @@ func (sqlStore *SQLStore) IsInstallationBackupRunning(installationID string) (bo
 	}
 
 	return ongoingBackups > 0, nil
+}
+
+// IsInstallationBackupBeingUsed checks if backup is being used by any DB restoration Operation or DB migration Operation
+func (sqlStore *SQLStore) IsInstallationBackupBeingUsed(backupID string) (bool, error) {
+	backupsCountBuilder := sq.
+		Select("Count (*)").
+		From(fmt.Sprintf("%s as b", backupTable)).
+		Where("b.ID = ?", backupID).
+		Where("b.DeleteAt = 0")
+
+	builder := backupsCountBuilder.
+		Join(fmt.Sprintf("%s as r on r.BackupID = b.ID", installationDBRestorationTable)).
+		Where(sq.Eq{"r.State": model.AllInstallationDBRestorationStatesPendingWork}).
+		Where("r.DeleteAt = 0")
+	usedBackups, err := sqlStore.getCount(builder)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to count installation backups used by restoration operations")
+	}
+	if usedBackups > 0 {
+		return true, nil
+	}
+
+	builder = backupsCountBuilder.
+		Join(fmt.Sprintf("%s as m on m.BackupID = b.ID", installationDBMigrationTable)).
+		Where(sq.Eq{"m.State": model.AllInstallationDBMigrationOperationsStatesPendingWork}).
+		Where("m.DeleteAt = 0")
+	usedBackups, err = sqlStore.getCount(builder)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to count installation backups used by migration operations")
+	}
+
+	return usedBackups > 0, nil
+}
+
+func (sqlStore *SQLStore) getCount(builder sq.SelectBuilder) (int64, error) {
+	var total countResult
+	err := sqlStore.selectBuilder(sqlStore.db, &total, builder)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to query for count")
+	}
+	count, err := total.value()
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to get count")
+	}
+	return count, nil
 }
 
 // CreateInstallationBackup records installation backup to the database, assigning it a unique ID.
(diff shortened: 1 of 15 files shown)


internal/store/installation_db_migration.go
```diff
@@ -0,0 +1,305 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create migration operation from raw")
+		}
+		migrationOperations = append(migrationOperations, operation)
+	}
+	return migrationOperations, nil
+}
+
+// TODO: we should probably create some intermediary layer to not keep this logic in store.
+// For now tho transactions are not accessible outside the store, therefore it is implemented this way.
```
gabrieljackson, line 90:
```suggestion
// For now though transactions are not accessible outside the store, therefore it is implemented this way.
```

gabrieljackson (approved):
Looks good! I was hoping you could clear something up a bit for me:

> Adds MigratedInstallations to MultitenantDatabase - The new field will store installations migrated from this database until the logical database still exists on this DB cluster.

If I understand correctly then, this is used only while the installation data migration is in progress, but will then be cleaned up?

Szymongib:
> Looks good! I was hoping you could clear something up a bit for me:
> 
> > Adds MigratedInstallations to MultitenantDatabase - The new field will store installations migrated from this database until the logical database still exists on this DB cluster.
> 
> If I understand correctly then, this is used only while the installation data migration is in progress, but will then be cleaned up?

Good question @gabrieljackson! So the idea for it is that it will persist after the initial migration finishes and there will be a separate action required to finally "commit" the migration that will cleanup old logical DB and remove the Installation from `MigratedInstallations`

The purpose behind it is to be able to easily rollback the migration, lets say if we notice that something is not correct when the migration finishes.

Moving forward as we gain confidence that everything works fine, we may move the "commit" part as a final step of migration  but for now I think it will be useful option to be able to revert the migration.

gabrieljackson:
Oh, that makes a lot of sense. Thanks for the explination!

internal/store/installation_backup.go
```diff
@@ -96,6 +97,51 @@ func (sqlStore *SQLStore) IsInstallationBackupRunning(installationID string) (bo
 	return ongoingBackups > 0, nil
 }
 
+// IsInstallationBackupBeingUsed checks if backup is being used by any DB restoration Operation or DB migration Operation
+func (sqlStore *SQLStore) IsInstallationBackupBeingUsed(backupID string) (bool, error) {
+	backupsCountBuilder := sq.
+		Select("Count (*)").
+		From(fmt.Sprintf("%s as b", backupTable)).
```
gigawhitlocks, line 104: Can you tell me why string formatting is being used here instead of the `?` syntax on the other lines?

Szymongib: Those are used to format the table names with aliases, it is static element of the query not an external parameter therefore it cannot be formatted with `?`syntax.

gigawhitlocks: oh OK I understand, I guess the `AS` SQL keyword is new to me, thanks for clearing that up.

internal/store/migrations.go
```diff
@@ -1238,6 +1238,79 @@ var migrations = []migration{
...
+				DeleteAt BIGINT NOT NULL,
+				LockAcquiredBy TEXT NULL,
+				LockAcquiredAt BIGINT NOT NULL
+			);
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = e.Exec(`ALTER TABLE MultitenantDatabase RENAME TO MultitenantDatabaseTemp;`)
```
gigawhitlocks, line 1268: I believe it's possible to just alter this table in place by adding the new column. I don't think you have to rename to a temp table and copy over. This pattern appears in a few places in our migrations but I believe it was erroneously believed at one time that sqlite won't allow you to add a column with ALTER TABLE and so the pattern has proliferated, but I think you can just do ALTER TABLE ADD COLUMN and then skip INSERT INTO on line 1291

It's possible that I'm wrong about this, as I'm pretty sure I've pointed this out in @gabrieljackson 's code before, and now I'm second-guessing my memory about the outcome, but if it's possible to just use ALTER TABLE to add the column without copying the data around I'd prefer that here

gigawhitlocks (approved):
This is awesome as usual Szymon. You've got a couple patterns in this that I should borrow to improve the ImportSupervisor. 

The one comment here is one that I'm not 100% certain about, but I'd _love it_ if you were able to either confirm or invalidate my suggestion. That said: it's a small optimization to a task that runs before server runtime, so it's not super important -- hence the approval.

internal/store/migrations.go
Szymongib: Thanks for checking on that. Yes, you are absolutely right that sqlite allows adding column with ALTER TABLE. The point here is that I want new column to be the same as `InstallationsRaw` as they store data in the same form. This means that the column should be `BYTEA NOT NULL`. This means we can only add column with `DEFAULT` constraint as all existing rows will have to get some value (it cannot be NULL). I do not want to stick with DEFAULT constraint on the column tho so that it we can keep it exactly the same as `InstallationsRaw` and possibly avoid some surprises in the future. The issue with sqlite arises here as it does not support dropping the constraint on the column.

The SQL would look something like that and it does work with Postgres but unfortunately not with sqlite:
```sql
ALTER TABLE MultitenantDatabase ADD COLUMN MigratedInstallationsRaw BYTEA NOT NULL DEFAULT '[]';
ALTER TABLE MultitenantDatabase ALTER COLUMN MigratedInstallationsRaw DROP DEFAULT; -- SQLite does not supprot this one
```

PR #450, opened 2021-04-30
Proposal: give the mock packages unique names

gigawhitlocks:
#### Summary

This is a proposal PR; I noticed something in the codebase that I didn't like and I quickly put together a solution for the problem. If y'all don't think this is a positive change, I won't be too sad, and we can just close this PR.

I noticed while working with mocks in PR #439 that all of the packages of mocked interfaces found under `internal/mocks` are in _separate packages_ but each separate package has the same name,  `mocks`, so if you try to import more than one of them, you're forced to alias them (see `import_tests.go` in the PR for an example of this).

I also noticed that in other packages, besides `supervisor`, the tests that import any of these packages:
1) Only ever use one of them, so the namespace collision isn't noticed
2) Already contain a no-op alias to the name `mocks` in each of their import lines, e.g. https://github.com/mattermost/mattermost-cloud/blob/master/internal/testlib/aws.go#L9

While the aliases like the one on aws.go#L9 do nothing currently, since each of these packages is named `mocks`, they are helpful to this PR, as they prevent me from needing to modify the tests using these packages already.

In this PR, I have given each of the mock packages a unique name.

Renaming these packages allows for two things:
1) You will be able to import without using an alias if you choose, and have a name more descriptive than "`mocks`", which may be desirable in some circumstances and, more importantly
2) There won't be a namespace collision between all of these packages that make them difficult to talk about and also force anyone using more than one of these packages to alias them at import time

#### Ticket Link

No ticket, I just noticed a bunch of packages with the same name and it irritated me.

#### Release Note

```release-note
Gave the mocked interfaces packages unique names
```


gabrieljackson (approved):
👍

Szymongib (approved):

PR #454, opened 2021-05-10
Introduce components common functionalities package

Szymongib:
#### Summary
This PR introduces new package containing common functionalities regarding Provisioner components that are (or will be) used in different places in the project, to make sure functionalities like Triggering Backup or validating database for migration are always working the same and code does not need to be maintain in several places.

Many of those functions will be used by DB Migration supervisor coming in the next PR.

PR changes:
- Introduce `components` package.
- Use new package in Installation Backup API.
- Forbid deleting backup when used by restoration or migration.
- Format code.

#### Release Note

```release-note
Introduce components common functionalities package
```

### DIFF
--- a/internal/api/installation_backup.go
+++ b/internal/api/installation_backup.go
@@ -8,7 +8,7 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/mattermost/mattermost-cloud/internal/components"
+	"github.com/mattermost/mattermost-cloud/internal/common"
 
 	"github.com/mattermost/mattermost-cloud/internal/webhook"
 
@@ -54,15 +54,15 @@ func handleRequestInstallationBackup(c *Context, w http.ResponseWriter, r *http.
 	defer unlockOnce()
 
 	if installationDTO.State != model.InstallationStateHibernating {
-		c.Logger.Error("Cannot request backup for not hibernating installation")
+		c.Logger.Error("Cannot request backup for non-hibernating installation")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	backup, err := components.TriggerInstallationBackup(c.Store, installationDTO.Installation, c.Environment, c.Logger)
+	backup, err := common.TriggerInstallationBackup(c.Store, installationDTO.Installation, c.Environment, c.Logger)
 	if err != nil {
 		c.Logger.WithError(err).Error("Failed to trigger installation backup")
-		w.WriteHeader(components.ErrToStatus(err))
+		w.WriteHeader(common.ErrToStatus(err))
 		return
 	}
 
@@ -170,12 +170,12 @@ func handleDeleteInstallationBackup(c *Context, w http.ResponseWriter, r *http.R
 
 	isUsed, err := c.Store.IsInstallationBackupBeingUsed(backup.ID)
 	if err != nil {
-		c.Logger.WithError(err).Errorf("Failed to check if backup is being used")
+		c.Logger.WithError(err).Error("Failed to check if backup is being used")
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 	if isUsed {
-		c.Logger.Warn("Backup is being used by migration or restoration and cannot ne deleted")
+		c.Logger.Warn("Backup is being used by migration or restoration and cannot be deleted")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
--- a/internal/api/installation_backup.go
+++ b/internal/api/installation_backup.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"time"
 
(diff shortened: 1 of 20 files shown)


internal/api/installation_backup.go
```diff
@@ -51,38 +53,19 @@ func handleRequestInstallationBackup(c *Context, w http.ResponseWriter, r *http.
 	}
 	defer unlockOnce()
 
-	if err := model.EnsureInstallationReadyForBackup(installationDTO.Installation); err != nil {
-		c.Logger.WithError(err).Error("installation cannot be backed up")
+	if installationDTO.State != model.InstallationStateHibernating {
+		c.Logger.Error("Cannot request backup for not hibernating installation")
```
gigawhitlocks, line 57: incredibly minor grammar nit: "non-hibernating installation" is slightly better

gigawhitlocks (approved):
Looks good, just one tiny comment you'll probably find justifiably annoying.
Also, I wonder if a name like `common` might be better than `components`? 0/5 on that but every package contains components while not every package contains _common_ components, so the name _components_ feels a little redundant to me? But I don't feel strongly about it, so take or leave the suggestion as you will.

Oh but also: I love the idea of a common components package and also I love all these clean-up / improvement PRs you've been posting Szymon

internal/api/installation_backup.go
```diff
@@ -187,6 +168,18 @@ func handleDeleteInstallationBackup(c *Context, w http.ResponseWriter, r *http.R
 		return
 	}
 
+	isUsed, err := c.Store.IsInstallationBackupBeingUsed(backup.ID)
+	if err != nil {
+		c.Logger.WithError(err).Errorf("Failed to check if backup is being used")
```
gabrieljackson, line 173:
```suggestion
		c.Logger.WithError(err).Error("Failed to check if backup is being used")
```

internal/api/installation_backup.go
```diff
@@ -187,6 +168,18 @@ func handleDeleteInstallationBackup(c *Context, w http.ResponseWriter, r *http.R
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if isUsed {
+		c.Logger.Warn("Backup is being used by migration or restoration and cannot ne deleted")
```
gabrieljackson, line 178:
```suggestion
		c.Logger.Warn("Backup is being used by migration or restoration and cannot be deleted")
```

internal/components/installation_backup.go
```diff
@@ -0,0 +1,61 @@
...
+
+type installationBackupStore interface {
+	IsInstallationBackupRunning(installationID string) (bool, error)
+	CreateInstallationBackup(backup *model.InstallationBackup) error
+	GetWebhooks(filter *model.WebhookFilter) ([]*model.Webhook, error)
+}
+
+// TriggerInstallationBackup verifies that backup can be started for an Installation and triggers it.
+func TriggerInstallationBackup(store installationBackupStore, installation *model.Installation, env string, logger log.FieldLogger) (*model.InstallationBackup, error) {
+	if err := model.EnsureInstallationReadyForBackup(installation); err != nil {
```
gabrieljackson, line 25: Slight styling nit in that we don't normally condense the err check to one line in `mattermost-cloud`, but if we want to change this guideline I am okay with it. Just want it to be a conscious choice by the team if we do.

internal/components/installation_restoration.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type installationRestorationStore interface {
+	TriggerInstallationRestoration(installation *model.Installation, backup *model.InstallationBackup) (*model.InstallationDBRestorationOperation, error)
+	GetWebhooks(filter *model.WebhookFilter) ([]*model.Webhook, error)
+}
+
+// TriggerInstallationDBRestoration validates, triggers and reports installation database restoration.
+func TriggerInstallationDBRestoration(store installationRestorationStore, installation *model.Installation, backup *model.InstallationBackup, env string, logger log.FieldLogger) (*model.InstallationDBRestorationOperation, error) {
+	if err := model.EnsureInstallationReadyForDBRestoration(installation, backup); err != nil {
```
gabrieljackson, line 23: Same here.

gabrieljackson (approved):

Szymongib:
Thanks @gigawhitlocks!

> Also, I wonder if a name like `common` might be better than `components`?

I was initially thinking that `common` would be quite vague but you are right, it is probably still better than `components` 😄  
I will change it to `common` maybe we will come up with better name in the future.

internal/components/installation_backup.go
Szymongib: Oh, 0/5 on that, I probably just copied it from some PoC I was doing prior. No need to change convention imo.

PR #456, opened 2021-05-12
Add installation recovery command

gabrieljackson:
This new command uses direct store and AWS client connections to
update deleted installation resources so that the provisioning
server will recreate them. Currently, only certain database and
filestore types are supported. In the future, this logic will be
moved directly into the provisioning server and exposed via an API.

Note: I know there is a lot of logic in the command, but I would like
to keep most of it in one spot until we can do a follow-up PR to
move it to the server itself.

Fixes https://mattermost.atlassian.net/browse/MM-35668

```release-note
Add installation recovery command
```

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -476,7 +476,7 @@ var installationRecoveryCmd = &cobra.Command{
 			Paging: model.AllPagesNotDeleted(),
 		}, false, false)
 		if err != nil {
-			return errors.Wrap(err, "failed to get cluster installations filtered by DNS")
+			return errors.Wrap(err, "failed to get installations filtered by DNS")
 		}
 		if len(installations) != 0 {
 			return errors.Errorf("the requested installation's DNS is now in use by %d installations", len(installations))
@@ -539,7 +539,22 @@ var installationRecoveryCmd = &cobra.Command{
 		if !locked {
 			return errors.New("failed to acquire lock on multitenant database")
 		}
-		defer sqlStore.UnlockMultitenantDatabase(db.ID, instanceID, false)
+		defer func() {
+			unlocked, err := sqlStore.UnlockMultitenantDatabase(db.ID, instanceID, false)
+			if err != nil {
+				logger.WithError(err).Error("Failed to unlock multitenant database")
+				return
+			}
+			if unlocked != true {
+				logger.Error("Failed to release lock for multitenant database")
+			}
+		}()
+
+		// Refresh the database object in case updates were made.
+		db, err = sqlStore.GetMultitenantDatabase(databaseID)
+		if err != nil {
+			return errors.Wrap(err, "failed to get database")
+		}
 
 		// Handle follow-up attempts from a partial recovery.
 		// NOTE: this ignores DB weighting.
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -7,9 +7,12 @@ package main
 import (
 	"os"
 
+	sdkAWS "github.com/aws/aws-sdk-go/aws"
+	toolsAWS "github.com/mattermost/mattermost-cloud/internal/tools/aws"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/cloud/installation.go
```diff
@@ -423,3 +433,146 @@ var installationShowStateReport = &cobra.Command{
...
+		})
+		if err != nil {
+			return errors.Wrap(err, "failed to get cluster installations")
+		}
+		if len(clusterInstallations) != 1 {
+			return errors.Errorf("expected to find 1 cluster installation, but found %d", len(clusterInstallations))
+		}
+		clusterInstallation := clusterInstallations[0]
+
+		db, err := sqlStore.GetMultitenantDatabase(databaseID)
```
Szymongib, line 505: nit: We might want to fetch database immediately before locking or even after the lock is acquired. Not sure how long is the operation of restoring secret but due to nature of `Installations` field on multitenant database, if other provisioner added entry to it, we might actually remove it on update.

cmd/cloud/installation.go
```diff
@@ -423,3 +433,146 @@ var installationShowStateReport = &cobra.Command{
...
+		logger.Info("Updating multitenant database")
+
+		locked, err := sqlStore.LockMultitenantDatabase(db.ID, instanceID)
+		if err != nil {
+			return errors.Wrap(err, "failed to lock multitenant database")
+		}
+		if !locked {
+			return errors.New("failed to acquire lock on multitenant database")
+		}
+		defer sqlStore.UnlockMultitenantDatabase(db.ID, instanceID, false)
```
Szymongib, line 542: It is rather unlikely but unlocking can fail, so it would be good to log some warning in that case.

cmd/cloud/installation.go
```diff
@@ -423,3 +433,146 @@ var installationShowStateReport = &cobra.Command{
...
+		}
+
+		// DNS could have been claimed by a new installation, so we need to check
+		// that as well.
+		installations, err := sqlStore.GetInstallations(&model.InstallationFilter{
+			DNS:    installation.DNS,
+			Paging: model.AllPagesNotDeleted(),
+		}, false, false)
+		if err != nil {
+			return errors.Wrap(err, "failed to get cluster installations filtered by DNS")
```
Szymongib, line 479:
```suggestion
			return errors.Wrap(err, "failed to get installations filtered by DNS")
```

Szymongib (approved):
Looks good!
Only thing that we might want to adjust is database locking part.

spirosoik (approved):
great work @gabrieljackson thanks a lot 🎉 . minor comments, no blockers

stylianosrigas (approved):

cmd/cloud/installation.go
```diff
@@ -423,3 +433,146 @@ var installationShowStateReport = &cobra.Command{
...
+		if !locked {
+			return errors.New("failed to acquire lock on multitenant database")
+		}
+		defer sqlStore.UnlockMultitenantDatabase(db.ID, instanceID, false)
+
+		// Handle follow-up attempts from a partial recovery.
+		// NOTE: this ignores DB weighting.
+		if !db.Installations.Contains(installationID) {
+			db.Installations.Add(installationID)
+			err = sqlStore.UpdateMultitenantDatabase(db)
```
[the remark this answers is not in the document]
gabrieljackson: This is actually a style choice that was made early on in this repo. We don't condense the error check.

cmd/cloud/installation.go
gabrieljackson: Ah, thanks for reminding me. I had meant to update that part.

cmd/cloud/installation.go
gabrieljackson: Oh very good point. I wanted to verify all the resources existed before actions were taken so let me refresh the db resource once we have it under lock.

internal/tools/aws/secrets_manager.go
```diff
@@ -215,6 +215,27 @@ func (a *Client) secretsManagerEnsureSecretDeleted(secretName string, logger log
 	return nil
 }
 
+// SecretsManagerRestoreSecret restores a deleted secret.
+func (a *Client) SecretsManagerRestoreSecret(secretName string, logger log.FieldLogger) error {
+	_, err := a.Service().secretsManager.RestoreSecret(&secretsmanager.RestoreSecretInput{
+		SecretId: aws.String(secretName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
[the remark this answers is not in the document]
gabrieljackson: I do like this and will consider it for a follow-up. I went with matching error handling with the other AWS methods for now, but something good to refactor later I think.

internal/store/cluster_installation.go
```diff
@@ -149,6 +149,23 @@ func (sqlStore *SQLStore) DeleteClusterInstallation(id string) error {
 	return nil
 }
 
+// RecoverClusterInstallation recovers a given cluster installation from the deleted state.
+func (sqlStore *SQLStore) RecoverClusterInstallation(clusterInstallation *model.ClusterInstallation) error {
```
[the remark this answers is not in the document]
gabrieljackson: Yeah, we mutate the object and then pass it in to update the database accordingly.

PR #457, opened 2021-05-17
Create k8s license secrets with unique names

gabrieljackson:
This change results in the provisioner creating license secrets
with unique names based on the installation license secret value.
This allows the operator to know that the mattermost app pods
should be rolled to obtain this new value. This also ensures that
external systems can properly wait for license updates to occur
via provisioner webhooks.

Fixes https://mattermost.atlassian.net/browse/MM-35808

```release-note
Create k8s license secrets with unique names
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster_installation.go
+++ b/internal/provisioner/kops_provisioner_cluster_installation.go
@@ -614,13 +614,13 @@ func prepareCILicenseSecret(installation *model.Installation, clusterInstallatio
 	return licenseSecretName, nil
 }
 
+// generateCILicenseName generates a unique license secret name by using a short
+// sha256 hash.
 func generateCILicenseName(installation *model.Installation, clusterInstallation *model.ClusterInstallation) string {
-	sum := fmt.Sprintf("%x", sha256.Sum256([]byte(installation.License)))
-	if len(sum) > 7 {
-		sum = sum[0:6]
-	}
-
-	return fmt.Sprintf("%s-%s-license", makeClusterInstallationName(clusterInstallation), sum)
+	return fmt.Sprintf("%s-%s-license",
+		makeClusterInstallationName(clusterInstallation),
+		fmt.Sprintf("%x", sha256.Sum256([]byte(installation.License))),
+	)
 }
 
 // cleanupOldLicenseSecrets removes an secrets matching the license naming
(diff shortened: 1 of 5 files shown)


internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -580,15 +614,37 @@ func (provisioner *KopsProvisioner) prepareCILicenseSecret(installation *model.I
...
 	}
+
+	return fmt.Sprintf("%s-%s-license", makeClusterInstallationName(clusterInstallation), sum)
+}
+
+// cleanupOldLicenseSecrets removes an secrets matching the license naming
+// convention except the current license secret name. Pass in a blank name value
+// to cleanup all license secrets.
+func cleanupOldLicenseSecrets(currentSecretName string, clusterInstallation *model.ClusterInstallation, k8sClient *k8s.KubeClient, logger log.FieldLogger) error {
+	secrets, err := k8sClient.Clientset.CoreV1().Secrets(clusterInstallation.Namespace).List(context.Background(), metav1.ListOptions{})
```
Szymongib, line 630: I think we could actually use labels to mark secrets with the license, which might be a little cleaner than relying on naming but this should work fine either so 1/5 on that.

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -580,15 +614,37 @@ func (provisioner *KopsProvisioner) prepareCILicenseSecret(installation *model.I
...
 
-func (provisioner *KopsProvisioner) deleteLicenseSecret(clusterInstallation *model.ClusterInstallation, k8sClient *k8s.KubeClient, logger log.FieldLogger) error {
-	secretName := fmt.Sprintf("%s-license", makeClusterInstallationName(clusterInstallation))
-	err := k8sClient.Clientset.CoreV1().Secrets(clusterInstallation.Namespace).Delete(context.Background(), secretName, metav1.DeleteOptions{})
-	if k8sErrors.IsNotFound(err) {
-		logger.Warnf("Secret %s/%s not found, assuming already deleted", clusterInstallation.Namespace, secretName)
-		return nil
+func generateCILicenseName(installation *model.Installation, clusterInstallation *model.ClusterInstallation) string {
+	sum := fmt.Sprintf("%x", sha256.Sum256([]byte(installation.License)))
+	if len(sum) > 7 {
```
Szymongib, line 619: nit: I think it is actually impossible for sha256 to have length different than 32 bytes.

Szymongib (approved):
LGTM!
I would consider using labels to mark license secrets, but leaving it up to you.

internal/provisioner/kops_provisioner_cluster_installation.go
gabrieljackson: Yeah, that was a large amount of extra caution on my part, but that's probably completely unnecessary so let me remove it.

internal/provisioner/kops_provisioner_cluster_installation.go
gabrieljackson: Yeah, agreed. I had to start with this simply because the original license secrets didn't have any labels, but we can definitely move to this in a follow-up.

ethervoid (approved):
LGTM! Really cool stuff :)

PR #464, opened 2021-05-27
Add metrics for prometheus about build info and update installation

spirosoik:
#### Summary
- Include info about what version is running so we can group with each version
- Add installation update tasks historgram and observe duration

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35834

#### Release Note

```release-note
Add build info gauge and update installation duration time histogram for prometheus metrics
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost-cloud/internal/provisioner"
-	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -686,7 +685,6 @@ func (s *InstallationSupervisor) configureInstallationDNS(installation *model.In
 }
 
 func (s *InstallationSupervisor) updateInstallation(installation *model.Installation, instanceID string, logger log.FieldLogger) string {
-	prometheusTimer := s.initialUpdateTasks(logger)
 	// Before starting, we check the installation and group sequence numbers and
 	// sync them if they are not already. This is used to check if the group
 	// configuration has changed during the upgrade process or not.
@@ -801,10 +799,7 @@ func (s *InstallationSupervisor) updateInstallation(installation *model.Installa
 
 	logger.Info("Finished updating clusters installations")
 
-	state := s.waitForUpdateStable(installation, instanceID, logger)
-	prometheusTimer.ObserveDuration()
-
-	return state
+	return s.waitForUpdateStable(installation, instanceID, logger)
 }
 
 func (s *InstallationSupervisor) waitForUpdateStable(installation *model.Installation, instanceID string, logger log.FieldLogger) string {
@@ -832,7 +827,7 @@ func (s *InstallationSupervisor) waitForUpdateStable(installation *model.Install
 
 	logger.Info("Finished updating installation")
 
-	return model.InstallationStateStable
+	return s.finalUpdateTasks(installation, logger)
 }
 
 // Unused stub function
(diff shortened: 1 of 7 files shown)


spirosoik:
The reason for build info:

https://www.robustperception.io/exposing-the-software-version-to-prometheus
https://www.robustperception.io/why-info-style-metrics-have-a-value-of-1

internal/supervisor/installation.go
```diff
@@ -796,7 +798,10 @@ func (s *InstallationSupervisor) updateInstallation(installation *model.Installa
 
 	logger.Info("Finished updating clusters installations")
 
-	return s.waitForUpdateStable(installation, instanceID, logger)
+	state := s.waitForUpdateStable(installation, instanceID, logger)
+	prometheusTimer.ObserveDuration()
```
gabrieljackson, line 802: Unfortunately, this doesn't hold up when we are running multiple provisioners in HA. One provisioner could start the installation update and another one could finish it. Also, I am pretty sure multiple installation updates all occurring at the same time would make this metric incorrect.

gabrieljackson (commented):
The build info part of this is awesome and looks good.

Unfortunately, I don't think the installation update metrics will be valid.

internal/supervisor/installation.go
spirosoik: hmmm I see, the same occurs for creation too right? Or we lock the creation to the provision which started it?

Szymongib: Yes it does.

The lock works in a way that prevents other Provisioners to modify Installation when other instance is already working on it but most likely during one lock the given Provisioner will do only some part of work and release the lock (as for example Operator did not finish spinning up Pods yet or something else is not ready). Then there is no guarantee that different Provisioner won't pick it up and perform next step.

spirosoik: I am asking because I see we use almost the same logic for creation metrics but I will figure it out how we can change this and be safe for HA

spirosoik: @gabrieljackson ^

spirosoik:
@Szymongib @gabrieljackson that's the only way to achieve the right timer actually with a new migration and add the field of `UpdateAt`. The issue is that when we will add this field in practice the data of `UpdateAt = 0` for everything are invalid. The only way to fill in this column is to fetch the k8s deployment, get the datetime and apply it. Not sure if it worths though

spirosoik:
Actually I can implement the same for `DeleteAt`

internal/supervisor/installation.go
```diff
@@ -1361,6 +1361,12 @@ func (s *InstallationSupervisor) finalCreationTasks(installation *model.Installa
 	return model.InstallationStateStable
 }
 
+func (s *InstallationSupervisor) finalUpdateTasks(installation *model.Installation, logger log.FieldLogger) string {
```
Szymongib, line 1364: The Installation is updated several times during the creation.
It starts in state `InstallationStateCreationRequestd` then it goes to `InstallationStateCreationPreProvisioning`, `InstallationStateCreationInProgress` etc.

Not sure if that was your intention but we would basically need to have a metric for each state transition.

Szymongib (commented):
I think the idea with `UpdateAt` could work but we would need to be a little careful with execution.

To go with this approach, I think it might be better to introduce `StateUpdateAt` instead of `UpdateAt`, for example, lock and unlock is technically an update too, but does not change the timestamp.

There are also some other gotchas such as that installation can join/leave the group (triggered by API call) in which case we could not go through the regular update function.

Ideally, to make it work we would need a single code path to update the state (something like the `UpadateInstallationState` method, but it would need to be the only one that is touching the `State`) and the separate one for changing other fields.

I think @gabrieljackson had some ideas about gathering metrics, using webhooks so maybe it is something worth considering as an alternative.

internal/supervisor/installation.go
spirosoik: My intention was to be able to track every update action and setup the base for it. I totally agree that we need to track every state transition as we can find bottlenecks.

spirosoik:
Thanks for the great feedback @Szymongib  and the implementation details. 

> Ideally, to make it work we would need a single code path to update the state (something like the `UpadateInstallationState` method, but it would need to be the only one that is touching the `State`) and the separate one for changing other fields.

I think ideally the state updates could be like immutable events rather than changing a field. So if they were immutable events reduce them we would always have the right state, of course we need to take care the out of order cases. But of course this makes the implementation a bit more complicated with more changes.

Szymongib:
> I think ideally the state updates could be like immutable events rather than changing a field. So if they were immutable events in a table, we would able to reduce them and calculate the right state and have metrics in an easier manner per each state.

Yes, I agree, but as you mentioned it might be overkill, as we would probably need to swap two-thirds of the current Provisioner and probably Postgres too 😄

spirosoik:
> Yes, I agree, but as you mentioned it might be overkill, as we would probably need to swap two-thirds of the current Provisioner and probably Postgres too 😄

Yes it definitely is over-engineering 😄, there are approaches first to try. Let's see @gabrieljackson's input

PR #465, opened 2021-06-09
MM-35892 - Add API for DB migrations

Szymongib:
#### Summary

- Add API for DB migrations, to perform following operations:
  - Trigger DB migration
  - List migration operations
  - Get migration operation
  - Commit migration
  - Rollback migration

Committing migration is synchronous operation that does not involve supervisor to make it simpler. All it does is removing logical database from the previously used DB cluster.

DB migration rollback is asynchronous operation. It works by switching Installation to use old DB cluster, therefore old logical database. All changes made to database after migration will be lost. The rollback cannot be performed if migration was already committed.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35892

#### Release Note

```release-note
Add API for database migrations
```

### DIFF
--- a/internal/api/installation_db_migration.go
+++ b/internal/api/installation_db_migration.go
@@ -213,7 +213,7 @@ func handleCommitInstallationDatabaseMigration(c *Context, w http.ResponseWriter
 	defer unlockOnce()
 
 	if dbMigrationOperation.State != model.InstallationDBMigrationStateSucceeded {
-		c.Logger.Warn("Cannot commit not succeeded DB migration")
+		c.Logger.Warn("Cannot commit DB migration that hasn't succeeded")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
@@ -293,7 +293,7 @@ func handleRollbackInstallationDatabaseMigration(c *Context, w http.ResponseWrit
 func validateDBMigration(c *Context, installation *model.Installation, migrationRequest *model.InstallationDBMigrationRequest, currentDB *model.MultitenantDatabase) error {
 	if migrationRequest.DestinationDatabase != model.InstallationDatabaseMultiTenantRDSPostgres ||
 		installation.Database != model.InstallationDatabaseMultiTenantRDSPostgres {
-		return errors.Errorf("db migration is supported when both source and destination are %q database", model.InstallationDatabaseMultiTenantRDSPostgres)
+		return errors.Errorf("db migration is supported only when both source and destination are %q database types", model.InstallationDatabaseMultiTenantRDSPostgres)
 	}
 
 	if migrationRequest.DestinationMultiTenant == nil {
--- a/internal/api/installation_db_migration.go
+++ b/internal/api/installation_db_migration.go
@@ -1,0 +1,328 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package api
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-cloud/internal/common"
+	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/webhook"
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+)
+
+// initInstallationMigration registers installation migration operation endpoints on the given router.
+func initInstallationMigration(apiRouter *mux.Router, context *Context) {
+	addContext := func(handler contextHandlerFunc) *contextHandler {
+		return newContextHandler(context, handler)
+	}
+
+	migrationsRouter := apiRouter.PathPrefix("/operations/database/migrations").Subrouter()
+
+	migrationsRouter.Handle("", addContext(handleTriggerInstallationDatabaseMigration)).Methods("POST")
+	migrationsRouter.Handle("", addContext(handleGetInstallationDBMigrationOperations)).Methods("GET")
+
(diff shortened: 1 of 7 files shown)


internal/api/installation_db_migration.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	if len(dbMigrations) > 0 {
+		c.Logger.Error("DB migration cannot be started if other successful migration is not committed")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	currentDB, err := c.Store.GetMultitenantDatabaseForInstallationID(installationDTO.ID)
+	if err != nil {
+		c.Logger.WithError(err).Errorf("failed to get current multi-tenant database for installation")
```
gabrieljackson, line 76:
```suggestion
		c.Logger.WithError(err).Error("failed to get current multi-tenant database for installation")
```

internal/api/installation_db_migration.go
```diff
@@ -0,0 +1,328 @@
...
+
+	dbMigrationOperation, status, unlockOnce := lockInstallationDBMigrationOperation(c, migrationID)
+	if status != 0 {
+		w.WriteHeader(status)
+		return
+	}
+	defer unlockOnce()
+
+	if dbMigrationOperation.State != model.InstallationDBMigrationStateSucceeded {
+		c.Logger.Warn("Cannot commit not succeeded DB migration")
```
gabrieljackson, line 216:
```suggestion
		c.Logger.Warn("Cannot commit DB migration that hasn't succeeded")
```

internal/api/installation_db_migration.go
```diff
@@ -0,0 +1,328 @@
...
+	c.Supervisor.Do()
+
+	w.WriteHeader(http.StatusAccepted)
+	outputJSON(c, w, dbMigrationOperation)
+}
+
+func validateDBMigration(c *Context, installation *model.Installation, migrationRequest *model.InstallationDBMigrationRequest, currentDB *model.MultitenantDatabase) error {
+	if migrationRequest.DestinationDatabase != model.InstallationDatabaseMultiTenantRDSPostgres ||
+		installation.Database != model.InstallationDatabaseMultiTenantRDSPostgres {
+		return errors.Errorf("db migration is supported when both source and destination are %q database", model.InstallationDatabaseMultiTenantRDSPostgres)
```
gabrieljackson, line 296:
```suggestion
		return errors.Errorf("db migration is supported only when both source and destination are %q database types", model.InstallationDatabaseMultiTenantRDSPostgres)
```

internal/api/installation_db_migration.go
```diff
@@ -0,0 +1,328 @@
...
+	destinationDB, err := c.Store.GetMultitenantDatabase(migrationRequest.DestinationMultiTenant.DatabaseID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get destination multi-tenant database")
+	}
+	if destinationDB == nil {
+		return errors.Errorf("destination database with id %q not found", migrationRequest.DestinationMultiTenant.DatabaseID)
+	}
+	if destinationDB.DatabaseType != model.DatabaseEngineTypePostgres {
+		return errors.Errorf("destination database is not Postgres")
+	}
```
gabrieljackson, line 312: Isn't this redundant with the earlier check for `InstallationDatabaseMultiTenantRDSPostgres`?

gabrieljackson (approved):
Looks good! Only some minor comments.

internal/api/lock.go
```diff
@@ -150,3 +150,38 @@ func lockInstallationBackup(c *Context, backupID string) (*model.InstallationBac
...
+		c.Logger.WithError(err).Error("failed to lock db migration operation")
+		return nil, http.StatusInternalServerError, nil
+	} else if !locked {
+		c.Logger.Error("failed to acquire lock for db migration operation")
+		return nil, http.StatusConflict, nil
+	}
+
+	unlockOnce := sync.Once{}
+
+	return dbMigrationOperation, 0, func() {
```
gigawhitlocks, line 177: Since this method uses HTTP status codes to signal its error status when it returns, would it make more sense to return an HTTP success code (OK or Accepted) here instead of 0? When reading the call sites of this function, it wasn't clear to me what the 0 return code that we were checking for meant, and I feel that the mix of HTTP status codes with a Unix/C convention (0 for success) is a bit inelegant.

gigawhitlocks (approved):
Looks good! I just had a couple of suggestions and one question, but nothing major. I trust your judgement about what to do in those places after reading my feedback and thinking about it, whether you want to incorporate the suggestions or not, so I'm approving the PR regardless.

internal/api/lock.go
Szymongib: This is the pattern we use in all lock methods in this file, for example https://github.com/mattermost/mattermost-cloud/blob/master/internal/api/lock.go#L37.

I am 0/5 on changing it but I think consistency is more important so if we decide to go for it, we should change all similar lock methods.

internal/api/installation_db_migration.go
```diff
@@ -0,0 +1,325 @@
...
+
+	err := c.Store.TriggerInstallationDBMigrationRollback(dbMigrationOperation, installationDTO.Installation)
+	if err != nil {
+		c.Logger.WithError(err).Error("Failed to trigger db migration rollback")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	unlockOnce()
+	c.Supervisor.Do()
```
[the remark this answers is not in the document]
Szymongib: `Do` will send notification to scheduler's channel that will trigger supervisors to work and it will not receive new messages until they are done, therefore other simultanous call to `Do` will be ignored. You can see details here https://github.com/mattermost/mattermost-cloud/blob/master/internal/supervisor/scheduler.go#L38-L50.

We use this pattern in other API handlers as well https://github.com/mattermost/mattermost-cloud/blob/master/internal/api/installation.go#L231.

PR #468, opened 2021-06-18
Mm 35023 mm apps migration v1

Mshahidtaj:
#### Summary

- Add API(s) & CLI commands for Cluster installation 's migration from source cluster to target cluster. The migration operations can be summarised as  follows:
   - Trigger Cluster installation Migration for all or single cluster id
   - Trigger DNS switch after cluster installation migration
   - Trigger cluster role switch between source & target clusters
   - Trigger Delete inactive cluster installations

The migration process for Cluster installation(s), migrate the installation(s) from source cluster to target cluster. it exposes a seperate end point for each important step to support more control & flexibility , like actual migration from cluster A to B, DNS switch , Cluster tags reset & deleting old cluster installation from source cluster after successful migration.

At a moment logical design is based on sequential steps & little bit on existing installation supervisor but for future enhancement the existing logic can be extended to introduce a new migration supervisor (if required).
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33569

#### Release Note

```release-note
Add API(s) & CLI commands for cluster migration.
```


internal/api/cluster_installation_test.go
```diff
@@ -554,3 +554,299 @@ func TestRunClusterInstallationMattermostCLI(t *testing.T) {
...
+
+	err = sqlStore.CreateClusterInstallation(clusterInstallation2)
+	require.NoError(t, err)
+
+	t.Run("valid migration test", func(t *testing.T) {
+		mcir := &model.MigrateClusterInstallationRequest{ClusterID: sourceClusterID, TargetCluster: targetClusterID, InstallationID: "", DNSSwitch: true, LockInstallation: true}
+		t.Log(mcir)
+		err := client.MigrateClusterInstallation(mcir)
+		require.NoError(t, err)
+	})
```
Szymongib, line 648: This test actually causes server to panic as Cluster with such ID does not exist and the `nil pointer dereference` occurs. We should handle `nil` in the handler. The test succeeded because there is a bug in client method.

model/client.go
```diff
@@ -1184,3 +1184,88 @@ func (c *Client) makeSecurityCall(resourceType, id, securityType, action string)
...
+		return errors.Errorf("failed with status code %d", resp.StatusCode)
+	}
+}
+
+// MigrateDNS requests the migration of cluster installation(s) from the configured provisioning server.
+func (c *Client) MigrateDNS(request *MigrateClusterInstallationRequest) error {
+	resp, err := c.doPost(c.buildURL("/api/cluster_installations/migrate/dns"), request)
+	if err != nil {
+		return nil
+	}
```
Szymongib, line 1210: We should not return `nil` if the error occurred. 
I see it in all new client methods so I assume it was copy & paste issue 🙂

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+
+// handleMigrateClusterInstallations responds to Post /api/cluster_installation/migrate.
+func handleMigrateClusterInstallations(c *Context, w http.ResponseWriter, r *http.Request) {
+	mcir, err := model.NewMigrateClusterInstallationRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to decode cluster migration request", mcir)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	c.Logger = c.Logger.WithField("migration request for cluster installation", mcir)
```
Szymongib, line 376: Not sure if we want to include whole request as a logger field as it is a little clunky:
```
... migration request for cluster installation="{ 12345 67899 false false}"
```
Maybe adding two fields with `source-cluster-id` and `target-cluster-id` would be slightly cleaner.

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+	}
+	if newClusterInstallations == nil {
+		c.Logger.WithError(err).Error("no matching cluster installations found for")
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	// verify that the allows installation is false for the source cluster before migration starts
+	cluster, err := c.Store.GetCluster(mcir.ClusterID)
+	if err != nil {
```
Szymongib, line 468: Unfortunately `cluster` here still can be `nil` even if no error occurred. This happens when such cluster does not exist at all.

We need to handle this case as otherwise the line bellow that references `cluster.AllowInstallation` causes a panic.

internal/store/cluster_installation_test.go
```diff
@@ -564,3 +565,269 @@ func TestDeleteClusterInstallations(t *testing.T) {
...
+	var installations []string
+	installations = append(installations, installation1.ID)
+	installations = append(installations, installation2.ID)
+	newCIsIDs := getClusterInstallationIDs(clusterInstallations)
+	var installationIDs []string
+	for _, ci := range clusterInstallations {
+		installationIDs = append(installationIDs, ci.InstallationID)
+	}
+	err = sqlStore.SwitchDNS(oldCIsIDs, newCIsIDs, installationIDs)
+	require.NoError(t, err)
```
Szymongib, line 741: I think it might be worth it to do a quick check after that, that everything switched correctly be fetching the resources and checking the values that should change.

Szymongib (commented):
Looks good overall!

I have some concerns regarding `DeleteInActiveClusterInstallationByClusterID` that I have described more in depth in the comments.

Also due to a bug in a client code I think that API tests might need some more work after it is fixed.

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+		Paging:         model.AllPagesNotDeleted(),
+	}
+
+	clusterInstallations, err := c.Store.GetClusterInstallations(filter)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to query cluster installations")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if clusterInstallations == nil {
```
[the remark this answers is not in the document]
Mshahidtaj: Done , still overwriting java footprint :) where using length func with null object create `NullPointerException.`

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+	} else {
+		c.Logger.WithError(err).Error("allow installation must be set to false for the source cluster.")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
Mshahidtaj: Done. Thank you

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+			c.Logger.Error("failed to migrate DNS records")
+			w.WriteHeader(status)
+			return
+		}
+	} else {
+		c.Logger.WithError(err).Error("allow installation must be set to false for the source cluster.")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
Mshahidtaj: Done. Thank you

internal/api/cluster_installation_test.go
Mshahidtaj: Good point @Szymongib , made the required the changes. Thank you.

model/client.go
Mshahidtaj: Yeah, it was following the existing pattern :) . Fixed.

internal/api/cluster_installation.go
Mshahidtaj: Done. Thanks

internal/store/cluster_installation_test.go
Mshahidtaj: Added some verification steps.

internal/api/cluster_installation.go
Mshahidtaj: Yes, Good point. fixed

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+		InstallationID: mcir.InstallationID,
+		Paging:         model.AllPagesNotDeleted(),
+	}
+	clusterInstallations, err := c.Store.GetClusterInstallations(filter)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to query cluster installations")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if clusterInstallations == nil {
```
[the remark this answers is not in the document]
Mshahidtaj: Done.Thanks

internal/api/cluster_installation.go
```diff
@@ -359,3 +364,240 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+	c.Logger = c.Logger.WithField("clusterID", clusterID)
+	if len(clusterID) == 0 {
+		c.Logger.Error("missing mandatory source cluster in a migration request")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	// Deleting multiple inactive cluster installations
+	c.Logger.Infof("deleting inactive cluster installations for cluster ID %s", clusterID)
+	err := c.Store.DeleteInActiveClusterInstallationByClusterID(clusterID)
```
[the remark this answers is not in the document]
Mshahidtaj: Thanks for letting me know. made the required changes.

cmd/cloud/cluster_installation.go
```diff
@@ -245,3 +273,90 @@ var clusterInstallationMattermostCLICmd = &cobra.Command{
...
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		cluster, _ := command.Flags().GetString("source-cluster")
+		targetcluster, _ := command.Flags().GetString("target-cluster")
+		installation, _ := command.Flags().GetString("installation")
+
+		client.MigrateClusterInstallation(&model.MigrateClusterInstallationRequest{ClusterID: cluster, TargetCluster: targetcluster, InstallationID: installation, DNSSwitch: false, LockInstallation: false})
```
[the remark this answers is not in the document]
Mshahidtaj: Good Point, As discussed , raise the ticket to incorporate the change https://mattermost.atlassian.net/browse/MM-36660

cmd/cloud/cluster_installation.go
```diff
@@ -245,3 +273,90 @@ var clusterInstallationMattermostCLICmd = &cobra.Command{
...
+		command.SilenceUsage = true
+
+		serverAddress, _ := command.Flags().GetString("server")
+		client := model.NewClient(serverAddress)
+
+		sourceCluster, _ := command.Flags().GetString("source-cluster")
+		targetcluster, _ := command.Flags().GetString("target-cluster")
+		installation, _ := command.Flags().GetString("installation")
+
+		client.MigrateClusterInstallation(&model.MigrateClusterInstallationRequest{SourceClusterID: sourceCluster, TargetClusterID: targetcluster, InstallationID: installation, DNSSwitch: false, LockInstallation: false})
```
Szymongib, line 291: I think I mentioned it earlier but we should handle errors in CLI (there are several places):
```suggestion
		err := client.MigrateClusterInstallation(&model.MigrateClusterInstallationRequest{SourceClusterID: sourceCluster, TargetClusterID: targetcluster, InstallationID: installation, DNSSwitch: false, LockInstallation: false})
		if err != nil {
		         return err
		}
```

Szymongib (commented):
Looks good! Thanks for making the changes.
There are couple more not handled errors that I think is important to do so, the rest are some suggestions for handling less common cases that may occur.

One more thing is that we try to stick to convention of log messages to start with capital letter, so this:
```
c.Logger.WithError(err).Error("Failed to something")
```
instead of that:
```
c.Logger.WithError(err).Error("failed to something")
```

internal/store/cluster_installation.go
```diff
@@ -166,6 +164,23 @@ func (sqlStore *SQLStore) RecoverClusterInstallation(clusterInstallation *model.
...
+func (sqlStore *SQLStore) DeleteInActiveClusterInstallationByClusterID(clusterID string) error {
+	_, err := sqlStore.execBuilder(sqlStore.db, sq.
+		Update("ClusterInstallation").
+		Set("State", model.ClusterInstallationStateDeletionRequested).
+		Where("ClusterID = ?", clusterID).
+		Where("IsActive = ?", false).
+		Where("DeleteAt = 0"),
+	)
+	if err != nil {
+		return errors.Wrap(err, "failed to mark inactive cluster installation as deleted")
```
Mshahidtaj, line 178: @Szymongib Do we want to apply "initial capital" for these error messages as well ? Currently all the existing code is following the "init lower case" pattern. I'm afraid that making the new methods with "init upper case"  will create inconsistency unless we r determined to migrate all the existing error messages to "init upper case" gradually .

Szymongib: No, the errors should stay lower case, only the log messages should start with upper.

model/client.go
```diff
@@ -1289,7 +1289,7 @@ func (c *Client) MigrateClusterInstallation(request *MigrateClusterInstallationR
 		return nil
 
 	default:
-		return errors.Errorf("failed with status code %d", resp.StatusCode)
+		return errors.Errorf("Failed with status code %d", resp.StatusCode)
```
Szymongib, line 1292: I might have not been clear enough in my previous comment. The errors should stay lowercase, only the logged messages start with upper. Sorry for confusion.

Szymongib (approved):
Great work implementing all that and adjusting on the go!

internal/api/cluster_installation.go
```diff
@@ -359,3 +367,269 @@ func handleRunClusterInstallationMattermostCLI(c *Context, w http.ResponseWriter
...
+func handleSwitchClusterRoles(c *Context, w http.ResponseWriter, r *http.Request) {
+	mcir, err := model.NewMigrateClusterInstallationRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithError(err).Error("Failed to decode cluster migration request")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	c.Logger = c.Logger.WithField("cluster_installation", mcir)
+
+	err = c.AwsClient.SwithClusterTags(mcir.SourceClusterID, mcir.TargetClusterID, c.Logger)
```
stylianosrigas, line 609:
```suggestion
	err = c.AwsClient.SwitchClusterTags(mcir.SourceClusterID, mcir.TargetClusterID, c.Logger)
```

internal/api/context.go
```diff
@@ -106,6 +113,11 @@ type Provisioner interface {
 	GetClusterResources(*model.Cluster, bool) (*k8s.ClusterResources, error)
 }
 
+// AwsClient describes the interface required to communicate with the AWS
+type AwsClient interface {
+	SwithClusterTags(clusterID string, targetClusterID string, logger logrus.FieldLogger) error
```
stylianosrigas, line 118:
```suggestion
	SwitchClusterTags(clusterID string, targetClusterID string, logger logrus.FieldLogger) error
```

internal/supervisor/installation_test.go
```diff
@@ -440,6 +440,9 @@ func (a *mockAWS) GetVpcResourcesByVpcID(vpcID string, logger log.FieldLogger) (
 func (a *mockAWS) TagResourcesByCluster(clusterResources aws.ClusterResources, clusterID string, owner string, logger log.FieldLogger) error {
 	return nil
 }
+func (a *mockAWS) SwithClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error {
```
stylianosrigas, line 443:
```suggestion
func (a *mockAWS) SwitchClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error {
```

internal/tools/aws/client.go
```diff
@@ -79,6 +79,7 @@ type AWS interface {
 
 	GetVpcResourcesByVpcID(vpcID string, logger log.FieldLogger) (ClusterResources, error)
 	TagResourcesByCluster(clusterResources ClusterResources, clusterID string, owner string, logger log.FieldLogger) error
+	SwithClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error
```
stylianosrigas, line 82:
```suggestion
	SwitchClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error
```

internal/tools/aws/cluster_management.go
```diff
@@ -409,3 +409,24 @@ func (a *Client) TagResourcesByCluster(clusterResources ClusterResources, cluste
 	}
 	return nil
 }
+
+// SwithClusterTags after migration.
```
stylianosrigas, line 413:
```suggestion
// SwitchClusterTags after migration.
```

internal/tools/aws/cluster_management.go
```diff
@@ -409,3 +409,24 @@ func (a *Client) TagResourcesByCluster(clusterResources ClusterResources, cluste
...
+func (a *Client) SwithClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error {
```
stylianosrigas, line 414:
```suggestion
func (a *Client) SwitchClusterTags(clusterID string, targetClusterID string, logger log.FieldLogger) error {
```

stylianosrigas (approved):
LGTM, a small typo.

gabrieljackson (commented):
Great work! A few minor comments, but one main question about DNS record cleanup.

internal/supervisor/installation.go
```diff
@@ -637,8 +637,10 @@ func (s *InstallationSupervisor) waitForCreationStable(installation *model.Insta
 }
 
 func (s *InstallationSupervisor) configureInstallationDNS(installation *model.Installation, instanceID string, logger log.FieldLogger) string {
+	isActive := true
 	clusterInstallations, err := s.store.GetClusterInstallations(&model.ClusterInstallationFilter{
 		InstallationID: installation.ID,
+		IsActive:       &isActive,
```
gabrieljackson, line 643: Here specifically is where I see us creating the new DNS record, but is the old record ever cleaned up?

gabrieljackson (approved):
I actually just refreshed my memory on the DNS logic and see that it's a create or change which should work!

All the other comments are optional so LGTM.

internal/supervisor/installation.go
Mshahidtaj: Good Question , here we r updating the existing DNS record with new load balancer value.

PR #469, opened 2021-06-22
Add CLI commands for migration and restoration

Szymongib:
#### Summary

This PR introduces new commands to manage Installation DB migrations and restorations as well as provides new flags to `server` command to start supervisors.
The commands are under following subcommands:
- `cloud installation operation db-migration ...`
- `cloud installation operation restoration ...`

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-35892

#### Release Note

```release-note
Add CLI commands for Installation DB migration and restoration
```


cmd/cloud/backup.go
```diff
@@ -5,9 +5,10 @@
 package main
 
 import (
-	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
 	"os"
 
+	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
+
```
gigawhitlocks, line 11: I think this is an extraneous newline, that or the import on line 12 is supposed to be part of the stanza here so that it's grouped with this internal library

cmd/cloud/installation_operation.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "github.com/spf13/cobra"
+
+func init() {
+	installationOperationCmd.AddCommand(installationRestorationOperationCmd)
+	installationOperationCmd.AddCommand(installationDBMigrationOperationCmd)
+}
+
+var installationOperationCmd = &cobra.Command{
+	Use:   "operation",
```
gigawhitlocks, line 15: `operation` is a pretty vague/general term for this family of commands. Please correct me if this is not the correct understanding of usage, e.g. :

`cloud installation operation list`

I don't think that this command makes it clear what kind of operation the user is listing. Maybe `migration` is a better sub-command for this family?

`cloud installation migration list`

or `cloud installation db-migration` ?

gigawhitlocks (approved):
Very minor comments, one's purely style, the other is just a naming suggestion. Good stuff!

cmd/cloud/installation_operation.go
Szymongib: The examples are in PR description. The `operation` is just subcommand grouping other commands that reference operations, so it is like that:
```
cloud installation operation db-migration ...
cloud installation operation restoration ...
```

cmd/cloud/backup.go
Szymongib: Apparently both variations are correct from `goimports` perspective. I removed the new line.

gabrieljackson (approved):
LGTM!

cmd/cloud/installation_operation.go
gigawhitlocks: So yes, I did understand the usage. "Operation" is very generic and I don't understand its value in the command structure. The resulting command doesn't make much sense to me; everything under `installation` is already an operation. These are db operations; shouldn't the command they are grouped under communicate that?

Szymongib: Thanks for the clarification Ian.

So I went with the `operation` subcommand as from the Provisioner side those operations are different than regular create, list, delete, etc. There are actually structs called `DBMigrationOperation` and `DBRestorationOperation` and data for them is preserved in DB. Those are returned as part of commands, they also work in a similar manner therefore I wanted to group them under a single concept of `operation`.

In the example of `cloud installation operation db-migration list` I find it easier to reason about it this way as I am listing `operations` not `db migrations`. 

I think that in the use-case of Provisioner where users of CLI are basically developers/SREs it is better to not abstract that as much as one would in the case of a tool designed to hide the complexity from the user and keep it closer to the internal mechanism as I think it is easier then to understand the system from the perspective of developer (you do not need to traverse different levels of abstractions). 

That said if that would be a more client-facing application I agree with your suggestion.

Anyway, I am 1/5 on that. I am going to move forward with the PR, as it's there for a while but I am open to discussing it further and potentially changing.

PR #476, opened 2021-07-11
Add multitenant database update API endpoint

gabrieljackson:
The new endpoint allows for updating `MaxInstallationsPerLogicalDatabase`
on proxy databases. This also includes tests for the existing database
list endpoint.

Fixes https://mattermost.atlassian.net/browse/MM-36949

```release-note
Add multitenant database update API endpoint
```

### DIFF
--- a/internal/api/databases.go
+++ b/internal/api/databases.go
@@ -64,19 +64,19 @@ func handleUpdateDatabase(c *Context, w http.ResponseWriter, r *http.Request) {
 	databaseID := vars["database"]
 	c.Logger = c.Logger.WithField("database", databaseID)
 
-	database, status, unlockOnce := lockDatabase(c, databaseID)
-	if status != 0 {
-		w.WriteHeader(status)
-		return
-	}
-	defer unlockOnce()
-
 	patchDatabaseRequest, err := model.NewPatchDatabaseRequestFromReader(r.Body)
 	if err != nil {
 		c.Logger.WithError(err).Error("failed to decode request")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
+
+	database, status, unlockOnce := lockDatabase(c, databaseID)
+	if status != 0 {
+		w.WriteHeader(status)
+		return
+	}
+	defer unlockOnce()
 
 	if patchDatabaseRequest.Apply(database) {
 		err = c.Store.UpdateMultitenantDatabase(database)
@@ -90,6 +90,6 @@ func handleUpdateDatabase(c *Context, w http.ResponseWriter, r *http.Request) {
 	unlockOnce()
 
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusAccepted)
+	w.WriteHeader(http.StatusOK)
 	outputJSON(c, w, database)
 }
--- a/internal/api/databases.go
+++ b/internal/api/databases.go
@@ -17,8 +17,11 @@ func initDatabases(apiRouter *mux.Router, context *Context) {
 		return newContextHandler(context, handler)
 	}
 
-	databaseRouter := apiRouter.PathPrefix("/databases").Subrouter()
-	databaseRouter.Handle("", addContext(handleGetDatabases)).Methods("GET")
+	databasesRouter := apiRouter.PathPrefix("/databases").Subrouter()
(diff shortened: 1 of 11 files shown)


internal/api/databases.go
```diff
@@ -53,3 +56,40 @@ func handleGetDatabases(c *Context, w http.ResponseWriter, r *http.Request) {
...
+	defer unlockOnce()
+
+	patchDatabaseRequest, err := model.NewPatchDatabaseRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to decode request")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if patchDatabaseRequest.Apply(database) {
```
Szymongib, line 81: I think it would be useful to indicate to the user whether the patch has been applied or not, or perhaps even return an error when trying to change the MaxInstallationPerDatabase value for DB other than PgBouncer one. It is all good while we remember it but might lead to some confusion in the future 🙂

internal/api/databases.go
```diff
@@ -53,3 +56,40 @@ func handleGetDatabases(c *Context, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	defer unlockOnce()
+
+	patchDatabaseRequest, err := model.NewPatchDatabaseRequestFromReader(r.Body)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to decode request")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
```
Szymongib, line 79: nit: Maybe we could switch those two around to eliminate unnecessary locking if the request body is invalid. I think we do it the other way around in different places too.

internal/api/databases.go
```diff
@@ -53,3 +56,40 @@ func handleGetDatabases(c *Context, w http.ResponseWriter, r *http.Request) {
...
+			c.Logger.WithError(err).Error("failed to update database")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+	}
+
+	unlockOnce()
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusAccepted)
```
Szymongib, line 93: Should it return `202` or just `200`? I do not see any async process that we start, so I would be in favor of 200.

Szymongib (approved):
Looks good!

Thanks for adding tests!

internal/api/databases.go
gabrieljackson: Yep, I think that makes more sense as well. I was mimicking the UpdateCluster endpoint which also has no async processes, but returns the same status. We should probably standardize that one at a later time if we can do so safely.

internal/api/databases.go
gabrieljackson: Yeah, this is a good idea and one I was also wondering if we should do someday. This type of patch logic is very common so I am thinking we design a general improvement and make sure all API endpoints get switched over at the same time.

internal/api/lock.go
```diff
@@ -185,3 +185,38 @@ func lockInstallationDBMigrationOperation(c *Context, operationID string) (*mode
...
+		c.Logger.WithError(err).Error("failed to lock database")
+		return nil, http.StatusInternalServerError, nil
+	} else if !locked {
+		c.Logger.Error("failed to acquire lock for database")
+		return nil, http.StatusConflict, nil
+	}
+
+	unlockOnce := sync.Once{}
+
+	return database, 0, func() {
```
gigawhitlocks, line 212: Is there a reason that when we use this pattern we don't use HTTP StatusOK for the middle return code? I know that `0` is "OK" according to UNIX conventions, but the other values in this method returned as the second return value are HTTP codes -- shouldn't we stick to that pattern?

I saw another function in another PR somewhat recently with this mix of conventions and I called it out, but I don't remember what the outcome of that was.

gigawhitlocks (approved):
LGTM, just one semantic complaint

internal/api/lock.go
gabrieljackson: Yeah, I stuck with `0` as that is how all the other lock helpers are set up as well. Totally okay with updating them all to be something different in the future, but just want to keep the standard till then.

PR #477, opened 2021-07-12
Set `binary_parameters=yes` for PGBouncer databases

gabrieljackson:
This appears to correct an issue with PGBouncer transaction pooling
where errors were sometimes encountered on PG prepared statements.
The result of the change is that all query parameters will now be
sent as a binary when using the PGBouncer proxy.

Understanding of the issue as well as the fix was all due to this awesome
blog post: https://blog.bullgare.com/2019/06/pgbouncer-and-prepared-statements

Reviewers please note this callout:
> The only problem we’ve run into in a golang application is about saving JSONb parameters.

I am hoping this change works out for us and that there is no known conflict
with this in product.

Fixes https://mattermost.atlassian.net/browse/MM-36988

```release-note
Set `binary_parameters=yes` for PGBouncer databases
```

### DIFF
--- a/internal/tools/aws/helpers_sql.go
+++ b/internal/tools/aws/helpers_sql.go
@@ -43,13 +43,18 @@ func MattermostPostgresConnStrings(schema, username, password string, dbCluster 
 
 // MattermostPostgresPGBouncerConnStrings formats the connection strings used by
 // Mattermost servers to access a PostgreSQL database with a PGBouncer proxy.
-func MattermostPostgresPGBouncerConnStrings(username, password, database string) (string, string) {
-	dbConnection := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s?connect_timeout=10&sslmode=disable",
+//
+// Regarding binary_parameters:
+// https://blog.bullgare.com/2019/06/pgbouncer-and-prepared-statements
+func MattermostPostgresPGBouncerConnStrings(username, password, database string) (string, string, string) {
+	dbConnection := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s?connect_timeout=10&sslmode=disable&binary_parameters=yes",
 		username, password, database)
-	readReplicas := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s-ro?connect_timeout=10&sslmode=disable",
+	readReplicas := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s-ro?connect_timeout=10&sslmode=disable&binary_parameters=yes",
+		username, password, database)
+	connectionCheck := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s?connect_timeout=10&sslmode=disable",
 		username, password, database)
 
-	return dbConnection, readReplicas
+	return dbConnection, readReplicas, connectionCheck
 }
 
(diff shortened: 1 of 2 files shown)


jwilander (approved):

agnivade:
Ahh .. this is one of the problems of using an outdated library. Preparing and executing queries _every single time_ without using a statement cache is just bad design. We thought of using pgx sometime back, but the pgx interface is only recommended to be used when postgres is the _only_ driver being used in the application. But even with a statement cache, the first statement will always fail, so we would need some custom code (like wrapping the first `PREPARE` in a transaction) to make this work.

Overall, this is fine. I tested with our [codebase](https://github.com/mattermost/mattermost-server/pull/17908) and seems to work fine. But we plan to use the `jsonb` data type very soon in v6. But that has it's own set of problems with gorp because gorp doesn't support JSON. :angry: 

In short, to make jsonb work with `binary_parameters`, we need to cast a byte slice into a custom type.

```go
type MyJSON []byte

func (val MyJSON) Value() (driver.Value, error) {
	return append([]byte{0x01}, val...), nil
}
```

And we also need to add the query param in our default data_source and make an announcement in the dev meeting to update everyone's dev config. So that prod and dev remains as consistent as possible, and we can surface bugs quicker. @gabrieljackson - would you mind doing the necessary things?

/cc @streamer45 @isacikgoz 

Btw, I noticed the branch name :wink:

internal/tools/aws/helpers_sql.go
```diff
@@ -43,13 +43,18 @@ func MattermostPostgresConnStrings(schema, username, password string, dbCluster
...
 // Mattermost servers to access a PostgreSQL database with a PGBouncer proxy.
-func MattermostPostgresPGBouncerConnStrings(username, password, database string) (string, string) {
-	dbConnection := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s?connect_timeout=10&sslmode=disable",
+//
+// Regarding binary_parameters:
+// https://blog.bullgare.com/2019/06/pgbouncer-and-prepared-statements
+func MattermostPostgresPGBouncerConnStrings(username, password, database string) (string, string, string) {
+	dbConnection := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s?connect_timeout=10&sslmode=disable&binary_parameters=yes",
+		username, password, database)
+	readReplicas := fmt.Sprintf("postgres://%s:%s@pgbouncer.pgbouncer:5432/%s-ro?connect_timeout=10&sslmode=disable&binary_parameters=yes",
```
agnivade, line 52: Strictly speaking, you don't need this for reading. But fine to keep it consistent. Maybe good to refactor the entire thing as a net.URL and stringify it to reduce duplication. 0/5

agnivade (approved):

lieut-data (approved):
Thanks for looping me in! Looking forward to using `JSONB` in the future, but can confirm we're not relying on it in Playbooks right now -- only "vanilla" `JSON` for Postgres.

gabrieljackson:
>  I tested with our codebase and seems to work fine.

Thanks so much for doing this @agnivade and for the additional context. I am happy to announce this to the rest of engineering and make the default data_source change, but we want to do some additional validation testing to ensure we have a clear path forward with PGBouncer before we fully commit. Once that is complete I will take care of these tasks.

> Btw, I noticed the branch name

😉 

Thanks everyone else for the review!

agnivade:
Sounds good @gabrieljackson. Nice work!

PR #479, opened 2021-07-15
Add kubecost helm chart

stafot:
Issue: MM-36940

#### Summary
Adds kubecost helm-chart to have more granular costs allocation.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36940

#### Release Note

```release-note
Adds kubecost helm-chart to have more granular costs allocation.

```

PR #480, opened 2021-07-19
[MM-36314] Add support for Stackrox and remove local utility Helm values dependency.

stylianosrigas:
#### Summary
- Add support for Stackrox 
- Remove local utility Helm values dependency.

#### Release Note

```release-note
- Add support for Stackrox 
- Remove local utility Helm values dependency.
```


stafot:
Need similar fix as the one applied on https://github.com/mattermost/mattermost-cloud/pull/479/commits/d782325a9c4a39bc4fef3c58a20043d74d146f38 to pass tests.

stafot:
Need similar fix as the one applied on https://github.com/mattermost/mattermost-cloud/pull/479/commits/d782325a9c4a39bc4fef3c58a20043d74d146f38 to pass tests.

stylianosrigas:
> Need similar fix as the one applied on [d782325](https://github.com/mattermost/mattermost-cloud/commit/d782325a9c4a39bc4fef3c58a20043d74d146f38) to pass tests.

It was another part @stafot. The logic is different now as we handle the utilities remotely.

spirosoik (approved):
LGTM

stafot (approved):
Great work! If this merged before kubecost I will refactor https://github.com/mattermost/mattermost-cloud/pull/479 accordingly.

cmd/cloud/server.go
```diff
@@ -136,6 +137,13 @@ var serverCmd = &cobra.Command{
 			os.Setenv(model.MattermostChannel, mattermostChannel)
 		}
 
+		utilitiesGitURL, _ := command.Flags().GetString("utilities-git-url")
+		if utilitiesGitURL != "" {
+			model.SetUtilityDefaults(utilitiesGitURL)
+		} else {
+			return errors.New("utilities-git-url must be set")
+		}
+
```
Szymongib, line 146: Can be slightly simplified 😉 
```suggestion
		if utilitiesGitURL == "" {
			return errors.New("utilities-git-url must be set")
		}
		model.SetUtilityDefaults(utilitiesGitURL)

```

internal/provisioner/stackrox.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/mattermost/mattermost-cloud/internal/tools/kops"
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+type stackrox struct {
+	awsClient      aws.AWS
```
Szymongib, line 19: I think we are not using `awsClient` as a field, so I think it could be removed.

model/cluster_request.go
```diff
@@ -84,6 +84,9 @@ func (request *CreateClusterRequest) SetDefaults() {
 	if _, ok := request.DesiredUtilityVersions[PgbouncerCanonicalName]; !ok {
 		request.DesiredUtilityVersions[PgbouncerCanonicalName] = PgbouncerDefaultVersion
 	}
+	if _, ok := request.DesiredUtilityVersions[StackroxCanonicalName]; !ok {
```
Szymongib, line 87: Just realized we already had a mechanism in place that we rediscovered again in our discussion 😂 

I just wonder if we should:
- Add a check if `ValuesPath` is not empty if the utility is present. For example:
```go
	if _, ok := request.DesiredUtilityVersions[PgbouncerCanonicalName]; !ok {
		request.DesiredUtilityVersions[PgbouncerCanonicalName] = PgbouncerDefaultVersion
	} else if request.DesiredUtilityVersions[PgbouncerCanonicalName].Values() == "" {
		request.DesiredUtilityVersions[PgbouncerCanonicalName].ValuesPath = PgbouncerDefaultVersion.ValuesPath
	}
```
As if I am not mistaken we are able to pass nonempty `Version` but empty `ValuesPath` which probably would not work.
- I think we should also add set default in `ProvisionClusterRequest` as such:
```go
	pgBouncerUtility, ok := request.DesiredUtilityVersions[PgbouncerCanonicalName]
	if ok && pgBouncerUtility.Values() == ""  {
		request.DesiredUtilityVersions[PgbouncerCanonicalName].ValuesPath = PgbouncerDefaultVersion.ValuesPath
	}
```
So that we do not need to pass the full values path together with the version if we want to upgrade the utility.


I might be wrong on that as I did not test it and going with my gut here 😄

cmd/cloud/cluster.go
```diff
@@ -32,20 +32,22 @@ func init() {
...
+	clusterCreateCmd.Flags().String("nginx-version", "", "The version of Nginx Internal to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().String("nginx-internal-version", "", "The version of Nginx to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().String("teleport-version", "", "The version of Teleport to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().String("pgbouncer-version", "", "The version of Pgbouncer to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().String("stackrox-version", "", "The version of Stackrox to provision. Use 'stable' to provision the latest stable version published upstream.")
+	clusterCreateCmd.Flags().String("prometheus-operator-values", "", "The branch name of the desired chart value file's version for Prometheus Operator")
+	clusterCreateCmd.Flags().String("thanos-values", "", "The branch name of the desired chart value file's version for Thanos")
```
Szymongib, line 44: > The branch name of the desired chart value file's version for Thanos

Is it still just a branch or do we need to provide a full URL? I think we should be able to do it with just a branch but that would require some changes.

Szymongib (commented):
Looks good, I just wonder if we are handling ValuesPath correctly in all cases.

internal/provisioner/stackrox.go
stylianosrigas: Thanks removed :)

cmd/cloud/cluster.go
stylianosrigas: Its the whole URL. Updated the cmd help.

Szymongib (approved):
LGTM!

model/cluster_request.go
Szymongib: We have discussed to resolve it as a follow up https://mattermost.atlassian.net/browse/MM-37273

PR #481, opened 2021-07-19
Add DB migration e2e test

Szymongib:
#### Summary
This PR introduces the Installation DB migration e2e test I heavily used when developing the feature.
It directly tests Installation DB migration and indirectly tests DB backup and restore.

For the test to run 2 multitenant DB clusters are required in the VPC of a cluster.

We can later enhance tests so that they can do all necessary setup or run as a part of a larger e2e test.

#### Release Note

```release-note
Add DB migration e2e test
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,19 @@ Run the go tests to test:
 $ go test ./...
 ```
 
+#### DB Migration end-to-end tests
+
+There are some end-to-end tests located in `./e2e` directory. 
+
+Tests can be run with `make e2e`.
+
+E2e test for DB migration requires the following setup:
+- Local instance of Provisioner.
+- Workload cluster able to handle at least 2 installations of size `1000users`.
+- 2 Multi tenant Postgres databases provisioned in the same VPC as the workload cluster.
+- The destination database to which the migration can be performed should be exported as environment variable `DESTINATION_DB`.
+- Kubeconfig of workload cluster exported locally.
+
 ### Deleting a cluster and installations
 Before deleting a cluster you will **have** to delete the installations first on it.
 
--- a/e2e/pkg/dns.go
+++ b/e2e/pkg/dns.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+//+build e2e
+
+package pkg
+
+import (
+	"fmt"
+	"math/rand"
+)
+
+const (
+	installationDNSFormat = "e2e-test-%s.%s.cloud.mattermost.com"
+)
+
+// GetDNS returns e2e test dns with random suffix.
+func GetDNS(env string) string {
+	installationDNS := fmt.Sprintf(installationDNSFormat, randStringBytes(4), env)
+	return installationDNS
+}
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyz"
+
+func randStringBytes(n int) string {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+	return string(b)
+}
--- a/e2e/pkg/dns.go
+++ b/e2e/pkg/dns.go
@@ -17,8 +17,7 @@ const (
 
 // GetDNS returns e2e test dns with random suffix.
 func GetDNS(env string) string {
-	installationDNS := fmt.Sprintf(installationDNSFormat, randStringBytes(4), env)
-	return installationDNS
+	return fmt.Sprintf(installationDNSFormat, randStringBytes(4), env)
 }
 
 const letterBytes = "abcdefghijklmnopqrstuvwxyz"
--- a/e2e/pkg/export.go
+++ b/e2e/pkg/export.go
@@ -1,0 +1,148 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+//+build e2e
+
+package pkg
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/pkg/errors"
+)
+
+// SQLSettings is struct copied from Mattermost Server
+type SQLSettings struct {
+	DriverName                  *string
+	DataSource                  *string
+	DataSourceReplicas          []string
+	DataSourceSearchReplicas    []string
+	MaxIdleConns                *int
+	ConnMaxLifetimeMilliseconds *int
+	MaxOpenConns                *int
+	Trace                       *bool
+	AtRestEncryptKey            *string
+	QueryTimeout                *int
+}
+
+// GetConnectionString fetches the connection string from cluster request's config.
+func GetConnectionString(client *model.Client, clusterInstallationID string) (string, error) {
+	out, err := client.RunMattermostCLICommandOnClusterInstallation(clusterInstallationID, []string{"config", "show", "--json"})
+	if err != nil {
+		return "", errors.Wrap(err, "while execing config show")
+	}
+
+	settings := struct {
+		SQLSettings SQLSettings
+	}{}
+
+	err = json.Unmarshal(out, &settings)
+	if err != nil {
+		return "", errors.Wrap(err, "while unmarshalling sql setting")
+	}
+
+	return *settings.SQLSettings.DataSource, nil
+}
+
+// BulkLine represents bulk export line.
+type BulkLine struct {
+	Type string `json:"type"`
+}
+
+// ExportStats are statistics of bulk export.
+type ExportStats struct {
+	Teams          int
+	Channels       int
+	Users          int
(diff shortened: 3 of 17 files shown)


e2e/pkg/export.go
```diff
@@ -0,0 +1,148 @@
...
+
+	destination := fileName
+	defer func() {
+		err := os.Remove(destination)
+		if err != nil {
+			logger.WithError(err).Warnf("failed to cleanup file %s", destination)
+		}
+	}()
+
+	// The solution with `kubectl cp` is pretty hacky, but should be fine for the purpose of tests.
```
gigawhitlocks, line 99: :grin:

README.md
```diff
@@ -166,6 +166,19 @@ Run the go tests to test:
 $ go test ./...
 ```
 
+#### DB Migration end-to-end tests
```
gigawhitlocks, line 169: It'd be nice if this could be automated (even partially) however I think just putting instructions in the README is the right first step, especially because this kind of thing can be laborious to automate with (in my opinion) not much value return. I intend to follow your lead for other e2e tests in this repo, as well as for the AWAT.

But maybe it makes sense to open a JIRA ticket to represent the work of automating setup for e2e tests at some point? I'm 0/5 on that suggestion, but I thought it was worth throwing it out there anyway.

e2e/pkg/installation.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
```
gigawhitlocks, line 17: this import got grouped incorrectly

e2e/pkg/installation.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// WaitForInstallationAvailability pings installation until it responds successfully.
+func WaitForInstallationAvailability(dns string, log logrus.FieldLogger) error {
+	err := WaitForFunc(NewWaitConfig(20*time.Minute, 20*time.Second, 2, log), func() (bool, error) {
+		resp, err := http.Get(pingURL(dns))
+		if err != nil {
+			log.WithError(err).Error("Error while pining request")
```
gigawhitlocks, line 25: `pinging` is misspelled but the error message could also include a little more context for debugging:

```suggestion
			log.WithError(err).Errorf("Error while making ping request to Installation %s", dns)
```

e2e/pkg/installation_builder.go
```diff
@@ -0,0 +1,68 @@
...
+package pkg
+
+import "github.com/mattermost/mattermost-cloud/model"
+
+// InstallationBuilder is a helper to create CreateInstallationRequest.
+type InstallationBuilder struct {
+	request *model.CreateInstallationRequest
+}
+
+// NewInstallationBuilderWithDefaults sets up new InstallationBuilder with resonable defaults.
```
gigawhitlocks, line 16:
```suggestion
// NewInstallationBuilderWithDefaults sets up new InstallationBuilder with reasonable defaults.
```

e2e/pkg/installation_builder.go
```diff
@@ -0,0 +1,68 @@
...
+func NewInstallationBuilderWithDefaults() *InstallationBuilder {
```
gigawhitlocks, line 17: Couldn't this just be a function `NewInstallationWithDefaults()` that returns a `&model.CreateInstallationRequest` and then you don't need all the helper methods or the `InstallationBuilder` type? It's not clear to me what the value of this type is, especially since it isn't a real factory pattern, it's just a type that wraps (without embedding) the `CreateInstallationRequest` so you need a new `InstallationBuilder` for each `CreateInstallationRequest`, which seems to be roughly the same as if you had a function that simply returned `CreateInstallationRequest`s with predefined defaults, but with more indirection.

There's also a strong possibility that I'm not understanding something

e2e/workflow/installation.go
```diff
@@ -0,0 +1,207 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
gigawhitlocks, line 1: Awesome, these look like Installation creation/deletion e2e tests which we also need outside of the context of db migrations so this is really two-for-one here :grinning:

gigawhitlocks:
This is great, Szymon, I'm going to borrow a bunch of your patterns for the AWAT both for consistency and because I think they're good patterns

README.md
Szymongib: Yeah, automatic full setup at this point would be overkill at this point. I think it can be part of some more wholesome e2e test, also doing the setup could be much easier with terraform, etc. than doing it here, especially we would need some cleanup also.

I will create a more broad ticket for incorporating DB migration tests to some e2e flow.

e2e/pkg/installation_builder.go
Szymongib: Yeah, so this function does not contain all necessary parameters for creating installation - mainly DNS - and others can be parametrized by the test itself.

The idea is that `NewCreateInstallationRequestWithDefaults()` gives you some base for which you can then override fields that you need in a more elegant way (at least in my opinion 😄 ) than just directly changing struct values. This allows you to add your customization by chaining methods. There is an example in one of the test steps:
```go
installationBuilder := pkg.NewInstallationBuilderWithDefaults().
	DNS(pkg.GetDNS(w.env)).
	DB(w.Params.DBType).
	FileStore(w.Params.FileStoreType)
```

> It's not clear to me what the value of this type is, especially since it isn't a real factory pattern

Not sure if I understand what do you mean here. In my understanding factory pattern is something different and it does not aim to be one - it is a simple builder. That said the interpretations of different patterns nowadays vary wildely especially across different languages so we just might have a different understanding.

gigawhitlocks: Seeing the usage helps me understand the pattern that you were going for. I thought it was intended to be a Factory pattern (I've seen Factories called Builders before, so that's the source of the confusion I think), but like you said -- the Factory pattern is something different. Thanks for the clarification.

Is there a reason we shouldn't add these setters to the `CreateInstallationRequest` method set instead of creating a new type only for testing? I don't think it hurts to have them there, and then you could do (assuming the first method `NewInstallationBuilderWithDefaults` was rewritten to be a function that returns a `*CreateInstallationRequest`) 

```go
installationrequest := pkg.NewCreateInstallationRequestWithDefaults().
    DNS(pkg.GetDNS).
    DB(w.Params.DBType).
    FileStore(w.Params.FileStoreType)
 ```

and then in subsequent operations you can access `installationrequest` without indirection, instead of accessing it as `installationBuidler.request` every time. Or maybe I'm still misunderstanding something.

Anyway don't worry about it too much, the usage example helps me understand why it's done this way and if the above suggestion doesn't strike you as a great idea, it's really not that important.

**Review by gigawhitlocks — APPROVED** [review]

File: e2e/pkg/installation_builder.go (hunk 6 above)
Comment (Szymongib) [reply e2e/pkg/installation_builder.go:17]: Oh, I see you mean to define it directly on the type. Yeah, that would be possible for sure and would serve the purpose well. The only downside I see would be that methods need to be declared in the same package as their receivers therefore we would need to put it into `model` whereas it would be used just in tests so that is why I went with a different type. 

0/5 on whatever way we want to approach it.

Comment (gigawhitlocks) [reply e2e/pkg/installation_builder.go:17]: >The only downside I see would be that methods need to be declared in the same package as their receivers therefore we would need to put it into model whereas it would be used just in tests so that is why I went with a different type.

Yeah, when I made my original comment I actually thought those methods might already exist on the type in `model`, so I thought this method set was redundant. I don't think it would hurt to add them there instead, but I also don't want to make you do more work on this PR unnecessarily, so I'd say at this point just merge it as-is. 

But thanks for answering my questions, anyway, because it's helped my understanding & if we ever desire to use this pattern in the main codebase, outside of tests, we can always move these methods & modify the test as I suggested in my last comment at that time.

File: e2e/pkg/dns.go [hunk 8]
```diff
@@ -0,0 +1,32 @@
...
+)
+
+const (
+	installationDNSFormat = "e2e-test-%s.%s.cloud.mattermost.com"
+)
+
+// GetDNS returns e2e test dns with random suffix.
+func GetDNS(env string) string {
+	installationDNS := fmt.Sprintf(installationDNSFormat, randStringBytes(4), env)
+	return installationDNS
```
gabrieljackson, line 21: nit:
```suggestion
	return fmt.Sprintf(installationDNSFormat, randStringBytes(4), env)
```

e2e/pkg/installation_builder.go
gabrieljackson: > & if we ever desire to use this pattern in the main codebase, outside of tests, we can always move these methods & modify the test

Yep, definitely good with seeing how this works and bringing it into the core code later if we want.

e2e/pkg/wait.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NewWaitConfig create new WaitConfig.
+func NewWaitConfig(timeout, interval time.Duration, tolerateErrs int, log logrus.FieldLogger) WaitConfig {
+	return WaitConfig{
+		Timeout:        timeout,
+		Interval:       interval,
+		TolerateErrors: tolerateErrs,
+		Logger:         log,
+	}
+}
```
gabrieljackson, line 34: One other option for this is to setup a webhook receiving channel like I did with `ctest`. Probably good to consider later in the broader E2E framework planning.

gabrieljackson (approved):
Very nice!

PR #484, opened 2021-07-28
Mm 37196 migrating hibernated installations

Mshahidtaj:
#### Summary
This PR contains the fixes for the migration of hibernated installations.Such as

- Create  cluster installations in scale down mode for hibernating installations during migrations to target cluster

- Re-adjust respective ingress annotations to shows that it is hibernated in the target cluster

- Update the DNS entry with correct nginx LB for hibernating installation

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37196

#### Release Note

```release-note
  Handling migration of hibernated installations & DNS switch.
```

### DIFF
--- a/internal/api/cluster_installation.go
+++ b/internal/api/cluster_installation.go
@@ -492,7 +492,6 @@ func handleMigrateDNS(c *Context, w http.ResponseWriter, r *http.Request) {
 	newClusterInstallationIDs := getClusterInstallationIDs(newClusterInstallations)
 	var installationIDs []string
 	var hibernatedInstallationIDs []string
-	c.Logger.Infof("Total DNS records to migrate: %s", (len(installationIDs) + len(hibernatedInstallationIDs)))
 
 	for _, ci := range clusterInstallations {
 		installation, err := c.Store.GetInstallation(ci.InstallationID, false, false)
@@ -507,6 +506,8 @@ func handleMigrateDNS(c *Context, w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	totalInstallations := len(installationIDs) + len(hibernatedInstallationIDs)
+	c.Logger.Infof("Total DNS records to migrate: %s", totalInstallations)
 	status := dnsMigration(c, mcir, clusterInstallationIDs, newClusterInstallationIDs, installationIDs, hibernatedInstallationIDs)
 	if status != 0 {
 		c.Logger.Error("Failed to migrate DNS records")
--- a/internal/api/cluster_installation.go
+++ b/internal/api/cluster_installation.go
@@ -488,19 +488,45 @@ func handleMigrateDNS(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// DNS Switch
-	clusterInstallationIDs := getClusterInstallationIDs(clusterInstallations)
-	newClusterInstallationIDs := getClusterInstallationIDs(newClusterInstallations)
-	c.Logger.Infof("Total DNS records to migrate: %s", len(clusterInstallations))
-	var installationIDs []string
+	var stableInstallationIDs []string
+	var hibernatedInstallationIDs []string
 	for _, ci := range clusterInstallations {
-		installationIDs = append(installationIDs, ci.InstallationID)
+		installation, err := c.Store.GetInstallation(ci.InstallationID, false, false)
+		if err != nil {
+			c.Logger.WithError(err).Errorf("Failed to get refreshed installation")
+			return
+		}
+		if installation.State == model.InstallationStateHibernating {
+			hibernatedInstallationIDs = append(hibernatedInstallationIDs, ci.InstallationID)
+		} else {
+			stableInstallationIDs = append(stableInstallationIDs, ci.InstallationID)
+		}
 	}
-	status := dnsMigration(c, mcir, clusterInstallationIDs, newClusterInstallationIDs, installationIDs)
-	if status != 0 {
-		c.Logger.Error("Failed to migrate DNS records")
-		w.WriteHeader(status)
-		return
+
+	c.Logger.Infof("Total DNS records to migrate: %s", (len(stableInstallationIDs) + len(hibernatedInstallationIDs)))
+	if len(stableInstallationIDs) > 0 {
+		clusterInstallationIDs := getClusterInstallationIDs(clusterInstallations, stableInstallationIDs)
+		newClusterInstallationIDs := getClusterInstallationIDs(newClusterInstallations, stableInstallationIDs)
+		status := dnsMigration(c, mcir, clusterInstallationIDs, newClusterInstallationIDs, stableInstallationIDs, model.InstallationStateCreationDNS)
+		if status != 0 {
+			c.Logger.Error("Failed to migrate DNS records")
+			w.WriteHeader(status)
+			return
+		}
 	}
+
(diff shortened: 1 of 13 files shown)


internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -193,6 +187,17 @@ func (provisioner *kopsCIAlpha) HibernateClusterInstallation(cluster *model.Clus
 	return nil
 }
 
+func configInstallationForHibernation(mattermost *mmv1alpha1.ClusterInstallation) {
+	// Hibernation is currently considered changing the Mattermost app deployment
+	// to 0 replicas in the pod. i.e. Scale down to no Mattermost apps running.
+	// The current way to do this is to set a negative replica count in the
+	// k8s custom resource. Custom ingress annotations are also used.
+	// TODO: enhance hibernation to include database and/or filestore.
+	mattermost.Spec.Replicas = 0
```
Szymongib, line 196: If I am not mistaken for old CR we need to set it to -1. I believe there is a const that was used before `hibernationReplicaCount`.

internal/store/cluster_installation_test.go
```diff
@@ -735,7 +735,7 @@ func TestSwitchDNS(t *testing.T) {
 	for _, ci := range clusterInstallations {
 		installationIDs = append(installationIDs, ci.InstallationID)
 	}
-	err = sqlStore.SwitchDNS(oldCIsIDs, newCIsIDs, installationIDs)
+	err = sqlStore.SwitchDNS(oldCIsIDs, newCIsIDs, installationIDs, nil)
```
Szymongib, line 738: I think it would be good to add at least one hibernating installation to verify it is switched to the correct state.

internal/supervisor/installation.go
```diff
@@ -1455,3 +1463,23 @@ func getAnnotationsNames(annotations []*model.Annotation) []string {
 func elapsedTimeInSeconds(createAtMillis int64) float64 {
 	return float64(time.Now().Unix()*1000-createAtMillis) / 1000
 }
+
+// migratingHibernateInstallation deeal with dns update for hibernated installations during migration
+func (s *InstallationSupervisor) migratingHibernateInstallation(installation *model.Installation, instanceID string, logger log.FieldLogger) string {
```
Szymongib, line 1468: I think it would be good to adjust the name of this method so that it is straightforward that we are migrating DNS not something else.

Szymongib (commented):
Looks good! A couple of minor comments but I think that should take us through current issues to have e2e migration flow.

As we discussed earlier to cover different edge cases we might need to change our approach slightly in follow up tickets.

Szymongib (approved):
LGTM

stylianosrigas (approved):
LGTM 👌

internal/api/cluster_installation.go
```diff
@@ -490,17 +490,31 @@ func handleMigrateDNS(c *Context, w http.ResponseWriter, r *http.Request) {
 	// DNS Switch
 	clusterInstallationIDs := getClusterInstallationIDs(clusterInstallations)
 	newClusterInstallationIDs := getClusterInstallationIDs(newClusterInstallations)
-	c.Logger.Infof("Total DNS records to migrate: %s", len(clusterInstallations))
 	var installationIDs []string
+	var hibernatedInstallationIDs []string
```
gabrieljackson, line 494: nit: these lines can be combined
```suggestion
	var installationIDs, hibernatedInstallationIDs []string
```

internal/provisioner/kops_provisioner_cluster_installation.go
```diff
@@ -193,6 +187,17 @@ func (provisioner *kopsCIAlpha) HibernateClusterInstallation(cluster *model.Clus
...
 
+func configInstallationForHibernation(mattermost *mmv1alpha1.ClusterInstallation) {
+	// Hibernation is currently considered changing the Mattermost app deployment
+	// to 0 replicas in the pod. i.e. Scale down to no Mattermost apps running.
+	// The current way to do this is to set a negative replica count in the
+	// k8s custom resource. Custom ingress annotations are also used.
+	// TODO: enhance hibernation to include database and/or filestore.
+	mattermost.Spec.Replicas = hibernationReplicaCount
+	mattermost.Spec.IngressAnnotations = getHibernatingIngressAnnotations()
+	mattermost.Spec.Size = ""
```
gabrieljackson, line 198: Is there a reason we are setting the size to `""` here? I don't think we want to do that necessarily. It has been a while since I reviewed the v1alpha1 operator logic, but I am pretty sure we shouldn't clear this value out.

internal/provisioner/kops_provisioner_cluster_installation_beta.go
```diff
@@ -139,6 +138,17 @@ func (provisioner *kopsCIBeta) HibernateClusterInstallation(cluster *model.Clust
...
 
+func configureInstallationForHibernation(mattermost *mmv1beta1.Mattermost) {
+	// Hibernation is currently considered changing the Mattermost app deployment
+	// to 0 replicas in the pod. i.e. Scale down to no Mattermost apps running.
+	// The current way to do this is to set a negative replica count in the
+	// k8s custom resource. Custom ingress annotations are also used.
+	// TODO: enhance hibernation to include database and/or filestore.
+	mattermost.Spec.Replicas = int32Ptr(0)
+	mattermost.Spec.IngressAnnotations = getHibernatingIngressAnnotations()
+	mattermost.Spec.Size = ""
```
gabrieljackson, line 149: Same here. Are we sure the operator isn't just inserting the default value when this is done?

internal/supervisor/installation.go
```diff
@@ -641,49 +643,55 @@ func (s *InstallationSupervisor) waitForCreationStable(installation *model.Insta
...
+	if err != nil {
+		logger.WithError(err).Error("Failed to create DNS CNAME record")
+		return model.InstallationStateCreationDNS
+	}
+
+	logger.Infof("Successfully configured DNS %s", installation.DNS)
+
+	return s.waitForCreationStable(installation, instanceID, logger)
+}
+func (s *InstallationSupervisor) getPublicLBEndpoint(installation *model.Installation, logger log.FieldLogger) ([]string, error) {
```
gabrieljackson, line 663:
```suggestion
}

func (s *InstallationSupervisor) getPublicLBEndpoint(installation *model.Installation, logger log.FieldLogger) ([]string, error) {
```

gabrieljackson (approved):
I am approving this as is, but please review my questions and double-check that we actually want to be clearing out the size value for the CRs.

internal/provisioner/kops_provisioner_cluster_installation.go
Mshahidtaj: hi @gabrieljackson , Valid concern , Actually during local testing i found that operator overriding the replica=0 back to 1 based on size value which was not the expected behaviour for hibernated cluster installations. After checking with @Szymongib we decided to set the `size=""`.
And operator started respecting our hibernating values after setting the `size=""`. I had a concern about waking up same installation but testing shows that waking up also working as expected.  Does it answer your question ?

gabrieljackson: Hmm, I reviewed the operator logic and see that there is some difference in handling between and empty size and not when it comes to individual resource settings. If your tests worked well with this then let's merge as is and we can investigate later if needed.

PR #485, opened 2021-07-29
MM-33987 - Skip cluster deletion if it does not exist

Szymongib:
#### Summary

This PR prevents Cluster from ending up in `deletion-failed` state if kops cluster was never created.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-33986

#### Release Note

```release-note
Skip cluster deletion if it does not exist
```


gabrieljackson (approved):

Mshahidtaj (approved):
LGTM. Just curious, will these changes also take care of the scenario , where cluster creation got failed without storing any state in the s3 bucket but there is some metadata in the database for it.

Szymongib:
> LGTM. Just curious, will these changes also take care of the scenario , where cluster creation got failed without storing any state in the s3 bucket but there is some metadata in the database for it.

Most of those cases should already be handled and the bug fix should "unlock" it, so yes.

PR #487, opened 2021-07-29
Keep two versions of calico used based on cni used in cluster

stafot:
Calico-node pods failed to scheduled in aws cni clusters.
Since co-hosting two CNIs is temporary decided to keep both versions and applied the proper one depending on the CNI used in cluster.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37444

#### Release Note

```release-note
BugFix: Keep old calico version for policy only usage.
```


stylianosrigas (approved):
👌

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -367,6 +364,20 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 		},
 	}
 
+	if len(cluster.ProvisionerMetadataKops.Networking) > 0 && (cluster.ProvisionerMetadataKops.Networking == "amazonvpc" || cluster.ProvisionerMetadataKops.Networking == "weave") {
```
Mshahidtaj, line 367: I think we may also want to check `cluster.ProvisionerMetadataKops != nil` otherwise in some edge cases checking `cluster.ProvisionerMetadataKops.Networking' on nill parent may cause nil pointer exception

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -367,6 +364,20 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
...
+		files = append(files, k8s.ManifestFile{
+			Path:            "manifests/calico-policy-only-aws-cni.yaml",
+			DeployNamespace: "kube-system",
+		})
+	}
+
+	if len(cluster.ProvisionerMetadataKops.Networking) > 0 && cluster.ProvisionerMetadataKops.Networking == "calico" {
```
Mshahidtaj, line 374:
```
if len(cluster.ProvisionerMetadataKops.Networking) > 0 && cluster.ProvisionerMetadataKops.Networking == "calico" {
		files = append(files, k8s.ManifestFile{
			Path:            "manifests/calico-cni.yaml",
			DeployNamespace: "kube-system",
		})
	} else {
        files = append(files, k8s.ManifestFile{
			Path:            "manifests/calico-policy-only-aws-cni.yaml",
			DeployNamespace: "kube-system",
		})
}
```
as we want this path `Path: "manifests/calico-policy-only-aws-cni.yaml",` for anything other then `calico`

Mshahidtaj: will be good if we can also rename `manifests/calico-policy-only-aws-cni.yaml` to something like `manifests/calico-network-policy-only.yaml",`

Mshahidtaj (changes requested):
Good Work @stafot , some minor suggestions.

Mshahidtaj (approved):
LGTM

PR #489, opened 2021-08-02
Add a new tool for testing provisioner installations like a real user

spirosoik:
#### Summary
The tool is called testwick and can be build by itself as a CLI tool. The
implementation though is like a framework which you can use by itself and
write your own scenarios but of course always respecting the order of the
actions.

Right now it creates installations and tear down them in the end. It runs
couple of actions to these installations like:
- Create Channel
- Create Users
- Create Teams
- Add team members
- Post messages

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37408

#### Release Note
```release-note
Adding a new isolated tool called `testwick` and it is used to create multiple installations and interact with them
```


spirosoik:
@Szymongib it's almost done but I need to add the test for testwicker. Though before open the PR feedback will be very valuable.

cmd/tools/testwick/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/mattermost/mattermost-cloud/cmd/tools/testwick
```
Szymongib, line 1: I wonder what are advantages of adding it as a separate module?

cmd/tools/testwick/testwicker.go
```diff
@@ -0,0 +1,246 @@
...
+// PostMessage post messages to the different channels which are created
+// automatically by the wicker
+func (w *TestWicker) PostMessage(samples int) *TestWicker {
+	w.logger.WithField("DNS", w.installation.DNS).Info("Posting messages")
+	if w.channelID == "" {
+		w.Error = fmt.Errorf("failed to post message. You need to create channels first")
+		return w
+	}
+	for i := 0; i < samples; i++ {
+		_, response := w.mmClient.CreatePost(&mmodel.Post{
```
Szymongib, line 77: Did not know there is a MM client like that 😄  +1 for using it.

cmd/tools/testwick/testwicker.go
```diff
@@ -0,0 +1,246 @@
...
+	if response.StatusCode != 201 {
+		w.Error = fmt.Errorf("failed to create channel status = %d, message = %s", response.StatusCode, response.Error.Message)
+		w.logger.WithError(w.Error).Error()
+		return w
+	}
+	return w
+}
+
+// WaitForInstallationStable waits for installation to become stable
+func (w *TestWicker) WaitForInstallationStable(ctx context.Context) *TestWicker {
```
Szymongib, line 170: We have some similar functions in our e2e tests https://github.com/mattermost/mattermost-cloud/blob/master/e2e/pkg/installation.go#L52
I think it might be worth it to standardize it down the line in some package, but that is outside of the scope of this PR.

Szymongib (commented):
Looks really good at a first glance! Great work!
 Ping me when ready to final review 😉

cmd/tools/testwick/go.mod
spirosoik: avoid bumping aws sdk version etc in the main go module

cmd/tools/testwick/testwicker.go
spirosoik: yeah which is amazing actually :)

spirosoik:
@Szymongib I think to write the unit will be an overkill as I need to create a mock for the whole mattermost api client as it doesn't exist as I see.

cmd/tools/testwick/testwicker.go
```diff
@@ -0,0 +1,246 @@
...
+// TestWicker data struct for test wicker scenarios
+type TestWicker struct {
+	provisionerClient *cmodel.Client
+	mmClient          *mmodel.Client4
+	logger            *logrus.Logger
+	installation      *cmodel.InstallationDTO
+	userID            string
+	channelID         string
+	teamID            string
+	Error             error
```
Szymongib, line 28: I am not sure about error handling here. Let's consider the following scenario:
- `CreateInstallation` failes
- `Error` is set to some value but the flow continues
- `WaitForInstallationStable` will also fail as the installation was not created
- It will override `Error` and the process will continue most likely failing further along the way

If I am not missing anything this will cause only the last error to be logged therefore it may be hard to troubleshoot.

cmd/tools/testwick/testwicker.go
```diff
@@ -0,0 +1,246 @@
...
+		w.Error = fmt.Errorf("failed logging user: username = %s, status code = %d, message = %s", u.Username, response.StatusCode, response.Error.Message)
+		return w
+	}
+	w.userID = userLogged.Id
+	return w
+}
+
+// IsUP cpings the installation till return statusCode=200 and status=OK
+func (w *TestWicker) IsUP(ctx context.Context) *TestWicker {
+	w.mmClient = mmodel.NewAPIv4Client(fmt.Sprintf("https://%s", w.installation.DNS))
```
Szymongib, line 227: Could we initialize the client earlier? I think we already have a DNS prior to the creation of installation, and initializing it here might lead to some confusion/errors if we will be extending the code, adding new methods, and mixing things up.

Szymongib (commented):
@spirosoik I think you could just define an interface here with methods you need and mock only those methods. That is pretty common practice in Go.

That said I am fine with not including unit tests for now as this is test code actually and there is not much logic 😄  If we will have time down the line it may be beneficial but should be good for now.

cmd/tools/testwick/testwicker.go
spirosoik: sure agree

cmd/tools/testwick/testwicker.go
spirosoik: Yeah this is true, I think I can split it to `testwicker, error`

spirosoik:
> @spirosoik I think you could just define an interface here with methods you need and mock only those methods. That is pretty common practice in Go.
> 
> That said I am fine with not including unit tests for now as this is test code actually and there is not much logic 😄 If we will have time down the line it may be beneficial but should be good for now.

yeah I totally forgot the interface as I struggle with nginx in community :P !!! I will add the tests it's pretty easy actually and I will pass the interface in the `New...` factory method and all good

cmd/tools/testwick/testwicker.go
spirosoik: @Szymongib what if we just keep a list of errors and dump them as logs if `len(w.Error) > 0`. Trying to think how we can keep the declarative usage but still keep all the logs.

spirosoik: Or even better have a `Runner()` which gets as input functions and iterate through them:

```go
runner.Execute([func1, func2])
```

internally the runner checks for errors and fails accordingly so we can avoid the multiple `if err!=nil` for every step

Szymongib: That sounds good, I think we actually may have something similar in e2e tests https://github.com/mattermost/mattermost-cloud/blob/master/e2e/workflow/engine.go#L30

spirosoik:
@Szymongib when you have time check the workflow runner, I created my own as it's slight different and I don't need the in parallel queue for now etc.

Szymongib (approved):
LGTM! Great stuff!

PR #490, opened 2021-08-02
Add installation deployment report command

gabrieljackson:
This new command outputs information for all resources related to
an installation's deployment for quick review.

Also, some commonly used time logic has been refactored into the
model package to eliminate duplication.

Note: I know the report string format logic is pretty "meh". It was
just meant to be a quick win and I didn't find a go library that gave
me what I was looking for so this will have to do for now.

Fixes https://mattermost.atlassian.net/browse/MM-37461

```release-note
Add installation deployment report command
```

### DIFF
--- a/cmd/cloud/backup.go
+++ b/cmd/cloud/backup.go
@@ -7,7 +7,6 @@ package main
 import (
 	"os"
 
-	"github.com/mattermost/mattermost-cloud/internal/tools/utils"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
@@ -100,7 +99,7 @@ var backupListCmd = &cobra.Command{
 					backup.InstallationID,
 					string(backup.State),
 					backup.ClusterInstallationID,
-					utils.TimeFromMillis(backup.RequestAt).Format("2006-01-02 15:04:05 -0700 MST"),
+					model.TimeFromMillis(backup.RequestAt).Format("2006-01-02 15:04:05 -0700 MST"),
 				})
 			}
 			table.Render()
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -656,12 +656,7 @@ var installationDeploymentReportCmd = &cobra.Command{
 		output += fmt.Sprintf(" ├ Database Type: %s\n", installation.Database)
 		if model.IsMultiTenantRDS(installation.Database) {
 			databases, err := client.GetMultitenantDatabases(&model.GetDatabasesRequest{
-				DatabaseType: model.DatabaseEngineTypePostgresProxy,
-				Paging: model.Paging{
-					Page:           0,
-					PerPage:        model.AllPerPage,
-					IncludeDeleted: true,
-				},
+				Paging: model.AllPagesWithDeleted(),
 			})
 			if err != nil {
 				return errors.Wrap(err, "failed to query installation database")
@@ -689,11 +684,7 @@ var installationDeploymentReportCmd = &cobra.Command{
 
 		clusterInstallations, err := client.GetClusterInstallations(&model.GetClusterInstallationsRequest{
 			InstallationID: installation.ID,
-			Paging: model.Paging{
-				Page:           0,
-				PerPage:        model.AllPerPage,
-				IncludeDeleted: true,
-			},
+			Paging:         model.AllPagesWithDeleted(),
 		})
 		if err != nil {
 			return errors.Wrap(err, "failed to query cluster installations")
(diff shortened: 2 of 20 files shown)


cmd/cloud/installation.go
```diff
@@ -615,3 +620,101 @@ var installationRecoveryCmd = &cobra.Command{
...
+			output += fmt.Sprintf(" │ └ Deleted: %s\n", installation.DeletionDateString())
+		}
+		output += fmt.Sprintf(" ├ DNS: %s\n", installation.DNS)
+		output += fmt.Sprintf(" ├ Size: %s\n", installation.Size)
+		output += fmt.Sprintf(" ├ Affinity: %s\n", installation.Affinity)
+		output += fmt.Sprintf(" ├ Environment Variables: %d\n", len(installation.MattermostEnv))
+		output += fmt.Sprintf(" ├ Database Type: %s\n", installation.Database)
+		if model.IsMultiTenantRDS(installation.Database) {
+			databases, err := client.GetMultitenantDatabases(&model.GetDatabasesRequest{
+				DatabaseType: model.DatabaseEngineTypePostgresProxy,
```
Szymongib, line 659: Is it intentional to look up only PG Bouncer databases? I think it would be useful to get a database ID also in the case of a regular multitenant one.

cmd/cloud/installation.go
```diff
@@ -615,3 +620,101 @@ var installationRecoveryCmd = &cobra.Command{
...
+				Paging: model.Paging{
+					Page:           0,
+					PerPage:        model.AllPerPage,
+					IncludeDeleted: true,
+				},
```
Szymongib, line 664: Could be slightly simplified:
```suggestion
				Paging: model.AllPagesWithDeleted(),
```

cmd/cloud/installation.go
```diff
@@ -615,3 +620,101 @@ var installationRecoveryCmd = &cobra.Command{
...
+			output += fmt.Sprintf(" │ └ Description: %s\n", group.Description)
+		}
+
+		clusterInstallations, err := client.GetClusterInstallations(&model.GetClusterInstallationsRequest{
+			InstallationID: installation.ID,
+			Paging: model.Paging{
+				Page:           0,
+				PerPage:        model.AllPerPage,
+				IncludeDeleted: true,
+			},
```
Szymongib, line 696:
```suggestion
			Paging: model.AllPagesWithDeleted(),
```

Szymongib (approved):
Nice improvement! I think it will prove to be very useful.

cmd/cloud/installation.go
gabrieljackson: Oh thanks, that was from testing and I forgot to remove it.

cmd/cloud/installation.go
```diff
@@ -615,3 +620,92 @@ var installationRecoveryCmd = &cobra.Command{
 		return nil
 	},
 }
+
+var installationDeploymentReportCmd = &cobra.Command{
```
gigawhitlocks, line 624: I'm a little curious why you chose to make this a new command instead of a flag like `--table`

cmd/cloud/backup.go
```diff
@@ -100,7 +99,7 @@ var backupListCmd = &cobra.Command{
 					backup.InstallationID,
 					string(backup.State),
 					backup.ClusterInstallationID,
-					utils.TimeFromMillis(backup.RequestAt).Format("2006-01-02 15:04:05 -0700 MST"),
+					model.TimeFromMillis(backup.RequestAt).Format("2006-01-02 15:04:05 -0700 MST"),
```
gigawhitlocks, line 102: A super minor PR nitpick: moving this helper function to `model` is a change I support, but it would've been nice to see it in its own commit so that I could view the meat of the change surrounding the new command separately from this rename

cmd/cloud/installation.go
```diff
@@ -615,3 +620,92 @@ var installationRecoveryCmd = &cobra.Command{
...
+			IncludeGroupConfigOverrides: true,
+		})
+		if err != nil {
+			return errors.Wrap(err, "failed to query installation")
+		}
+		if installation == nil {
+			return nil
+		}
+
+		output := fmt.Sprintf("Installation: %s\n", installation.ID)
```
gigawhitlocks, line 646: I feel pretty confident that it's not worth spending the time to refactor this, but in the future for building big output strings like this one you might want to try out the [template](https://pkg.go.dev/text/template) package if you haven't before. It's super handy for building stuff like this without a million calls to `fmt.Sprintf`

gigawhitlocks (approved):
I just had those non-actionable comments. Looks good to me!

PR #491, opened 2021-08-03
MM-37571-locking-installations for all type of installations

Mshahidtaj:
#### Summary
Fix for locking both hibernating & non hibernating installations during dns switch with some test coverage.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37571

#### Release Note

```release-note
Fix for locking both hibernating & non hibernating installations during dns switch with some test coverage.
```

### DIFF
--- a/internal/api/cluster_installation_test.go
+++ b/internal/api/cluster_installation_test.go
@@ -743,6 +743,170 @@ func TestMigrateDNS(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	installation3 := &model.Installation{
+		OwnerID:   "owner3",
+		Version:   "version3",
+		Image:     "custom-image",
+		DNS:       "dns3.example.com",
+		Database:  model.InstallationDatabaseMysqlOperator,
+		Filestore: model.InstallationFilestoreMinioOperator,
+		Affinity:  model.InstallationAffinityIsolated,
+		State:     model.InstallationStateHibernating,
+	}
+
+	err = sqlStore.CreateInstallation(installation3, nil)
+	require.NoError(t, err)
+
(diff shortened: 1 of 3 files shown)


internal/api/cluster_installation_test.go
```diff
@@ -743,6 +743,170 @@ func TestMigrateDNS(t *testing.T) {
...
+			IsActive:       &isActiveClusterInstallations,
+		}
+		cis, err = sqlStore.GetClusterInstallations(filter)
+		require.NoError(t, err)
+		require.NotEmpty(t, cis)
+		assert.Len(t, cis, 3)
+	})
+
+}
+func TestMigrateDNSForHibernatingInstallation(t *testing.T) {
```
Szymongib, line 829: Are those tests `TestMigrateDNSForHibernatingInstallation` and `TestMigrateDNSForNonHibernatingInstallation` covering anything more than `TestMigrateDNS`? It seems that in `TestMigrateDNS` we already create both non-hibernating and hibernating installations. If so I would opt for deleting them to reduce the amount of maintenance required.

Szymongib (approved):

internal/api/cluster_installation_test.go
Mshahidtaj: During migration i came across 3 possible use cases.

1. - Migrating everything
2. - Migrating only hibernating installations 
3. - Migrating only non hibernating installations

e.g the locking fix in this PR can easily be identified earlier if we already had test coverage for above point 2 & 3. Although we r good for now but i think keeping them for future coverage would be good.

stylianosrigas (approved):
LGTM

PR #492, opened 2021-08-04
MM-37622-fix-clusterid-starting-with-digit

Mshahidtaj:
#### Summary
Fix for cluster ID to start with non numeric digit as with TF v0.12.31 , Terraform resource names cannot start with a digit.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37622

#### Release Note

```release-note
Fix for cluster ID to start with non numeric digit.
Also quick fix for calico cni manifest for new cluster creation if networking is calico.
```


gabrieljackson (approved):

model/id.go
```diff
@@ -24,3 +25,17 @@ func NewID() string {
...
+func ClusterNewID() string {
+	var b bytes.Buffer
+	encoder := base32.NewEncoder(encoding, &b)
+	encoder.Write(uuid.NewRandom())
+	encoder.Close()
+	b.Truncate(26) // removes the '==' padding
+	r := rune(b.String()[0])
+	if unicode.IsNumber(r) {
+		return ClusterNewID()
+	}
```
gabrieljackson, line 39: I would like to see this part of the logic changed to just generate a random letter and replace the first character of the ID if it was a number. In theory, this function could get "unlucky" which would result in it calling itself many times in a row. That said, this will rarely be used so I won't block the PR.

Mshahidtaj: @gabrieljackson fair point, i'have changed the code to replace the recursion with single random first char concatenation.

model/id.go
```diff
@@ -24,3 +27,21 @@ func NewID() string {
 	b.Truncate(26) // removes the '==' padding
 	return b.String()
 }
+
+// ClusterNewID is a globally unique identifier for cluster ID.  It is a [A-Z0-9] string 26
```
Szymongib, line 31: I would mention here that it starts with a letter.

model/id.go
```diff
@@ -24,3 +27,21 @@ func NewID() string {
...
+func ClusterNewID() string {
+	var b bytes.Buffer
+	encoder := base32.NewEncoder(encoding, &b)
+	encoder.Write(uuid.NewRandom())
+	encoder.Close()
+	b.Truncate(26) // removes the '==' padding
+	strID := b.String()
```
Szymongib, line 38: I think we could reuse `NewID` at the beginning as this is exactly the same code:
```suggestion
	strID := NewID()
```

Szymongib (approved):

stylianosrigas (approved):
LGTM 👌

PR #495, opened 2021-08-09
Cancel Imports made against deleted Installations

gigawhitlocks:
#### Summary

Cancels any Imports that are pending for deleted Installations so that the AWAT doesn't retry forever on an Installation that will never exist.

#### Ticket Link

No ticket, I just found this missing behavior during e2e testing

#### Release Note

```release-note
Added cancellation behavior to Imports against `deleted` Installations
```


gabrieljackson (approved):
👍

Szymongib (approved):

PR #496, opened 2021-08-10
Add unit tests for testwick

spirosoik:
#### Summary
The change includes couple of unit tests for testwick and different cases
covered with golang table driven approach.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37408

#### Release Note
```release-note
NONE
```


cmd/tools/testwick/testwicker_test.go
```diff
@@ -0,0 +1,734 @@
...
+			},
+			expected: func(wicker *TestWicker, err error) {
+				assert.NotNil(t, err)
+				assert.Contains(t, err.Error(), "failed to add a team member")
+			},
+		},
+		{
+			description: "add team member success",
+			setup: func(wicker *TestWicker) {
+				wicker.userID = "1234"
```
Szymongib, line 448: This test is failing on my machine, I think we need to specify `teamID` also.
```suggestion
				wicker.teamID = "1234"
				wicker.userID = "1234"
```

Szymongib (commented):
Nice tests! One is failing but that should be an easy fix.

I think we should run those as part of CI so that we can catch those things on PRs.

cmd/tools/testwick/testwicker_test.go
Szymongib: Sorry I think mocks need to be adjusted too.

spirosoik: let me check

spirosoik:
@Szymongib issues fixed and added the tests in CI too.

cmd/cloud/main.go
```diff
@@ -15,6 +16,7 @@ var rootCmd = &cobra.Command{
 	Use:   "cloud",
 	Short: "Cloud is a tool to provision, manage, and monitor Kubernetes clusters.",
 	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Print("skata")
```
Szymongib, line 19: Is it some leftover from print debugging, or am I missing something? 😄

Szymongib (approved):
Awesome! Thanks a lot for adding CI!

PR #498, opened 2021-08-12
Fix AWAT import completion synchronization bug

gigawhitlocks:
#### Summary

This PR adds a new state to Installations to represent when the Provisioner's Import Supervisor has finished supervising an ongoing import: `import-complete`. This new state allows for better synchronization with the AWAT, in order to solve the synchronization issue described here https://github.com/mattermost/awat/pull/18#discussion_r686629564 once and for all

Previously when an import was completed, the Import Supervisor would move an Installation's state back to `stable`, but this created an ambiguity for the AWAT in determining whether an Import was completed or had never been started.

Now, when the import is completed, the Import Supervisor will move the Installation to the `import-complete` state, and the AWAT will eventually receive that information via polling for the Installation's status, and mark the import complete in its database.

In order to move the Installation back to `stable`, the Import Supervisor will also now, before performing an import if one is available, check any Installations in `import-complete` state against the AWAT. If the AWAT is in agreement that the most recently completed import it has seen for that Installation is complete, then the Installation is moved to `stable`.

This extra state and functionality in the supervisor should resolve the possible race condition / synchronization problem Szymon identified and that doesn't seem to be avoidable with heuristics and boolean logic ;)

#### Ticket Link

This PR is necessary to unblock https://github.com/mattermost/awat/pull/18

#### Release Note

```release-note
Added new state import-complete to aid state synchronization with AWAT at end of imports
```

### DIFF
--- a/internal/supervisor/import.go
+++ b/internal/supervisor/import.go
@@ -105,88 +105,6 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store importSt
 	}
 }
 
-// completeImports transitions Installations back to stable after
-// checking with the AWAT to make sure that it has detected the
-// completion of the import
-func (s *ImportSupervisor) completeImports() error {
-	installationList, err := s.store.GetInstallations(
-		&model.InstallationFilter{
-			Paging: model.AllPagesNotDeleted(),
-			State:  model.InstallationStateImportComplete,
-		}, false, false)
-	if err != nil {
-		return errors.Wrap(err, "failed to list Installations")
-	}
-
-	for _, installation := range installationList {
-		importStatusList, err := s.awatClient.GetImportStatusesByInstallation(installation.ID)
-		if err != nil {
-			s.logger.WithError(err).Warnf("failed to get Import status for Installation %s", installation.ID)
-			continue
-		}
-		if len(importStatusList) == 0 {
-			s.logger.Warnf("no imports found for Installation with state %s, %s", installation.State, installation.ID)
-			continue
-		}
-		// find the most recently completed Import; that's the only one we
-		// care about
-		var mostRecentImport *awat.ImportStatus
-		for _, is := range importStatusList {
-			if is.State != awat.ImportStateComplete {
-				// An import is still running against this Installation.
-				// Multiple parallel imports against an Installation are not supported.
-				break
-			}
-			if mostRecentImport == nil || is.CompleteAt > mostRecentImport.CompleteAt {
-				mostRecentImport = is
-			}
-		}
-		if mostRecentImport == nil {
-			// an Import might still be running
-			continue
-		}
-
-		locked, err := s.store.LockInstallation(installation.ID, s.ID)
-		if err != nil {
-			return errors.Wrapf(err, "failed to lock Installation %s", installation.ID)
-		}
-		if !locked {
-			return errors.Errorf("failed to lock Installation %s", installation.ID)
-		}
-		defer func() {
-			unlocked, err := s.store.UnlockInstallation(installation.ID, s.ID, false)
-			if err != nil {
-				s.logger.WithError(err).Warnf("failed to unlock Installation %s to mark Import %s complete", installation.ID, mostRecentImport.ID)
-				return
-			}
-			if !unlocked {
-				s.logger.Warnf("failed to unlock Installation %s to mark Import %s complete", installation.ID, mostRecentImport.ID)
-			}
-		}()
-
-		// no Imports are running, the Installation may be moved to stable
-		installation.State = model.InstallationStateStable
-		err = s.store.UpdateInstallation(installation)
-		if err != nil {
-			s.logger.WithError(err).Error("failed to mark Installation stable")
-			// this continue is currently redundant but if anything is ever
-			// added to this function after this if stanza, it will be good
-			// to already have so it isn't omitted by mistake
-			continue
-		}
-		err = webhook.SendToAllWebhooks(s.store, &model.WebhookPayload{
-			Type:      model.TypeInstallation,
-			ID:        installation.ID,
-			NewState:  model.InstallationStateStable,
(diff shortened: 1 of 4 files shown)


internal/supervisor/import.go
```diff
@@ -99,9 +105,80 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+	installationList, err := s.store.GetInstallations(
+		&model.InstallationFilter{
+			Paging: model.AllPagesNotDeleted(),
+			State:  model.InstallationStateImportComplete,
+		}, false, false)
+	if err != nil {
+		return errors.Wrap(err, "failed to list Installations")
+	}
+
+	for _, installation := range installationList {
```
gabrieljackson, line 121: We probably want to work on the installations only once we have a valid lock. Not sure if we would have issues without that, but better safe than sorry.

gabrieljackson (commented):
Definitely seems like a robust solution for this issue. I have a few quick questions below.

internal/supervisor/import.go
gigawhitlocks: You're absolutely right

internal/supervisor/import.go
```diff
@@ -99,9 +105,80 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+		if installation == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
gigawhitlocks: Haha, well, it happens in unit tests and I didn't think about it very hard when trying to make the tests pass, but I guess it's an error with the tests, I probably need to change a mock slightly

internal/supervisor/import.go
gigawhitlocks: Addressed in commit 3d6566d

internal/supervisor/import.go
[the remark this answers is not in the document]
gigawhitlocks: Addressed in commit 090212d

internal/supervisor/import.go
```diff
@@ -99,9 +105,96 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+		&model.InstallationFilter{
+			Paging: model.AllPagesNotDeleted(),
+			State:  model.InstallationStateImportComplete,
+		}, false, false)
+	if err != nil {
+		return errors.Wrap(err, "failed to list Installations")
+	}
+
+	for _, installation := range installationList {
+		importStatusList, err := s.awatClient.GetImportStatusesByInstallation(installation.ID)
```
Szymongib, line 122: Do you think it would make sense to extract the logic inside the for loop into a separate function? This way when we have `defer` with Installation unlock it will unlock immediately after the function completes and does not need to wait for all other installations to finish, which could potentially reduce the time we hold the lock.

Or perhaps we can lock all fetch installations at the beginning just to be safe (see the comment below).

internal/supervisor/import.go
```diff
@@ -99,9 +105,96 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+		}()
+
+		// no Imports are running, the Installation may be moved to stable
+		installation.State = model.InstallationStateStable
+		err = s.store.UpdateInstallation(installation)
+		if err != nil {
+			s.logger.WithError(err).Error("failed to mark Installation stable")
+			// this continue is currently redundant but if anything is ever
+			// added to this function after this if stanza, it will be good
+			// to already have so it isn't omitted by mistake
```
Szymongib, line 174: I do not think it is redundant because of webhook 🙂

internal/supervisor/import.go
```diff
@@ -99,9 +105,96 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+			continue
+		}
+		err = webhook.SendToAllWebhooks(s.store, &model.WebhookPayload{
```
Szymongib, line 177: I think we might have missed the `err` check here.

internal/supervisor/import.go
```diff
@@ -99,9 +105,96 @@ func NewImportSupervisor(awsClient toolsAWS.AWS, awat AWATClient, store installa
...
+			if mostRecentImport == nil || is.CompleteAt > mostRecentImport.CompleteAt {
+				mostRecentImport = is
+			}
+		}
+		if mostRecentImport == nil {
+			// an Import might still be running
+			continue
+		}
+
+		locked, err := s.store.LockInstallation(installation.ID, s.ID)
```
Szymongib, line 149: This probably is not very likely, but I think we should refetch the Installation after lock (or possibly lock them all at the beginning of `completeImports`?) as some other Provisioner might have performed some change to it and we might revert it, or if we running 2 instances of Provisioner for imports weird race condition may happen here.

Szymongib (commented):
Thanks for quickly addressing the potential issues! I think this should prevent them.

Looks good, I suggested just a few minor improvements.

internal/supervisor/import.go
gabrieljackson: Yeah, I am not sure that we are perfect with it, but refetching locked resources is definitely the approved method of working on resources.

gabrieljackson (approved):
LGTM other than Szymon's comments.

gigawhitlocks:
Ugh I just messed up this branch and I'm going to have to force-push, I'm sorry

gigawhitlocks:
I accidentally pulled in Gabe's last change from `master` and then git really didn't like that :\ sorry about that guys

internal/supervisor/import.go
gigawhitlocks: haha, indeed

gigawhitlocks:
All of @Szymongib 's comments are addressed in that last commit ( a698216 )

Szymongib (approved):
LGTM

PR #504, opened 2021-08-23
Avoid installation creation failure when Sloth CRD doesn't not exist

stafot:
Avoid installation creation failure when Sloth CRD doesn't not exist on a cluster

Issue: MM-38073

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38073

#### Release Note

```release-note
BugFix: Avoid installation creation failure when Sloth CRD doesn't not exist on cluster.
```


spirosoik (approved):

internal/provisioner/kops_provisioner_installation_sli.go
```diff
@@ -58,11 +60,19 @@ func (provisioner *KopsProvisioner) makeSLIs(clusterInstallation *model.ClusterI
 
 func (provisioner *KopsProvisioner) createInstallationSLI(clusterInstallation *model.ClusterInstallation, k8sClient *k8s.KubeClient) error {
 	wait := 60
+	logger := provisioner.logger.WithFields(log.Fields{
```
Szymongib, line 63: I think we can just pass logger from the calling function as an argument, as it is already initialized there and it is more consistent with the rest of the codebase.

internal/provisioner/kops_provisioner_installation_sli.go
```diff
@@ -88,11 +98,16 @@ func (provisioner *KopsProvisioner) createIfNotExistInstallationSLI(clusterInsta
 
 func (provisioner *KopsProvisioner) deleteInstallationSLI(clusterInstallation *model.ClusterInstallation, k8sClient *k8s.KubeClient) error {
 	wait := 60
+	logger := provisioner.logger.WithFields(log.Fields{
```
Szymongib, line 101: Same here I think we can just pass logger as an argument.

internal/provisioner/kops_provisioner_installation_sli.go
```diff
@@ -58,11 +60,19 @@ func (provisioner *KopsProvisioner) makeSLIs(clusterInstallation *model.ClusterI
...
 	sli := provisioner.makeSLIs(clusterInstallation)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(wait)*time.Second)
 	defer cancel()
 	_, err := k8sClient.SlothClientsetV1.SlothV1().PrometheusServiceLevels("prometheus").Create(ctx, sli, metav1.CreateOptions{})
-	if err != nil {
+	if err != nil && k8sErrors.IsNotFound(err) {
+		logger.Debugf("Sloth CRD doesn't exist on cluster: %s", err)
+		return nil
+	}
+	if err != nil && !k8sErrors.IsNotFound(err) {
```
Szymongib, line 75: If `IsNotFound(...)` we would already return.
```suggestion
	if err != nil {
```

Szymongib (approved):
LGTM, only some minor cleanups.

Thanks for a quick fix!

PR #505, opened 2021-08-23
Improve secret manager secret cleanup

gabrieljackson:
 - Specifies 14 days for secret deletion instead of relying on the
   default of 30.
 - Corrects an issue where pgbouncer database secrets were not
   properly cleaned up.
 - Ensures all secret deletion goes through one set of logic that
   includes logging of actions taken.

Fixes https://mattermost.atlassian.net/browse/MM-38087 and https://mattermost.atlassian.net/browse/MM-38088

```release-note
Improve secret manager secret cleanup
```

### DIFF
--- a/internal/tools/aws/secrets_manager.go
+++ b/internal/tools/aws/secrets_manager.go
@@ -227,29 +227,29 @@ func (a *Client) secretsManagerGetRDSSecret(awsID string, logger log.FieldLogger
 }
 
 func (a *Client) secretsManagerEnsureIAMAccessKeySecretDeleted(awsID string, logger log.FieldLogger) error {
-	return a.secretsManagerEnsureSecretDeleted(IAMSecretName(awsID), logger)
+	return a.secretsManagerEnsureSecretDeleted(IAMSecretName(awsID), false, logger)
 }
 
 func (a *Client) secretsManagerEnsureRDSSecretDeleted(awsID string, logger log.FieldLogger) error {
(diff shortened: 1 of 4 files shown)


stafot (approved):
LGTM. 
Thanks for handled this so fast!

internal/tools/aws/secrets_manager.go
```diff
@@ -227,29 +227,29 @@ func (a *Client) secretsManagerGetRDSSecret(awsID string, logger log.FieldLogger
...
 func (a *Client) secretsManagerEnsureRDSSecretDeleted(awsID string, logger log.FieldLogger) error {
-	return a.secretsManagerEnsureSecretDeleted(RDSSecretName(awsID), logger)
+	return a.secretsManagerEnsureSecretDeleted(RDSSecretName(awsID), false, logger)
 }
 
-func (a *Client) secretsManagerEnsureSecretDeleted(secretName string, logger log.FieldLogger) error {
-	_, err := a.Service().secretsManager.DeleteSecret(&secretsmanager.DeleteSecretInput{
-		SecretId: aws.String(secretName),
+func (a *Client) secretsManagerEnsureSecretDeleted(secretName string, force bool, logger log.FieldLogger) error {
+	response, err := a.Service().secretsManager.DeleteSecret(&secretsmanager.DeleteSecretInput{
```
gigawhitlocks, line 238: tiny nit: I like to put the type on its own line in cases like this where the method call and type name of the argument are both quite long, just to aid in readability when the column is narrow like in GitHub diffs

gigawhitlocks (approved):

PR #506, opened 2021-08-23
Communicate error message to AWAT when an Import fails

gigawhitlocks:
#### Summary

This change adds error reporting on failure to the Import Supervisor. When an Import fails, the Import Supervisor marks the Import as `complete` on the AWAT and sends over an error message. It retries this forever until the AWAT returns a non-error status code. This means that it will retry forever until success, unless the Provisioner is restarted before the AWAT stops having an error. This is the only scenario when data is lost, and the only data that's lost will be the error message. The Import will be moved to 'complete' on the AWAT side regardless, eventually, because the Installation is transitioned to `import-complete` and the AWAT will eventually poll & see that state.

There is a slightly more robust design for this, but it's not much more robust and it's a larger change, so I'll explain: the Provisioner could store this error message in the database & I could introduce another Installation state `import-failed`, to allow the AWAT to poll for the error & to allow the error message to avoid being lost if the Provisioner crashes before the AWAT can receive it, but the only scenario the more complicated design avoids is that one. Personally, I don't think it's worth the extra database column that needs to be added, and the extra state, and I think this PR will work perfectly in practice. If you guys disagree, I'll be happy to change the behavior in the PR (but since that's a bit more involved, I'll probably open a JIRA ticket so we can prioritize it).

#### Ticket Link

Did this as part of the e2e testing ticket

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/supervisor/import.go
+++ b/internal/supervisor/import.go
@@ -143,12 +143,11 @@ func (s *ImportSupervisor) Do() error {
 						CompleteAt: awat.Timestamp(),
 						Error:      workError,
 					})
-				if err != nil {
-					s.logger.WithError(err).Errorf("failed to report error to AWAT for Import %s", work.ID)
-					continue
+				if err == nil {
+					return
 				}
+				s.logger.WithError(err).Errorf("failed to report error to AWAT for Import %s", work.ID)
 				time.Sleep(time.Second * 5)
-				return
 			}
 		}()
 	}
--- a/internal/supervisor/import.go
+++ b/internal/supervisor/import.go
@@ -134,6 +134,23 @@ func (s *ImportSupervisor) Do() error {
 	err = s.importTranslation(work)
 	if err != nil {
 		s.logger.WithError(err).Errorf("failed to perform work on Import %s", work.ID)
+		workError := err.Error()
+		go func() {
+			for {
+				err := s.awatClient.CompleteImport(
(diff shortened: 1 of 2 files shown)


internal/supervisor/import.go
```diff
@@ -134,6 +134,22 @@ func (s *ImportSupervisor) Do() error {
 	err = s.importTranslation(work)
 	if err != nil {
 		s.logger.WithError(err).Errorf("failed to perform work on Import %s", work.ID)
+		workError := err.Error()
+		go func() {
```
Szymongib, line 138: So this looks a little scary to me as a potential place for a memory leak. This probably does not apply right now as we do not expect many imports and even fewer of them (if any) to hit this edge case. I am just generally worried of unmanaged things running infinitely. That said I get your point about not investing time in some complex mechanism for an edge case we might never encounter. 

We could possibly consider some fixed number of retries with exponential backoff if we want to handle cases when service will be down for a significant amount of time but as mentioned above it might not apply at all for the case of imports so we probably will be fine with that.

Anyway, just giving some voice to my software paranoia 😛  Leaving it up to you 😉

Szymongib (approved):

internal/supervisor/import.go
mirshahriar: Added exponential backoff in this [PR](https://github.com/mattermost/mattermost-cloud/pull/811)

PR #507, opened 2021-08-23
MM-37273 Fix validation of ValuesPath and Chart for utilities at cluster create time

gigawhitlocks:
#### Summary

This simplifies the routine used when creating a `CreateClusterRequest` to check and set defaults on unset fields.

The ticket specifies that we need this for provision as well, however I don't believe this is correct because empty fields are handled as meaning "no change desired" in the case of a ProvisionClusterRequest

The ticket also specifies that this routine was missing, however, I believe that a lot of the code I've removed was intended to prevent the bug this change resolves, however, SetDefaults had become very large and doesn't have the same behavior for all of the utilities, and I believe that this is an error induced by the large, repetitive, and largely hard-coded nature of that method.

- To resolve this, I've moved all of the hard-coded utility defaults into a map so that in SetDefaults() we can iterate through them to ensure the correct defaults for each utility.

- I have also unceremoniously removed some old migration code with a TODO that I left saying to remove it. That code was only necessary until the clusters were updated with that version of the code, but I see that instead of it being removed, it's been diligently updated to satisfy the compiler, so I've removed it as part of this PR so that we can stop doing that.

- Also I simplified `getUtilityVersion` so that it is no longer a huge switch statement that must be continually updated. I'd do the same thing for setUtilityVersion but that's a bit more involved haha.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-37273

#### Release Note

```release-note
Fixed version and valuespath validation for cluster utilities
```

### DIFF
--- a/model/cluster_request.go
+++ b/model/cluster_request.go
@@ -31,6 +31,36 @@ type CreateClusterRequest struct {
 	VPC                    string                         `json:"vpc,omitempty"`
 }
 
+func (request *CreateClusterRequest) setUtilityDefaults(utilityName string) {
+	reqDesiredUtilityVersion, ok := request.DesiredUtilityVersions[utilityName]
+	if !ok {
+		request.DesiredUtilityVersions[utilityName] = DefaultUtilityVersions[utilityName]
+		return
+	}
+	if reqDesiredUtilityVersion.Chart == "" {
+		reqDesiredUtilityVersion.Chart = DefaultUtilityVersions[utilityName].Chart
+	}
+	if reqDesiredUtilityVersion.ValuesPath == "" {
+		reqDesiredUtilityVersion.ValuesPath = DefaultUtilityVersions[utilityName].ValuesPath
+	}
+}
+
+func (request *CreateClusterRequest) setUtilitiesDefaults() {
+	for _, utilityName := range []string{
+		PrometheusOperatorCanonicalName,
+		ThanosCanonicalName,
+		NginxCanonicalName,
+		NginxInternalCanonicalName,
+		FluentbitCanonicalName,
+		TeleportCanonicalName,
+		PgbouncerCanonicalName,
+		StackroxCanonicalName,
+		KubecostCanonicalName,
+	} {
+		request.setUtilityDefaults(utilityName)
+	}
+}
+
 // SetDefaults sets the default values for a cluster create request.
 func (request *CreateClusterRequest) SetDefaults() {
 	if len(request.Provider) == 0 {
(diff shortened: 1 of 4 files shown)


model/cluster_request.go
```diff
@@ -31,6 +31,36 @@ type CreateClusterRequest struct {
...
+		NginxCanonicalName,
+		NginxInternalCanonicalName,
+		FluentbitCanonicalName,
+		TeleportCanonicalName,
+		PgbouncerCanonicalName,
+		StackroxCanonicalName,
+		KubecostCanonicalName,
+	} {
+		request.setUtilityDefaults(utilityName)
+	}
```
Szymongib, line 61: Could we just iterate through the `DefaultUtilityVersions` map keys?
```suggestion
	for utilityName, _ := range DefaultUtilityVersions {
		request.setUtilityDefaults(utilityName)
	}
```

model/cluster_utility.go
```diff
@@ -37,120 +35,59 @@ const (
...
-	KubecostDefaultVersion = &HelmUtilityVersion{Chart: "1.83.1", ValuesPath: ""}
-)
+	KubecostCanonicalName: {Chart: "1.83.1", ValuesPath: ""},
+}
 
 // SetUtilityDefaults is used to set Utility default version and values.
 func SetUtilityDefaults(url string) {
-	if PrometheusOperatorDefaultVersion.ValuesPath == "" {
-		PrometheusOperatorDefaultVersion.ValuesPath = fmt.Sprintf("%s/api/v4/projects/33/repository/files/dev%%2Fprometheus_operator_values.yaml?ref=master", url)
+	if DefaultUtilityVersions[PrometheusOperatorCanonicalName].ValuesPath == "" {
```
Szymongib, line 62: Maybe we could refactor this to a map also? Something like `DefaultUtilityURLPath`?

Szymongib (commented):
I really like the refactor!

Regarding the provisioning part mentioned in the ticket, I think that it was about this part https://github.com/mattermost/mattermost-cloud/blob/master/cmd/cloud/cluster.go#L607-L609
From CLI we can only specify both `ValuesPath` and `Version` while from API we could specify one of those and the other remain empty - which I believe should be fixed now. 

I think tho, that it would be good to have the same behavior from the CLI and directly from API, so maybe we should change the validation in CLI and allow passing only one of `Version` or `ValuesPath` also.

model/cluster_request.go
gigawhitlocks: why yes, that does make more sense :facepalm:

gigawhitlocks:
>I think tho, that it would be good to have the same behavior from the CLI and directly from API, so maybe we should change the validation in CLI and allow passing only one of Version or ValuesPath also.

I understand; on CLI we should allow passing only one of these inputs and if it's a call to `provision`, provide the old value for the unspecified input, and if it's `create`, use the defaults like we do from the API.

 I will update this PR to match the behavior described.

model/cluster_utility.go
gigawhitlocks: I'm going to try to clean this up a bit but it's a little bit of a mess, regardless, because we parameterize the base URL of the gitlab repo (that's the `url` argument here), but we hard-code the project number (`33`) so it's actually not possible to use the parameter to point to a different Gitlab instance, anyway.. unless you happen to make the project holding the repository project number 33 on the new instance, too.

I don't think this is how the original implementation worked but I think the code has been iterated upon to this state and the fact that we're hard-coding a configuration value (`33`, the project number) seems to have been overlooked until now. I'm going to spike on cleaning this up a bit.

gigawhitlocks: OK have a look at fa2b153 and let me know if you think I should spent more time making it possible to actually pass in the various parameters. I think that fa2b153 is an incremental improvement over the original code here, in any case, and it should make it easier for us to make these various parameters able to be passed in at runtime as necessary in the future, without altering the current behavior too much.

gigawhitlocks:
Sorry about the force-pushes, I just messed up that last commit and had to redo it

PR #508, opened 2021-08-24
Ensure multitenant DB users are removed

gabrieljackson:
This corrects an issue where multitenant database users were not
deleted when database teardown occurred.

Note that pgbouncer databases were not affected by this.

Fixes https://mattermost.atlassian.net/browse/MM-38117

```release-note
Ensure multitenant DB users are removed
```

### DIFF
--- a/internal/tools/aws/database_multitenant.go
+++ b/internal/tools/aws/database_multitenant.go
@@ -542,7 +542,8 @@ func (d *RDSMultitenantDatabase) RollbackMigration(store model.InstallationDatab
 	rdsID := *rdsCluster.DBClusterIdentifier
 	logger = logger.WithField("rds-cluster-id", rdsID)
 
-	err = d.dropDatabase(rdsID, *rdsCluster.Endpoint, logger)
+	username := MattermostMultitenantDatabaseUsername(d.installationID)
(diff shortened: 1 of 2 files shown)


internal/tools/aws/database_multitenant.go
```diff
@@ -957,7 +960,7 @@ func (d *RDSMultitenantDatabase) removeMigratedInstallationFromMultitenantDataba
 	return nil
 }
 
-func (d *RDSMultitenantDatabase) dropDatabase(rdsClusterID, rdsClusterendpoint string, logger log.FieldLogger) error {
+func (d *RDSMultitenantDatabase) cleanupDatabase(rdsClusterID, rdsClusterendpoint, username string, logger log.FieldLogger) error {
```
Szymongib, line 963: I think we can actually get username inside `cleanupDatabase` function to reduce some repetition.
```suggestion
func (d *RDSMultitenantDatabase) cleanupDatabase(rdsClusterID, rdsClusterendpoint string, logger log.FieldLogger) error {
    username := MattermostMultitenantDatabaseUsername(d.installationID)
```

Szymongib (approved):
LGTM!

Suggested one minor improvement.

PR #513, opened 2021-08-31
mm-37913-avoid-reprovisioning-calico-cni

Mshahidtaj:
#### Summary
FOR cluster with Calico as a networking solution we should avoid re applying calico manifest for each provisioning request as it may impact IP tables rules etc.
- Only deploy calico CNI at cluster creation time if networking option is calico
- Only deploy or reprovision calico netpol if current networking option is other then calico

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37913

#### Release Note

```release-note
- deploy calico CNI at cluster creation time if networking option is calico.
- deploy or reprovision calico netpol if current networking option is other then calico

```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -374,14 +374,19 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 		},
 	}
 
-	// Here we need to handle both case i.e creating a new cluster with possible networking value in the changerequest object & 2nd provisioning existing cluster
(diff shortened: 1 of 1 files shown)


stafot (approved):
lgtm

gabrieljackson (approved):
I had a bit of trouble reasoning about this, but it should be fine. Hopefully I am not missing any edge cases, but we can test this out and make sure it checks out.

Mshahidtaj:
> I had a bit of trouble reasoning about this, but it should be fine. Hopefully I am not missing any edge cases, but we can test this out and make sure it checks out.

Sure , this change has been tested in dev with both calico & aws cni networking option and will also create a ticket to test in test environment.

PR #521, opened 2021-09-07
Introduce Resource interface and code generation to Provisioner

Szymongib:
#### Summary

This PR is a proposal for introducing a `Resource` interface and simple code generation tool that can help us be more consistent with implementation and reduce the need for some boilerplate code.

Code generation tool works by taking a Go type (or multiple types) and a list of generators to run for that type, making it easily extensible and suitable for many types.

Currently, there are only a few methods generated that would be useful in making some generic functions in Provisioners e2e tests but we can easily expand it we new `generators` for functions like `New...FromReader` etc.

For now the code was generated only for `Installation` and `Cluster` to not to overblow the PR but the same generators apply to many other types like `ClusterInstallation`, `InstallationDBMigrationOperation` etc.

Changes in PR:
- Add a simple code generation tool.
- Introduce `Resource` and `SupervisedResource` interfaces. 
- Generate code for `Installation` and `Cluster`.
- Update `boilerplate.py` to properly handle generated files.

#### Release Note

```release-note
Introduce Resource interface and code generation to Provisioner
```


gabrieljackson (approved):
Very nice!

gigawhitlocks (approved):

PR #524, opened 2021-09-10
Add database resource list and get API endpoints

gabrieljackson:
These new endpoints allow for clients to view database resources
that the provisioner is managing.

Note that the old database endpoints will be removed at a later
date.

Fixes https://mattermost.atlassian.net/browse/MM-38467

```release-note
Add database resource list and get API endpoints
```

### DIFF
--- a/internal/api/databases.go
+++ b/internal/api/databases.go
@@ -5,10 +5,7 @@
 package api
 
 import (
-	"net/http"
-
 	"github.com/gorilla/mux"
-	"github.com/mattermost/mattermost-cloud/model"
 )
 
 // initDatabases registers database endpoints on the given router.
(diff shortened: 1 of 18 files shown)


stylianosrigas (approved):
Awesome 👌

stafot (approved):
LGTM

model/client.go
```diff
@@ -1108,7 +1108,7 @@ func (c *Client) LeaveGroup(installationID string, request *LeaveGroupRequest) e
 }
 
 // GetMultitenantDatabases fetches the list of multitenant databases from the configured provisioning server.
-func (c *Client) GetMultitenantDatabases(request *GetDatabasesRequest) ([]*MultitenantDatabase, error) {
+func (c *Client) GetMultitenantDatabases(request *GetMultitenantDatabasesRequest) ([]*MultitenantDatabase, error) {
```
Szymongib, line 1111: We should update paths used by this method to the new convention. That way updating the Provisioner client will solve the issue of migrating to the new API and no new integrations will use it.

model/client.go
```diff
@@ -1131,8 +1131,28 @@ func (c *Client) GetMultitenantDatabases(request *GetDatabasesRequest) ([]*Multi
...
+		return nil, nil
+
+	default:
+		return nil, errors.Errorf("failed with status code %d", resp.StatusCode)
+	}
+}
+
 // UpdateMultitenantDatabase updates a multitenant database.
-func (c *Client) UpdateMultitenantDatabase(databaseID string, request *PatchDatabaseRequest) (*MultitenantDatabase, error) {
+func (c *Client) UpdateMultitenantDatabase(databaseID string, request *PatchMultitenantDatabaseRequest) (*MultitenantDatabase, error) {
```
Szymongib, line 1155: Same here, we probably should use new paths.

internal/api/databases.go
```diff
@@ -17,79 +14,15 @@ func initDatabases(apiRouter *mux.Router, context *Context) {
 		return newContextHandler(context, handler)
 	}
 
+	// TODO: retire these endpoints
```
Szymongib, line 17: It would probably be good to log a ticket for removing that alongside with updating any components that depend on this.

internal/api/db_database_schema.go
```diff
@@ -0,0 +1,79 @@
...
+package api
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// initDatabases registers database schema endpoints on the given router.
```
Szymongib, line 14:
```suggestion
// initDatabaseSchemas registers database schema endpoints on the given router.
```
We could also get rid of those comments for `init` methods as they do not provide much insight 0/5.

internal/api/db_logical_database.go
```diff
@@ -0,0 +1,78 @@
...
+package api
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// initDatabases registers database endpoints on the given router.
```
Szymongib, line 14:
```suggestion
// initLogicalDatabases registers logical database endpoints on the given router.
```

internal/api/db_logical_database.go
```diff
@@ -0,0 +1,78 @@
...
+		logicalDatabases = []*model.LogicalDatabase{}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	outputJSON(c, w, logicalDatabases)
+}
+
+// handleGetMultitenantDatabase responds to GET /api/databases/multitenant_database/{multitenant_database},
+// returning the multitenant database in question.
```
Szymongib, line 58:
```suggestion
// handleGetLogicalDatabase responds to GET /api/databases/logical_database/{logical_database},
// returning the logical database in question.
```

internal/api/db_logical_database.go
```diff
@@ -0,0 +1,78 @@
...
+
+	LogicalDatabasesRouter := apiRouter.PathPrefix("/logical_databases").Subrouter()
+	LogicalDatabasesRouter.Handle("", addContext(handleGetLogicalDatabases)).Methods("GET")
+
+	LogicalDatabaseRouter := apiRouter.PathPrefix("/logical_database/{logical_database:[A-Za-z0-9]{26}}").Subrouter()
+	LogicalDatabaseRouter.Handle("", addContext(handleGetLogicalDatabase)).Methods("GET")
+}
+
+// handleGetMultitenantDatabases responds to GET /api/databases/multitenant_databases,
+// returning a list of multitenant databases.
```
Szymongib, line 28:
```suggestion
// handleGetLogicalDatabases responds to GET /api/databases/logical_databases,
// returning a list of logical databases.
```

internal/api/db_multitenant_database.go
```diff
@@ -0,0 +1,118 @@
...
+package api
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-cloud/model"
+)
+
+// initDatabases registers database endpoints on the given router.
```
Szymongib, line 14:
```suggestion
// initMultitenantDatabases registers multitenant database endpoints on the given router.
```

Szymongib (commented):
Looks good!

I am 3/5 on updating paths of Multitenant DBs client methods, the rest is just cosmetics 😉

angeloskyratzakos (approved):
LGTM

model/client.go
gabrieljackson: So, I intentionally didn't update the paths here, because this would make the client unusable against provisioners that don't have the new API yet. The plan was to later update these client methods and remove the old endpoints at the same time.

Szymongib: Yeah that is true but only if you bump provisioner version in a `go.mod` and for a CLI against older provisioner which I guess would be the only issue.

Ideally, I think we should add new methods and deprecate those but as there are not many Provisioner API clients and we want to squeeze it in quickly I will leave it up to you.

Szymongib (approved):

model/client.go
gabrieljackson: We chatted about this a bit offline and I feel this is a reasonable path forward for now. We can adjust the cleanup process a bit if needed.

PR #525, opened 2021-09-10
Remove alerting block from installation SLIs

stafot:
Moved to a generalized alerting approach with thanos.
Issue: MM-38470

#### Summary
Remove alerting block from installation SLIs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38470

#### Release Note

```release-note
Remove alerting block from installation SLIs.
Moved to a generalized alerting approach with thanos.
```


stylianosrigas (approved):

stylianosrigas (approved):
Nice 👌

Szymongib (commented):
Currently, we only create the CR when it does not exist, we do not perform an update so that will cause existing installations to remain with the previous configuration (unless hibernated and woke up). Is that the desired behavior?

stafot:
@Szymongib This feature is only on test yet, so we are going to cleanup alla unneeded resources. I will handle it.

Szymongib (approved):

PR #526, opened 2021-09-10
Fix listing kops clusters if there is only one or none clusters

Szymongib:
#### Summary

This PR fixes listing Kops clusters that occur during cluster deletion.

Kops output for `kops get clusters -o json` is not consistent, and behaves in the following ways:
- If there are multiple clusters an array of clusters is returned.
- If there is only one cluster a single cluster object is returned (not as an array).

Additionally, if no cluster exists kops exits with 1 code indicating an error.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38450

#### Release Note

```release-note
Fix listing kops clusters if there is only one or no clusters
```

### DIFF
--- a/internal/provisioner/kops_provisioner.go
+++ b/internal/provisioner/kops_provisioner.go
@@ -6,6 +6,7 @@ package provisioner
 
 import (
 	"encoding/json"
+	"strings"
 
 	"github.com/mattermost/mattermost-cloud/k8s"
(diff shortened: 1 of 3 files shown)


stafot (approved):
LGTM. Thanks for handling this!!

gabrieljackson (approved):
Nice catch!

PR #528, opened 2021-09-10
Revert "Remove alerting block from installation SLIs"

stafot:
Reverts mattermost/mattermost-cloud#525

@stylianosrigas @Szymongib Unfortunately need to revert this change for now at least as breaking `sloth`  when you don't set Alerting block. So SLOs do not get in place.

Sorry for the inconvenience.
#### Release Note

```release-note
Reverts #525 
```


stafot:
I found the proper solution. Going to raise a PR

PR #538, opened 2021-09-21
MM-38292 adding logic to skip/avoid the already migrated installations

Mshahidtaj:
#### Summary
For better control over migration process , we need to introduce replay/resume feature in it. which means it implicitly take care of all-ready migrated installations & in next run it only migrate those installations which are not migrated yet.

Same for DNS switch , it should check & skip those installation for which dns switch has been completed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38292

#### Release Note

```release-note
Feature to skip already migrated installation during migration process
```

### DIFF
--- a/internal/api/cluster_installation.go
+++ b/internal/api/cluster_installation.go
@@ -381,11 +381,10 @@ func handleMigrateClusterInstallations(c *Context, w http.ResponseWriter, r *htt
 		"target-cluster-id": mcir.TargetClusterID,
 	})
 
-	IsMigrated := false
 	filter := &model.ClusterInstallationFilter{
 		ClusterID:      mcir.SourceClusterID,
 		InstallationID: mcir.InstallationID,
-		IsMigrated:     &IsMigrated,
+		TargetCluster:  mcir.TargetClusterID,
 		Paging:         model.AllPagesNotDeleted(),
 	}
 
@@ -512,7 +511,7 @@ func handleMigrateDNS(c *Context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	totalInstallations := len(installationIDs) + len(hibernatedInstallationIDs)
-	c.Logger.Infof("Total DNS records to migrate: %s", totalInstallations)
+	c.Logger.Infof("Total DNS records to migrate: %d", totalInstallations)
 	if totalInstallations == 0 {
 		c.Logger.Error("No installation(s) found for DNS  migration")
 		w.WriteHeader(http.StatusNotFound)
--- a/internal/api/cluster_installation.go
+++ b/internal/api/cluster_installation.go
@@ -396,7 +396,7 @@ func handleMigrateClusterInstallations(c *Context, w http.ResponseWriter, r *htt
 	}
 
 	// Get the CIs for migration
-	clusterInstallations, status := getClusterInstallationsForMigration(c, mcir)
+	clusterInstallations, status := GetClusterInstallationsForMigration(c, mcir)
 	if status != 0 {
 		c.Logger.WithError(err).Error("Failed to get CIs for migration")
 		w.WriteHeader(status)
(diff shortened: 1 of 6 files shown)


Szymongib (commented):
I like the approach for the DNS switch part, but I am a little hesitant about adding a new field to `ClusterInstallation`. I have suggested an alternative approach, let me know what do you think about it.

internal/api/cluster_installation.go
```diff
@@ -381,11 +381,10 @@ func handleMigrateClusterInstallations(c *Context, w http.ResponseWriter, r *htt
 		"target-cluster-id": mcir.TargetClusterID,
 	})
 
-	IsMigrated := false
 	filter := &model.ClusterInstallationFilter{
 		ClusterID:      mcir.SourceClusterID,
 		InstallationID: mcir.InstallationID,
-		IsMigrated:     &IsMigrated,
+		TargetCluster:  mcir.TargetClusterID,
```
Szymongib, line 387: Sorry I might have not been clear enough.

So I was thinking more of just doing a second query to the database and filtering in code like so:
```go
// This whole logic could be a separate method - `getClusterInstallationsForMigration` or something. 
toMigrateFilter := &model.ClusterInstallationFilter{
		ClusterID:      mcir.SourceClusterID,
		InstallationID: mcir.InstallationID,
                 Paging:         model.AllPagesNotDeleted(),
}
alredyMigratedFilter := &model.ClusterInstallationFilter{
		ClusterID:      mcir.TargetClusterID,
		InstallationID: mcir.InstallationID,
                 Paging:         model.AllPagesNotDeleted(),
}

// Here query both
sourceClusterCIs, err := c.Store.GetClusterInstallations(toMigrateFilter)
...
migratedCIs, err := c.Store.GetClusterInstallations(alredyMigratedFilter)
...

toMigrate := []*model.ClusterInstallation{}
for _, ci := range sourceClusterCIs {
	migrate := true
	for _, migrated := range migratedCIs {
		if ci.InstallationID == migrated.InstallationID {
			migrate = false
			break
		} 
	}
	if migrate {
		toMigrate = append(toMigrate, ci)
	}
}
```

Filtering on SQL level makes sense also, but I would suggest introducing a separate method for that like `GetClusterInstallationsForMigration` to not leak migration logic into a more generic construct that is `ClusterInstallationFilter`. With some thoughtful refactoring, we could probably reuse some parts of `getClusterInstallations` there.

Mshahidtaj: All good. I thought about the above loop based comparison but felt it would be more efficient if we can handle it at SQL level. (Performance wise)
I'll check if introducing seperate method like `GetClusterInstallationsForMigration` is not too much impact on existing logic otherwise it will be simpler to go with your suggestion.

internal/api/cluster_installation.go
```diff
@@ -650,3 +644,59 @@ func getSourceAndTargetCluster(c *Context, request model.MigrateClusterInstallat
 	}
 	return sourceCluster, targetCluster, nil
 }
+
+func getClusterInstallationsForMigration(c *Context, request model.MigrateClusterInstallationRequest) ([]*model.ClusterInstallation, int) {
```
Szymongib, line 648: I think it might be good to add a few table tests specifically for this method.

internal/store/cluster_installation.go
```diff
@@ -85,6 +85,10 @@ func (sqlStore *SQLStore) getClusterInstallations(db dbInterface, filter *model.
 	if filter.IsActive != nil {
 		builder = builder.Where("IsActive = ?", *filter.IsActive)
 	}
+	// To avoid already migrated installation to the Target Cluster
+	if len(filter.TargetCluster) > 0 {
+		builder = builder.Where("InstallationID not in (SELECT InstallationID from ClusterInstallation where ClusterID = ?)", filter.TargetCluster)
+	}
```
Szymongib, line 91: I believe this can be removed now.

model/cluster_installation.go
```diff
@@ -31,6 +31,7 @@ type ClusterInstallationFilter struct {
 	InstallationID string
 	ClusterID      string
 	IsActive       *bool
+	TargetCluster  string
```
Szymongib, line 34: Same here, I think we can remove it now.

Szymongib (commented):
Looks good, thanks for doing all the changes!
Only a couple of cleanups, then I think it is good to go.

Szymongib:
@Mshahidtaj looks like there are some issues with tests

Mshahidtaj:
> @Mshahidtaj looks like there are some issues with tests

Sorry, as they were passed locally so I assumed should also be fine 

> @Mshahidtaj looks like there are some issues with tests

@Szymongib All good now. Sorry tests were passed locally so I was under the false impression that it would pass on PR as well.

Szymongib (approved):

angeloskyratzakos (approved):

PR #547, opened 2021-10-05
Mm 3666 json output migration api

Mshahidtaj:
#### Summary
This change is mainly focused on modifying the  API's endpoints & respective clients for cluster installation migration to return the possible responses to the API calls.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36660

#### Release Note

```release-note
Returning JSON response from Cluster installation migration-related API's.
```

### DIFF
--- a/cmd/cloud/cluster_installation.go
+++ b/cmd/cloud/cluster_installation.go
@@ -310,11 +310,12 @@ var clusterInstallationsMigrationCmd = &cobra.Command{
 				InstallationID:   installation,
 				DNSSwitch:        false,
 				LockInstallation: false})
-		// Print any output and then check and handle errors.
-		printJSON(clusterInstallation)
+
 		if err != nil {
 			return err
 		}
+		// Print any output and then check and handle errors.
+		printJSON(clusterInstallation)
 		return nil
 	},
 }
@@ -340,11 +341,11 @@ var dnsMigrationCmd = &cobra.Command{
 				TargetClusterID:  targetcluster,
 				InstallationID:   installation,
 				LockInstallation: lockInstallation})
-		// Print any output and then check and handle errors.
-		printJSON(clusterInstallation)
 		if err != nil {
 			return err
 		}
+		// Print any output and then check and handle errors.
+		printJSON(clusterInstallation)
 		return nil
 	},
 }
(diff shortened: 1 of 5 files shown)


internal/api/cluster_installation.go
```diff
@@ -533,9 +541,25 @@ func handleDeleteInActiveClusterInstallationsByCluster(c *Context, w http.Respon
...
-
 	c.Logger.Infof("Successfully deleted inactive cluster installations for cluster ID: %s", clusterID)
+
+	IsActive := false
+	filter := &model.ClusterInstallationFilter{
+		ClusterID: clusterID,
+		IsActive:  &IsActive,
+		Paging:    model.AllPagesNotDeleted(),
+	}
+	deletedClusterInstallations, err := c.Store.GetClusterInstallations(filter)
```
Szymongib, line 552: Although unlikely, there is a possible race condition as we only fetch not deleted Installations. I think we could get just the count of those Installations in `DeleteInActiveClusterInstallationByClusterID` (as part of the transaction before deleting) or change the response.

model/client.go
```diff
@@ -1402,86 +1402,86 @@ func (c *Client) makeSecurityCall(resourceType, id, securityType, action string)
...
 
 	default:
-		return errors.Errorf("failed with status code %d", resp.StatusCode)
+		return nil, errors.Errorf("failed with status code %d", resp.StatusCode)
 	}
 }
 
 // SwitchClusterRoles requests the migration of cluster installation(s) from the configured provisioning server.
-func (c *Client) SwitchClusterRoles(request *MigrateClusterInstallationRequest) error {
+func (c *Client) SwitchClusterRoles(request *MigrateClusterInstallationRequest) (MigrateClusterInstallationRequest, error) {
```
Szymongib, line 1473: Shouldn't this be `MigrateClusterInstallationResponse`? Also, I would suggest returning the pointer for the sake of consistency.

cmd/cloud/cluster_installation.go
```diff
@@ -303,15 +303,21 @@ var clusterInstallationsMigrationCmd = &cobra.Command{
...
 				TargetClusterID:  targetcluster,
 				InstallationID:   installation,
 				DNSSwitch:        false,
 				LockInstallation: false})
+
 		if err != nil {
-			return err
+			return errors.Wrap(err, "failed to migrate cluster installation(s)")
+		}
+		// Print any output and then check and handle errors.
```
Szymongib, line 317: Those comments are no longer valid, please remove them.

Szymongib (approved):
LGTM! Thanks for addressing the comments!

Just one thing, please clean up the outdated comments in CLI code.

PR #562, opened 2021-10-22
Only reprovision utilities when forced or version is explicitly specified

gigawhitlocks:
#### Summary

This PR updates the Provisioner to allow skipping reprovisioning of utilities. 

Now when running cloud cluster provision, if there is no version specified for a utility (both values file path & chart version must be specified), the utility will not be re-provisioned.

If a version is specified for a utility, it will be provisioned to that version, whether that is the same as the current version or not.

If `--reprovision-all-utilities` is specified, the Provisioner will exhibit the old behavior: all utilities will be provisioned to their current version unless a different one is specified, in which case the utility with an explicit version will be provisioned to the explicit version set by the SRE.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38496

#### Release Note

```release-note
Provisioner will now skip re-provisioning utilities unless `--reprovision-all-utilities` is specified, or the utility's desired version is explicitly included on the command line. 
```


stylianosrigas:
@gigawhitlocks thanks for that, a couple of questions.
1. What happens if we reprovision a cluster without passing any utility flags? I guess it just skips the reprovisioning?
2. If we use sth other than commit hash like for example master, will it reprovision the cluster using the values stored in the database? I am asking in the case that we deleted one of the utilities, we want to reprovision the cluster but the commit has not changed.

stafot:
@gigawhitlocks 
While I think that this PR is on the correct direction I have two major concerns.
1. What happens on selective reprovisioning. F.e when I pass only some flags, not all.
2. Ideally we need a mechanism to be able to **force** a re-provision on a specific utility even if the commit hash has not changed. When we are in a broken state for example.

Out of scope of this PR but since our source of truth is the provisioner database, I would like to be able to have a `json` or `yaml` "manifest" type with the current values of the cluster to be able to export them and after manipulating them to be able to pass them as input.  This approach could possibly lead to a faster and less error prone at the same time re-provisioning flow.

gabrieljackson:
@gigawhitlocks can you address the questions and comments above?

gigawhitlocks:
>What happens if we reprovision a cluster without passing any utility flags? I guess it just skips the reprovisioning?

Yes, reprovisioning the Helm utilities will be skipped, because the Provisioner uses the old values

>What happens on selective reprovisioning. F.e when I pass only some flags, not all.

The Provisioner will keep the old versions for the ones you do not select. This is the same as the behavior of the currently released Provisioner. After this PR, it will skip reprovisioning them, instead of reprovisioning them to the current version as is the current behavior.

> If we use sth other than commit hash like for example master, will it reprovision the cluster using the values stored in the database? I am asking in the case that we deleted one of the utilities, we want to reprovision the cluster but the commit has not changed.

If you use a tag or branch name, the Provisioner will not know whether or not you've changed values in the values file. It will print a scary warning and then reprovision the utility (the same behavior as today) because it has no way to tell if the content in the file has changed if you use a tag. From the description of the PR:

>If the operator chooses to use a tag like master on a utility anyway, the Provisioner will complain (log the scary warning) and then always re-provision that utility.

Finally:

>Ideally we need a mechanism to be able to force a re-provision on a specific utility even if the commit hash has not changed. When we are in a broken state for example.

Kubernetes has this same limitation, and I explained that this change would have this limitation during our Planning meeting before implementing this design. If you want to force a reprovision with the PR as written, you can either commit an empty commit to the repo/branch that the cluster sources values from, or you can create a tag or branch at that commit and use that value to have the Provisioner re-provision the utility every time until you change it to a shasum.

 If this is too cumbersome, I wonder if you have a usage example for what you would like the CLI to look like to force a reprovision? I could add a `--force` flag that would have the Provisioner reprovision all of the utilities, but I'm hesitant to add a --force-utility flag for each utility, as the usage has already gotten very cumbersome, which brings me to this:

>Out of scope of this PR but since our source of truth is the provisioner database, I would like to be able to have a json or yaml "manifest" type with the current values of the cluster to be able to export them and after manipulating them to be able to pass them as input. This approach could possibly lead to a faster and less error prone at the same time re-provisioning flow.

I've been wanting to do (essentially) this refactor for almost a year. The usage of the utilities on the CLI has gotten completely out of hand since my original design, and I think there are a number of other points of repetition that could be vastly simplified. Unfortunately it wouldn't be a trivial change. It might be easier for me to advocate for the time to make this change if you write a JIRA ticket than it would be if I did.

gigawhitlocks:
Please let me know what y'all think :point_up: @gabrieljackson @stylianosrigas @stafot

gabrieljackson:
@stafot @stylianosrigas please review the comments when you get some time. Thanks!

stylianosrigas:
@gigawhitlocks Thanks a lot my friend for the detailed response and sorry for the delay to review. I think the proposed changes make lot of sense and they will improve the experience. I am 2/5 on adding this --force flag that will just reprovision all utilities when passed. This way we always reprovision with commit hashes for normal reasons but when sth bad happens and we need to delete a utility and redeploy we can just use that flag and reprovision all utilities like we do now. How difficult will it be for you to add this?

gigawhitlocks:
Just a heads up @stylianosrigas and @gabrieljackson after digging into this a bit, adding a `--force-reprovision` flag does seem more involved than I thought at first, unfortunately, because the flag has to actually be passed as part of the `ProvisionClusterRequest`. I do still plan to make this change as discussed during planning but I wanted y'all to be aware it might take me a little longer than I originally anticipated.

gigawhitlocks:
OK I have addressed the feedback in the prior commit.

Overview of the changes/usage:

Now when running `cloud cluster provision`, if there is no version specified for a utility (both values file path & chart version must be specified), the utility will be skipped and this will show up in the logs:

```
INFO[2021-11-19T12:48:25-06:00] Skipping reprovision of utility "stackrox-secured-cluster-services"  instance=wh6c7a7cy3bw5q3byuznph71ua provisioner=kops utility-group=UpgradeManifests
INFO[2021-11-19T12:48:25-06:00] Skipping reprovision of utility "kubecost"    instance=wh6c7a7cy3bw5q3byuznph71ua provisioner=kops utility-group=UpgradeManifests
INFO[2021-11-19T12:48:25-06:00] Skipping reprovision of utility "node-problem-detector"  instance=wh6c7a7cy3bw5q3byuznph71ua provisioner=kops utility-group=UpgradeManifests
```

If you specify a version for any utility, it will be (re)provisioned to that version

If `--reprovision-all-utilities` is specified, the original behavior will be followed: utilities whose versions are explicitly specified will be provisioned to those versions, and if not explicitly specified, the utility will be re-provisioned to the existing versions.

Let me know what you think @stylianosrigas and @gabrieljackson

gigawhitlocks:
Did not mean to close the PR lol, missed the comment button

model/cluster_utility_test.go
```diff
@@ -122,9 +103,9 @@ func TestSetDesired(t *testing.T) {
 			},
 			desiredVersions: nil,
 			expectedDesiredVersions: UtilityGroupVersions{
-				PrometheusOperator: &HelmUtilityVersion{ValuesPath: "prom", Chart: "1.0.0"},
-				Nginx:              &HelmUtilityVersion{ValuesPath: "nginx", Chart: "2.0.0"},
-				Teleport:           &HelmUtilityVersion{ValuesPath: "teleport", Chart: "5.0.0"},
+				PrometheusOperator: &HelmUtilityVersion{ValuesPath: "prometeus", Chart: "0.1"},
```
stylianosrigas, line 106:
```suggestion
				PrometheusOperator: &HelmUtilityVersion{ValuesPath: "desired-prometheus", Chart: "0.1"},
```

stylianosrigas (approved):
Just a minor fix, great work Ian thanks :)

gabrieljackson (approved):
If this behavior works for the SRE team then LGTM.

PR #568, opened 2021-11-08
Update CRDs instead of patching

Szymongib:
#### Summary

In some cases `Patching` CRD is not analogous to `Get` + `Update` as some fields are not handled properly.
Given that we fetch resources prior anyway and there should not be conflicting updates on CRDs we might as well rewrite `resourceVersion` and do `Update` to avoid issues with `Patch`.

Example issue:
```
"error":"CustomResourceDefinition.apiextensions.k8s.io \"mysqlclusters.mysql.presslabs.org\" is invalid: spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema"
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39916

#### Release Note

```release-note
Update CRDs instead of patching
```

### DIFF
--- a/k8s/custom_resources.go
+++ b/k8s/custom_resources.go
@@ -22,7 +22,7 @@ import (
 
 func (kc *KubeClient) createOrUpdateCustomResourceDefinitionBetaV1(crd *apixv1beta1.CustomResourceDefinition) (metav1.Object, error) {
 	ctx := context.TODO()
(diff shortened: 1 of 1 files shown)


spirosoik (approved):

gabrieljackson (approved):
Very interesting behavior. Thanks for the fix.

PR #569, opened 2021-11-09
Change k8s resource calculations

gabrieljackson:
Previously, resource calculations done on k8s clusters would take
all pods and nodes into account. This is an issue as custom kops
instance groups and master nodes could not run standard workloads
and should be ignored.

This change alters the calculation to only check the standard
worker nodes and the pods that are running on those nodes.

Fixes https://mattermost.atlassian.net/browse/MM-39075

```release-note
Change k8s resource calculations
```


Szymongib (approved):
LGTM!

This should definitely fix the issue but may impact installation creation time, this might be something worth looking into a little more as a future optimization.

stylianosrigas (approved):
Yessss!! Thanks a lot @gabrieljackson for that <3

PR #571, opened 2021-11-12
Adjust DB migration to Mattermost v6

Szymongib:
#### Summary

As `mattermost config` command was removed in v6, we need another way to change the config.
This is a hacky solution but the only other option that we see is the direct operation on DB which might also be problematic (config is a single DB column, we would need to always match MM version when unmarshalling it, etc.).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40053

#### Release Note

```release-note
Adjust DB migration to Mattermost v6
```


stafot (approved):

stylianosrigas (approved):
Thanks a lot for that @Szymongib 🚀

gabrieljackson (approved):
👍

PR #572, opened 2021-11-16
Cleanup database migration backups

gabrieljackson:
Removes the backups from before the pgbouncer refactor.

```release-note
Cleanup database migration backups
```


Szymongib (approved):
👍

PR #578, opened 2021-12-01
Implement event Producer and Deliverer

Szymongib:
#### Summary

This PR introduces events Producer and Deliverer as two main components responsible for creating and delivering events.

When an event is produced the old-fashioned webhook will be sent too for now. This should ensure compatibility between the two mechanisms as well as will reduce the impact on other parts of the code.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40427

#### Release Note

```release-note
Implement event Producer and Deliverer
```

### DIFF
--- a/internal/events/delivery.go
+++ b/internal/events/delivery.go
@@ -24,7 +24,7 @@ import (
 const (
 	contentTypeApplicationJSON = "application/json"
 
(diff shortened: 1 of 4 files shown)


gabrieljackson (approved):
Very nice!

PR #600, opened 2022-01-21
Cleanup Teleport utility destruction

Szymongib:
#### Summary

Cleanup destruction of resources that are no longer created.

Due to missing permission Provisioner fails to delete cluster. For all clusters created prior to migration to new Teleport we should cleanup resources manually.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41199

#### Release Note

```release-note
Cleanup Teleport utility destruction
```

### DIFF
--- a/internal/provisioner/teleport.go
+++ b/internal/provisioner/teleport.go
@@ -103,15 +103,6 @@ func (n *teleport) Destroy() error {
 		return errors.Wrap(err, "unable to delete Teleport bucket")
 	}
 
-	err = n.awsClient.DynamoDBEnsureTableDeleted(teleportClusterName, n.logger)
-	if err != nil {
-		return errors.Wrap(err, "unable to delete Teleport dynamodb table")
-	}
-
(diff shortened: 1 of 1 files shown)


stafot (approved):
LGTM

internal/provisioner/teleport.go
```diff
@@ -103,15 +103,6 @@ func (n *teleport) Destroy() error {
 		return errors.Wrap(err, "unable to delete Teleport bucket")
 	}
 
-	err = n.awsClient.DynamoDBEnsureTableDeleted(teleportClusterName, n.logger)
```
stylianosrigas, line 106: @Szymongib Can you remove the lines above as well that remove the Teleport bucket?

Szymongib: @stylianosrigas Sure, I was considering this however there is no permission issue and S3 buckets still exist for old clusters so I decided to leave it as it is so that the bucket will be cleaned up when a cluster is deleted.

stylianosrigas (approved):

PR #605, opened 2022-02-01
Use binary_parameters for all Postgres databases

agnivade:
This enforces more testing with that flag and failures
should get caught much faster.

### DIFF
--- a/internal/tools/aws/helpers_sql.go
+++ b/internal/tools/aws/helpers_sql.go
@@ -32,10 +32,13 @@ func RDSMySQLConnString(schema, endpoint, username, password string) string {
 
 // MattermostPostgresConnStrings formats the connection strings used by Mattermost
 // servers to access a PostgreSQL database.
+//
+// Regarding binary_parameters:
+// https://blog.bullgare.com/2019/06/pgbouncer-and-prepared-statements
 func MattermostPostgresConnStrings(schema, username, password string, dbCluster *rds.DBCluster) (string, string) {
-	dbConnection := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
+	dbConnection := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10&binary_parameters=yes",
 		username, password, *dbCluster.Endpoint, schema)
-	readReplicas := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
+	readReplicas := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10&binary_parameters=yes",
 		username, password, *dbCluster.ReaderEndpoint, schema)
 
 	return dbConnection, readReplicas


gabrieljackson (approved):
👍

internal/tools/aws/helpers_sql.go
```diff
@@ -32,10 +32,13 @@ func RDSMySQLConnString(schema, endpoint, username, password string) string {
 
 // MattermostPostgresConnStrings formats the connection strings used by Mattermost
 // servers to access a PostgreSQL database.
+//
+// Regarding binary_parameters:
+// https://blog.bullgare.com/2019/06/pgbouncer-and-prepared-statements
 func MattermostPostgresConnStrings(schema, username, password string, dbCluster *rds.DBCluster) (string, string) {
-	dbConnection := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10",
+	dbConnection := fmt.Sprintf("postgres://%s:%s@%s:5432/%s?connect_timeout=10&binary_parameters=yes",
```
Szymongib, line 39: I think it is a good idea to have it aligned with how things work with Pgbouncer but I am a little hesitant on adding this here as `binary_parameters` is not a recognized parameter in Postgres tools. It works with Postgres drivers which impacts how parameters are sent but does not work with any tools like `pg_isready`, `pg_dupm`, `pg_restore` etc.

We add least need to remove it for `DB_CONNECTION_CHECK_URL` passed to the secret used by the Operator as we use `pg_isready` in init container. We do the same for Pgbouncer installations [here](https://github.com/mattermost/mattermost-cloud/blob/c7458c4e7b7a1a84b7f248caa2d466620d2b2883/internal/tools/aws/helpers_sql.go#L54-L55)

Besides that adding it to all connection strings for multitenant Postgress will break DB migrations (moving logical databases from one DB cluster to another), unless we start stripping the param in a few places as those relies on `pg_dump` and `pg_restore`.

Perhaps we could have some flag to trigger it on or off on Provisioner level but for it to take effect we would need to roll all Installations which is not ideal. I guess another option could be enabling it in on a product level with some env var (or Provisioner Installation level with some new config), however, this does not seem good either as it would be another setting impacting connection string.

I am not opposed to this change, however, we need to be aware of its implications or other potential changes we would have to make.

agnivade: Thanks @Szymongib for the context! The idea was to just do this for spinwicks and not production installations. I am not very well-versed with the codebase, so my apologies for making a wider impact than needed. Do you think doing this just for spinwicks will have any knock-off effects?

Szymongib: No worries @agnivade 🙂 
I think we should be fine if we do it just for spinwicks as we do not plan to do any DB migrations there. I think new spinwicks already run on Pgbouncer tho (not 100% sure on that so perhaps @gabrieljackson could confirm).

agnivade: Great! Standing by.

agnivade: @gabrieljackson - could you confirm please? Thanks! :slightly_smiling_face:

spirosoik: yes new spinwicks already ran in PGBouncer.

agnivade: Ok perfect. Then there's no need for this. Thanks for the confirmation!

PR #610, opened 2022-02-17
Add cluster installation duration metrics

gabrieljackson:
This adds duration metrics for cluster installation processing. It
also corrects a few cases where cluster installation events were
not created on state change.

Fixes https://mattermost.atlassian.net/browse/MM-41915

```release-note
Add cluster installation duration metrics
```


Szymongib (approved):
👍

PR #615, opened 2022-03-01
Increase S3 import chunk size

gabrieljackson:
This allows for S3 file transfers from one bucket to another to
run in a more performant manner and to avoid issues where the
import bundle is too large to fit into 10000 chunks which is the
current limit.

https://mattermost.atlassian.net/browse/MM-42205

```release-note
Increase S3 import chunk size
```


Mshahidtaj (approved):
LGTM

stylianosrigas (approved):

PR #617, opened 2022-03-02
Update DNS logic to use both Route53 and Cloudflare

angeloskyratzakos:
#### Summary
Update DNS logic to use both Route53 and Cloudflare

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41494

#### Release Note

```release-note
use both route53 and cloudflare for DNS
```


angeloskyratzakos:
\update-branch

internal/tools/cloudflare/dns.go
```diff
@@ -0,0 +1,161 @@
...
+		Type:    "CNAME",
+		Content: dnsEndpoint,
+		TTL:     1,
+		Proxied: &proxied,
+	})
+	if err != nil {
+		logger.WithError(err).Error("failed to create DNS Record at Cloudflare")
+		return err
+	}
+	fmt.Println(recordResp)
```
gabrieljackson, line 113: Looks like a debug print we should remove?

gabrieljackson (commented):
Looks like we need to get tests passing on this before running an in-depth review. Also, looks like a test file called `blah.txt` got added that we should remove.

internal/supervisor/installation.go
```diff
@@ -643,7 +653,19 @@ func (s *InstallationSupervisor) configureInstallationDNS(installation *model.In
...
-		logger.WithError(err).Error("Failed to create DNS CNAME record")
+		logger.WithError(err).Error("Failed to create DNS CNAME record in Route53")
+		return model.InstallationStateCreationDNS
+	}
+
+	publicHostedZoneNameList := s.aws.GetPublicHostedZoneNames()
+	if err != nil {
+		logger.WithError(err).Error("Failed to get public hosted zone name list from Route53")
+		return model.InstallationStateCreationDNS
+	}
```
Szymongib, line 664: I think this could be called internally in `cloudflareClient.CreateDNSRecord`.

internal/tools/cloudflare/dns.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Cloudflarer interface that holds Cloudflare functions
+type Cloudflarer interface {
+	ZoneIDByName(zoneName string) (string, error)
+	DNSRecords(ctx context.Context, zoneID string, rr cf.DNSRecord) ([]cf.DNSRecord, error)
+	CreateDNSRecord(ctx context.Context, zoneID string, rr cf.DNSRecord) (*cf.DNSRecordResponse, error)
+	DeleteDNSRecord(ctx context.Context, zoneID, recordID string) error
+}
+
+// Client struct
```
Szymongib, line 29:
```suggestion
// Client is a wrapper on to of Cloudflare library client.
```

internal/tools/cloudflare/dns.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+func (c *Client) getZoneID(zoneName string) (zoneID string, err error) {
+	zoneID, err = c.cfClient.ZoneIDByName(zoneName)
+	if err != nil {
+		return "", err
+	}
+
+	return zoneID, err
```
Szymongib, line 47: Lets be explicit when returning `nil`.
```suggestion
	return zoneID, nil
```

internal/tools/cloudflare/dns.go
```diff
@@ -0,0 +1,164 @@
...
+	zoneName, found := c.getZoneName(zoneNameList, customerDNSName)
+	if !found {
+		return errors.Errorf("hosted zone for %q domain name not found", customerDNSName)
+	}
+
+	// Fetch the zone ID
+	zoneID, err := c.getZoneID(zoneName)
+	if err != nil {
+		logger.WithError(err).Error("failed to fetch Zone ID from Cloudflare")
+		return err
```
Szymongib, line 140: In general, we do not want to log an error and then return it. Let's wrap it and return to the caller to handle.

internal/tools/cloudflare/dns.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	recordID, err := c.getRecordID(zoneID, customerDNSName, logger)
+	if err != nil {
+		logger.WithError(err).Errorf("Failed to get record ID from Cloudflare for DNS: %s", customerDNSName)
+		return err
+	}
+
+	// Unable to find any record, skipping deletion
+	if err == nil && recordID == "" {
```
Szymongib, line 150: We return earlier when `err` is not nil so I think we can drop this check.
```suggestion
	if recordID == "" {
```

internal/tools/cloudflare/dns_test.go
```diff
@@ -0,0 +1,506 @@
...
+	for _, s := range samples {
+		t.Run(s.description, func(t *testing.T) {
+			mockCF.mockGetZoneName = s.setupName
+			mockCF.mockGetZoneID = s.setupZoneID
+			mockCF.mockGetRecordID = s.setupRecordID
+			mockCF.mockDeleteDNSRecord = s.setupDeleteDNSRecord
+			mockCF.mockDNSRecords = s.setupDNSRecord
+			client := NewClientWithToken(mockCF)
+			err := client.DeleteDNSRecord(s.customerDNSName, s.zoneNameList, logger)
+			if err != nil {
```
Szymongib, line 501: What if we expect an error but got nil? The test would pass in such a case but it should not.
```suggestion
			if s.expected != nil {
```

internal/tools/cloudflare/dns_test.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	for _, s := range samples {
+		t.Run(s.description, func(t *testing.T) {
+			mockCF.mockGetZoneName = s.setupName
+			mockCF.mockGetZoneID = s.setupID
+			mockCF.mockCreateDNSRecord = s.setupDNS
+			client := NewClientWithToken(mockCF)
+			err := client.CreateDNSRecord(s.customerDNSName, s.zoneNameList, s.dnsEndpoints, logger)
+			if err != nil {
```
Szymongib, line 349:
```suggestion
			if s.expected != nil {
```
Perhaps we could even skip the if completely and use just:
```go
assert.Equal(t, s.expected, err)
```
This should check if errors are the same but also pass if both sides are `nil`.

Szymongib (commented):
Looks good, few suggestions from my side.

internal/tools/cloudflare/dns_test.go
angeloskyratzakos: It was failing with the `Equal` the `EqualError` I think it works fine

internal/tools/cloudflare/dns_test.go
```diff
@@ -0,0 +1,536 @@
...
+	for _, s := range samples {
+		t.Run(s.description, func(t *testing.T) {
+			mockCF.mockDNSRecords = s.setup
+			client := NewClientWithToken(mockCF, nil)
+			name, err := client.getRecordID(s.zoneID, s.customerDNSName, logger)
+			result := Expected{name, err}
+			if err != nil {
+				assert.EqualError(t, s.expected, err.Error())
+				return
+			}
```
Szymongib, line 196: Not sure how is this supposed to work 🤔

Szymongib, line 196: Not sure how is this supposed to work 🤔

internal/tools/cloudflare/dns_test.go
```diff
@@ -0,0 +1,536 @@
...
+		t.Run(s.description, func(t *testing.T) {
+			mockCF.mockGetZoneName = s.setupName
+			mockCF.mockGetZoneID = s.setupID
+			mockCF.mockCreateDNSRecord = s.setupDNS
+			MockAWS.mockGetPublicHostedZoneNames = s.awsZoneNameList
+			client := NewClientWithToken(mockCF, MockAWS)
+			err := client.CreateDNSRecord(s.customerDNSName, s.dnsEndpoints, logger)
+			if s.expected != nil {
+				assert.EqualError(t, s.expected, err.Error())
+			}
```
Szymongib, line 368: Can we use `assert.Equal` here too?

internal/tools/cloudflare/dns_test.go
```diff
@@ -0,0 +1,536 @@
...
+			mockCF.mockGetRecordID = s.setupRecordID
+			mockCF.mockDeleteDNSRecord = s.setupDeleteDNSRecord
+			mockCF.mockDNSRecords = s.setupDNSRecord
+			mockAWS.mockGetPublicHostedZoneNames = s.awsZoneNameList
+			client := NewClientWithToken(mockCF, mockAWS)
+			err := client.DeleteDNSRecord(s.customerDNSName, logger)
+			fmt.Println(err)
+			if s.expected != nil {
+				assert.EqualError(t, s.expected, err.Error())
+			}
```
Szymongib, line 533: Same here, can we use `assert.Equla`? Also we should probably remove `fmt.Println`.

Szymongib, line 533: Same here, can we use `assert.Equla`? Also we should probably remove `fmt.Println`.

gabrieljackson (commented):
@angeloskyratzakos can you remove the empty `blah.txt` file that got included in this PR?

cmd/cloud/server.go
```diff
@@ -357,6 +361,18 @@ var serverCmd = &cobra.Command{
 
 		eventsProducer := events.NewProducer(sqlStore, eventsDeliverer, awsClient.GetCloudEnvironmentName(), logger)
 
+		var cloudflareClient supervisor.Cloudflarer
+		if cloudflareToken := os.Getenv("CLOUDFLARE_API_TOKEN"); cloudflareToken != "" {
+			cfClient, err := cf.NewWithAPIToken(cloudflareToken)
+			if err != nil {
+				return errors.Wrap(err, "Failed to initialize cloudflare client using API token")
```
gabrieljackson, line 368: nit:
```suggestion
				return errors.Wrap(err, "failed to initialize cloudflare client using API token")
```

There are a few more below as well.

gabrieljackson (approved):
Looks good!

PR #618, opened 2022-03-04
Add `PriorityEnv` support to Installations

Szymongib:
#### Summary

This PR adds support for priority environment variables to Provisioner.

When environtment variable is set in `PriorityEnv` it will take precedence over `MattermostEnv`and over group configuration, allowing for customization of certain things without need of removing installation from the group.

#### Release Note

```release-note
Add `PriorityEnv` support to Installations
```


gabrieljackson (approved):
Looks great!

PR #622, opened 2022-03-11
Add delete multi-tenant database endpoint

Szymongib:
#### Summary

This PR adds a new endpoint to mark the multitenant database as deleted.
Currently there is no endpoint to do that therefore deleting database needs 
to be done from DB level.

When deleting multitenant database, Provisioner will check if cluster does not
exist and will fail to mark DB as deleted unless `force` query param is set to `true`.

#### Release Note

```release-note
Add delete multi-tenant database endpoint
```


gabrieljackson (approved):
Nice! Thanks for adding this.

PR #624, opened 2022-03-21
Increase timeout k8s-spot-termination-handler

stafot:
Issue: MM-42644

#### Summary
Increase timeout k8s-spot-termination-handler.
This aims to fix some random failures we noticed on re-provisioning and on e2e clusters.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42644

#### Release Note

```release-note
Increase timeout k8s-spot-termination-handler.
This aims to fix some random failures we noticed on re-provisioning and on e2e clusters.
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -20,7 +20,6 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/pkg/errors"
-	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -291,7 +290,7 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 		return errors.Wrap(err, "failed to create bifrost secret")
 	}
 
-	// Need to remove two items from the calico because the fields after the creation are immutable so the
+	// Need to remove two items from the calico because the fields after the creation are immutable so
 	// create/update does not work. We might want to refactor this in the future to avoid this
 	logger.Info("Cleaning up some calico resources to reapply")
 	err = k8sClient.Clientset.CoreV1().Services("kube-system").Delete(ctx, "calico-typha", metav1.DeleteOptions{})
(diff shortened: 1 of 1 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -480,41 +481,50 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
...
-						Path:  "/spec/template/spec/containers/0/env/" + strconv.Itoa(i) + "/value",
-						Value: cluster.ID,
-					}}
+			attempts := 5
+			var daemonSetObj *appsv1.DaemonSet
+			for attempt := 0; attempt < attempts; attempt++ {
+				daemonSetObj, err = k8sClient.Clientset.AppsV1().DaemonSets(namespace).Get(ctx, daemonSet, metav1.GetOptions{})
+				if err != nil && attempt < 5 {
+					time.Sleep(5 * time.Second)
+					continue
```
Szymongib, line 490: I think the issue is actually here https://github.com/mattermost/mattermost-cloud/pull/624/files#diff-c47cc93d0609e79ef543c8fb97a4e2e1c222380084d5340f6cd8441e75b7fad7R269-R271

We initialize context with a timeout of 61 seconds, we shadow this variable in couple places, but here we probably refer to the context with the 61 seconds timeout, therefore if all other things will not complete in a given time the query will fail with `context deadline exceeded` as we observed it.

I would suggest making new context for each Kubernetes API based on the context we want to have for whole provisioning, which probably should be more than 61 seconds. I think that should solve the issue if not we can proceed with retries.

stafot: @Szymongib so to clarify your proposal is to create new context for this or to increase the wait time to 120 f.e?

Szymongib: Yeah, so we can start by increasing the wait time, we might need to cleanup/standardize how we handle context there.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -474,47 +474,52 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
...
-						k8s.PatchStringValue{
-							Op:    "replace",
-							Path:  "/spec/template/spec/containers/0/env/" + strconv.Itoa(i) + "/value",
-							Value: os.Getenv(model.MattermostChannel),
-						})
-				}
-				if envVar.Name == "MATTERMOST_WEBHOOK" && len(os.Getenv(model.MattermostWebhook)) > 0 {
-					payload = append(payload,
-						k8s.PatchStringValue{
+			if daemonSetObj.Spec.Selector != nil {
```
Szymongib, line 492: Could you help me understand why we are adding this check?

stafot: When in a cluster we don't have spot instances at all this selector expected to be nil. So in this case it is faster to bypass this block and continue to the next steps of provisioning.

Szymongib (approved):

PR #625, opened 2022-03-21
Use new CR ingress config

Szymongib:
#### Summary

Use new CR ingress config instead of deprecated fields.
Update Provisioner to use `ingressClassName` field instead of annotation.

#### Release Note

```release-note
Use new CR ingress config
```


stylianosrigas (approved):
@Szymongib LGTM, how does it handle updates on existing installations? It requires a group release or sth?

Szymongib:
@stylianosrigas Yes, existing ingresses will require an update for chage to take effect.

gabrieljackson (approved):

PR #632, opened 2022-04-07
Add cluster resource cache

gabrieljackson:
This new cache speeds up cluster selection when installations are
scheduled with the balanced scheduling server option.

Fixes https://mattermost.atlassian.net/browse/MM-43190

```release-note
Add cluster resource cache
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -6,6 +6,7 @@ package supervisor
 
 import (
 	"strings"
+	"time"
 
 	"github.com/mattermost/mattermost-cloud/internal/events"
 
@@ -83,7 +84,7 @@ type eventProducer interface {
 // provisioner abstracts the provisioning operations required by the installation supervisor.
 type installationProvisioner interface {
 	ClusterInstallationProvisioner(version string) provisioner.ClusterInstallationProvisioner
-	GetClusterResources(cluster *model.Cluster, onlySchedulable bool) (*k8s.ClusterResources, error)
+	GetClusterResources(cluster *model.Cluster, onlySchedulable bool, logger log.FieldLogger) (*k8s.ClusterResources, error)
 	GetPublicLoadBalancerEndpoint(cluster *model.Cluster, namespace string) (string, error)
 }
 
@@ -104,6 +105,15 @@ type InstallationSupervisor struct {
 	metrics           *metrics.CloudMetrics
 	eventsProducer    eventProducer
 	forceCRUpgrade    bool
+	cache             InstallationSupervisorCache
+}
+
+// InstallationSupervisorCache contains configuration and cached data for
+// cluster resources.
+type InstallationSupervisorCache struct {
+	initialized         bool
+	stop                chan bool
+	installationMetrics map[string]*k8s.ClusterResources
 }
 
 // InstallationSupervisorSchedulingOptions are the various options that control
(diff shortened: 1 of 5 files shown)


internal/supervisor/installation.go
```diff
@@ -1456,6 +1489,46 @@ func getAnnotationsNames(annotations []*model.Annotation) []string {
...
+	for {
+		clusters, err := s.store.GetClusters(&model.ClusterFilter{Paging: model.AllPagesNotDeleted()})
+		if err != nil {
+			cacheLogger.WithError(err).Warn("Failed to query clusters")
+		} else {
+			for _, cluster := range clusters {
+				clusterResources, err := s.provisioner.GetClusterResources(cluster, true, cacheLogger)
+				if err != nil {
+					cacheLogger.WithError(err).Error("Failed to get cluster resources")
+					continue
```
Szymongib, line 1515: Should we clean up the cache entry here? If something breaks in `GetClusterResources` we risk using outdated data for a potentially long time.

internal/supervisor/installation.go
```diff
@@ -1456,6 +1489,46 @@ func getAnnotationsNames(annotations []*model.Annotation) []string {
...
+				}
+				s.cache.installationMetrics[cluster.ID] = clusterResources
```
Szymongib, line 1517: We might need to use a mutex for the `map`. I am not sure if concurrent one write + multiple reads is thread-safe.

internal/supervisor/installation.go
```diff
@@ -156,10 +167,16 @@ func NewInstallationSupervisorSchedulingOptions(balanceInstallations bool, clust
 // Shutdown performs graceful shutdown tasks for the installation supervisor.
 func (s *InstallationSupervisor) Shutdown() {
 	s.logger.Debug("Shutting down installation supervisor")
+	s.cache.stop <- true
```
Szymongib, line 170: It is probably an unlikely edge case but if I am not mistaken this will block forever if we stop the supervisor before `Do` was ever called, as nothing will be listening to the channel.

We could start the cache goroutine in `NewInstallationSupervisor`. I am guessing you wanted to avoid that so that we do not hammer cache initialization on every restart, tho this may cause a tricky bug that will result in a non-graceful shutdown.

I think alternatively we can close the channel and this should trigger the receive.

Szymongib (commented):
Looks good. I think there might be a few edge cases.

internal/supervisor/installation.go
gabrieljackson: Good point. I had considered concurrent reads being okay, but reading the docs confirmed that we should lock even with one writer in the mix.

internal/supervisor/installation.go
gabrieljackson: Good catch thanks. I took a slightly different approach that I think should resolve all possible issues here. Let me know what you think.

internal/supervisor/installation.go
```diff
@@ -112,7 +113,9 @@ type InstallationSupervisor struct {
 // cluster resources.
 type InstallationSupervisorCache struct {
 	initialized         bool
+	running             bool
```
Szymongib, line 116: I think just closing the `stop` channel would be a bit simpler from a code perspective but I will leave it up to you.

Szymongib (approved):

PR #634, opened 2022-04-12
Support multiple domains for Installation

Szymongib:
#### Summary

This PR adds support for multiple domains for Installation.
All domains need to start with the newly introduced `Name` field which is unique across existing Installations.

This will require some conflict resolution with Cloudflare changes.

#### Release Note

```release-note
Support multiple domains for Installation
```


gabrieljackson (approved):
Looks good! Nice work.

PR #643, opened 2022-05-19
Add max pods per worker node configuration option

gabrieljackson:
This can be used to set the max number of pods that will run on
a kubernetes cluster node.

Fixes https://mattermost.atlassian.net/browse/MM-31253

```release-note
Add max pods per worker node configuration option
```

### DIFF
--- a/model/cluster_request.go
+++ b/model/cluster_request.go
@@ -30,6 +30,7 @@ type CreateClusterRequest struct {
 	NodeInstanceType       string                         `json:"node-instance-type,omitempty"`
 	NodeMinCount           int64                          `json:"node-min-count,omitempty"`
 	NodeMaxCount           int64                          `json:"node-max-count,omitempty"`
+	MaxPodsPerNode         int64                          `json:"max-pods-per-node,omitempty"`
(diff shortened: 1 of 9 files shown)


model/cluster_request.go
```diff
@@ -181,16 +188,20 @@ func NewUpdateClusterRequestFromReader(reader io.Reader) (*UpdateClusterRequest,
 
 // PatchUpgradeClusterRequest specifies the parameters for upgrading a cluster.
 type PatchUpgradeClusterRequest struct {
-	Version       *string        `json:"version,omitempty"`
-	KopsAMI       *string        `json:"kops-ami,omitempty"`
-	RotatorConfig *RotatorConfig `json:"rotatorConfig,omitempty"`
+	Version        *string        `json:"version,omitempty"`
+	KopsAMI        *string        `json:"kops-ami,omitempty"`
+	MaxPodsPerNode *int64         `json:"max-pods-per-node,omitempty"`
```
Szymongib, line 193: nit: I think in most places we use camel case, so unless there are some specific reasons against it I would stick to that.

Szymongib (approved):
LGTM, one question in regards to `UpgradeCluster`.

PR #657, opened 2022-05-27
Add endpoints to modify Group annotations

Szymongib:
#### Summary

Add endpoints and commands to add and remove annotations from Groups.

#### Release Note

```release-note
Add endpoints to modify Group annotations
```


gabrieljackson (approved):
Looks good!

PR #663, opened 2022-06-07
fix bug with rtcd helm chart prefix

angeloskyratzakos:
#### Summary
Reprovisioning fails due to this bug as the prefix does not match the already deployed

#### Release Note

```release-note
fix bug with rtcd helm chart prefix
```


stylianosrigas (approved):
LGTM

PR #666, opened 2022-06-15
Allow changing Installation group based on annotations

Szymongib:
#### Summary

Support changing installation group based on annotations.
```
cloud group assign --group-selection-annotation [ANNOTATIONS] --installation [INSTALLATION]
```

#### Release Note

```release-note
Allow changing Installation group based on annotations
```

### DIFF
--- a/cmd/cloud/group.go
+++ b/cmd/cloud/group.go
@@ -55,6 +55,11 @@ func init() {
 	groupJoinCmd.MarkFlagRequired("group")
 	groupJoinCmd.MarkFlagRequired("installation")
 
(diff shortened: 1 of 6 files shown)


gabrieljackson (commented):
Hmm, it looks like we are using a separate API endpoint from the original group join in order to maintain backwards compatibility. Is that correct? If so, does it make sense to at least combine the CLI functionality to possibly lower confusion a bit? My concern is that `group join` and `group assign` may be a bit hard to understand or one will simply be missed when reviewing commands.

gabrieljackson (approved):
👍

PR #667, opened 2022-06-16
Optimize DNS search path

stafot:
Optimize search path (aka ndots option) in Mattermost Cloud Workspaces

Issue: CLD-2512

#### Summary
Optimize search path (aka ndots option) in Mattermost Cloud Workspaces

- Kubelet sets up in every pod by default in /etc/resolv.conf the option ndots:5 (don’t resolve a domain name as is unless it is FQDN)
- This value is convenient but not efficient. Might do more than 5 queries per DNS lookup
- Linux default is ndots:1 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-2512

#### Release Note

```release-note
Optimize search path (aka ndots option) in Mattermost Cloud Workspaces
```


phoinixgrr (approved):
🎖️ 🏅 🎖️ 

Awesome work @stafot  👍

PR #675, opened 2022-07-05
Add flags to configure max Installations per DB cluster

Szymongib:
#### Summary

This adds 3 flags for different types of DBs:
- `max-installations-rds-postgres-pgbouncer`
- `max-installations-rds-postgres`
- `max-installations-rds-mysql`

It defaults to the hardcoded constants. If value is set to 0, the default const will be used as well.

#### Release Note

```release-note
Add flags to configure max Installations per DB cluster
```


stylianosrigas (approved):
This is awesome, thanks @Szymongib . Please ping me when merged to change the flags in Argo.

PR #681, opened 2022-07-12
Remove Blackbox related logic from hibernation flow

Szymongib:
#### Summary
Remove Blackbox related logic from hibernation flow. This is the first step to clean up Route53 dependency (if we decide to remove it completely).

We should wait with the merge until Blackbox is removed fully.

#### Release Note

```release-note
Remove Blackbox related logic from hibernation flow
```


stylianosrigas (approved):
LGTM

PR #683, opened 2022-07-13
Skip cluster resource check for certain states

gabrieljackson:
These states indicate the cluster isn't ready to have installations
scheduled on it.

Fixes https://mattermost.atlassian.net/browse/MM-45694

```release-note
Skip cluster resource check for certain states
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -1618,12 +1618,15 @@ func (s *InstallationSupervisor) manageInstallationResourcesCache() {
 			cacheLogger.WithError(err).Warn("Failed to query clusters")
 		} else {
 			for _, cluster := range clusters {
-				// Filter out a few states where we shouldn't try to obtain
-				// cluster resources.
-				if cluster.State == model.ClusterStateCreationRequested ||
-					cluster.State == model.ClusterStateCreationFailed ||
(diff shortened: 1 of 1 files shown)


internal/supervisor/installation.go
```diff
@@ -1618,6 +1618,15 @@ func (s *InstallationSupervisor) manageInstallationResourcesCache() {
 			cacheLogger.WithError(err).Warn("Failed to query clusters")
 		} else {
 			for _, cluster := range clusters {
+				// Filter out a few states where we shouldn't try to obtain
```
Szymongib, line 1621: What about other state such as `provisioning-failed` or `upgrade-failed` etc? Perhaps we should only calculate the resources when the cluster is `stable`?

gabrieljackson: Hmm, in theory the kubernetes API should be online and functioning even during these states. I know that installations won't be scheduled on clusters in these states, but we should be able to retrieve the resource totals still. I figured it was good to continue to monitor them so that if a dramatic change in resources available occurred during the non-stable time window that we captured that in the cache. What do you think?

gabrieljackson: Yep, this makes sense. Thanks for the idea. I will add it to be safe.

Szymongib (approved):
LGTM

PR #690, opened 2022-07-25
Support custom Installation sizes in Provisioner

Szymongib:
#### Summary
This PR adds support for custom Installation sizes in Provisioner.
The size should be specified in form:
```
[SIZE_NAME]-[NUMBER_OF_MM_REPLICAS]
```
If the number of replicas is not specified the default value for the size will be used.

Currently, there is only one size added, we can add more according to our needs.

#### Release Note

```release-note
Support custom Installation sizes in Provisioner
```

### DIFF
--- a/model/mattermost_size_test.go
+++ b/model/mattermost_size_test.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package model
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-operator/apis/mattermost/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseProvisionerSize(t *testing.T) {
+
+	provisionerXL6Replicas := SizeProvisionerXLResources
+	provisionerXL6Replicas.App.Replicas = 6
+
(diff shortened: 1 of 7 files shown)


model/mattermost_size_test.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		{
+			description:  "correct size, custom replicas (6)",
+			size:         "provisionerXL-6",
+			expectedSize: provisionerXL6Replicas,
+		},
+		{
+			description:  "correct size, ignore other vals",
+			size:         "provisionerXL-1-6",
+			expectedSize: provisionerXL1Replica,
```
gabrieljackson, line 41: Hmm, I believe we may want to explicitly block this so we don't end up with size values stored that have extra ignored characters at the end. Something like checking `len(parts) > 2` so we avoid possibly seeing values like `provisionerXL-3-34-333-3`.

gabrieljackson (commented):
Looks good. I just have one concern about how size parsing is done.

model/mattermost_size_test.go
Szymongib: Yeah, that was intentional to not make validation to strict in case we want to extend this logic further, but yeah we can adjust validation then.

gabrieljackson (approved):
Looks good!

PR #691, opened 2022-07-26
Add cluster resource threshold scheduling overrides

gabrieljackson:
These overrides can be used to tweak the threshold value used in
cluster scheduling for individual resource types.

If this type of scheduling behavior works well then the values will
be moved to the cluster resource.

Fixes https://mattermost.atlassian.net/browse/MM-45849

```release-note
Add cluster resource threshold scheduling overrides
```

### DIFF
--- a/internal/supervisor/installation.go
+++ b/internal/supervisor/installation.go
@@ -134,7 +134,9 @@ type InstallationSupervisorCache struct {
 // how installation scheduling occurs.
 type InstallationSupervisorSchedulingOptions struct {
 	balanceInstallations               bool
-	clusterResourceThreshold           int
+	clusterResourceThresholdCPU        int
+	clusterResourceThresholdMemory     int
+	clusterResourceThresholdPodCount   int
 	clusterResourceThresholdScaleValue int
 }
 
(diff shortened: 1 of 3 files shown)


internal/supervisor/installation.go
```diff
@@ -172,12 +174,43 @@ func NewInstallationSupervisor(
 }
 
 // NewInstallationSupervisorSchedulingOptions creates a new InstallationSupervisorSchedulingOptions.
-func NewInstallationSupervisorSchedulingOptions(balanceInstallations bool, clusterResourceThreshold, clusterResourceThresholdScaleValue int) InstallationSupervisorSchedulingOptions {
-	return InstallationSupervisorSchedulingOptions{
+func NewInstallationSupervisorSchedulingOptions(balanceInstallations bool, clusterResourceThreshold, thresholdCPUOverride, thresholdMemoryOverride, thresholdPodCountOverride, clusterResourceThresholdScaleValue int) InstallationSupervisorSchedulingOptions {
```
Szymongib, line 177: Since there is a bunch of logic now, I would add unit test for it. There is a lot of ints with similar names so it easy to slip up.

Szymongib (commented):
Looks good, I would suggest adding a unit test for overrides logic.

> If this type of scheduling behavior works well then the values will
be moved to the cluster resource.

Unless there is really compelling reason for this I would keep it as global Provisioner setting. More uniformity across configuration makes things simpler.

internal/supervisor/installation.go
gabrieljackson: Yep, good point. Tests added.

Szymongib (approved):

PR #695, opened 2022-07-28
Upgrade nginx to v4.2.0

stafot:
Issue: CLD-3792

#### Summary
Upgrade nginx to v4.2.0

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3792

#### Release Note

```release-note
Upgrade nginx to v4.2.0
```


Szymongib (approved):

spirosoik (approved):
lgtm. but may we should wait till we identify the root cause with the current issue we have for internal nginx

stafot:
@spirosoik Totally agree but the issue we have with `internal-nginx` is on version `4.0.18`. The reason I also prefer to hold off is to not introduce new variable on the play.

Mshahidtaj (approved):

PR #698, opened 2022-08-10
Fix max nodes cluster value updates

gabrieljackson:
This corrects a bug where cluster worker node max values were not
persisted. As part of the fix, all instances of manual YAML instance
group manifest manipulation were removed.

Fixes https://mattermost.atlassian.net/browse/MM-45218

```release-note
Fix max nodes cluster value updates
```


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -804,30 +802,48 @@ func (provisioner *KopsProvisioner) ResizeCluster(cluster *model.Cluster, awsCli
...
-
-		igManifest, err = grossKopsReplaceSize(
-			igManifest,
-			kopsMetadata.ChangeRequest.NodeInstanceType,
-			fmt.Sprintf("%d", changeMetadata.NodeMinCount),
-			fmt.Sprintf("%d", changeMetadata.NodeMaxCount),
-		)
-		if err != nil {
-			return errors.Wrap(err, "failed to update instance group yaml file")
+		// There is a bit of complexity with updating min and max instancegroup
```
Szymongib, line 805: I do not think this logic is tested in anyway, which makes me a bit uncomfortable as it is very easy to screw something up if we ever change something or even refactor. It may be a bit hard to do it given all the dependencies, but in my opinion, might be worth it.

gabrieljackson: Fair enough. Let me see if I can add these quickly.

Szymongib (approved):
LGTM! Thanks for adding tests!

PR #703, opened 2022-08-22
Add installation deletion limit

gabrieljackson:
When checking if an installation in the pending-deletion state
should be deleted, the supervisor now also respects a limit on the
maximum number of updating installations that are allowed at that
point in time. Any skipped installations due to hitting the limit
will be checked later and deleted.

Fixes https://mattermost.atlassian.net/browse/MM-46518

```release-note
Add installation deletion limit
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -110,7 +110,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("installation-enable-route53", false, "Specifies whether CNAME records for Installation should be created in Route53 as well.")
 	serverCmd.PersistentFlags().Bool("disable-dns-updates", false, "If set to true DNS updates will be disabled when updating Installations.")
 	serverCmd.PersistentFlags().Duration("installation-deletion-pending-time", 3*time.Minute, "The amount of time that installations will stay in the deletion queue before they are actually deleted. Set to 0 for immediate deletion.")
-	serverCmd.PersistentFlags().Int64("installation-deletion-max-updating", 25, "The maximum number of installations that can be updating when finalizing deletion of deletion-pending installations (similar to group max rolling).")
+	serverCmd.PersistentFlags().Int64("installation-deletion-max-updating", 25, "A soft limit on the number of installations that the provisioner will delete at one time from the group of deletion-pending installations.")
 
 	// DB clusters utilization configuration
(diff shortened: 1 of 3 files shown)


internal/supervisor/installation_deletion.go
```diff
@@ -66,7 +72,18 @@ func (s *InstallationDeletionSupervisor) Do() error {
...
+		s.logger.WithError(err).Warn("Failed to query for installation status")
+		return nil
+	}
+	s.currentlyUpdatingCounter = status.InstallationsUpdating
+
 	for _, installation := range installations {
+		if s.currentlyUpdatingCounter >= s.currentlyUpdatingLimit {
+			s.logger.Infof("Max installation updating counter (%d) reached for pending deletions", s.currentlyUpdatingLimit)
+			return nil
+		}
```
Szymongib, line 86: Not sure if it is the best way of doing this. Given that there are several provisioners changing installations state this limit is kind of raw estimation rather than the hard limit.

The number of processing installations can change between Supervising different installations therefore if we say there was 20 installations queried by the `Do` method, by the time we get to the 20th the number of updating installations could have changed significantly so we might be over the limit.

It might not be a huge issue but it is kind of unexpected behaviour so if we are fine with it, we should document it clearly, both with comments and in flag description.

cmd/cloud/server.go
```diff
@@ -110,6 +110,7 @@ func init() {
 	serverCmd.PersistentFlags().Bool("installation-enable-route53", false, "Specifies whether CNAME records for Installation should be created in Route53 as well.")
 	serverCmd.PersistentFlags().Bool("disable-dns-updates", false, "If set to true DNS updates will be disabled when updating Installations.")
 	serverCmd.PersistentFlags().Duration("installation-deletion-pending-time", 3*time.Minute, "The amount of time that installations will stay in the deletion queue before they are actually deleted. Set to 0 for immediate deletion.")
+	serverCmd.PersistentFlags().Int64("installation-deletion-max-updating", 25, "The maximum number of installations that can be updating when finalizing deletion of deletion-pending installations (similar to group max rolling).")
```
Szymongib, line 113: I think max group rolling is a hard limit since it is stored on the Group itself, I think with current implementation this behaves differently.

internal/supervisor/installation_deletion.go
gabrieljackson: Yep, I was aware of this being a simple soft limit, but it's definitely a good idea to clarify. Added a chunk of comments to call out this point. Thanks!

Szymongib (approved):

PR #704, opened 2022-08-23
Add API metrics

gabrieljackson:
API duration and count metrics have been added with labels for
different endpoints.

The `ResponseWriterWrapper` was taken from the implementation
in the Mattermost server in an effort to keep somewhat standardized.

Fixes https://mattermost.atlassian.net/browse/MM-46330

```release-note
Add API metrics
```


gabrieljackson:
Just realized there is one small change that is needed for path standardization. Pushing the update shortly.

internal/api/handler.go
```diff
@@ -14,24 +15,34 @@ import (
...
+		"method":  r.Method,
 		"path":    r.URL.Path,
 		"request": context.RequestID,
 	})
 
-	h.handler(context, w, r)
+	h.handler(context, ww, r)
+
+	elapsed := float64(time.Since(start)) / float64(time.Second)
+	context.Metrics.ObserveAPIEndpointDuration(h.handlerName, r.Method, ww.StatusCode(), elapsed)
```
Szymongib, line 38: Wouldn't it make more sense to use `r.URL.Path` instead of `handlerName`? I think this might be error prone. It is quite possible that we copy and paste the handler and only change the actual function not the name. Also when changing function name we might forget to update the handler name, and in case we update the handler name we suddenly created new series even tho the API path will remain the same.

Szymongib (approved):
I would prefer using `r.URL.Path` instead of handler name.

internal/api/handler.go
gabrieljackson: So I agree that it would probably be nicer to be able to use `r.URL.Path` here, but I ran into the issue I noted above.

In order to standardize prometheus labels for the different endpoints,
the api handler name is registered and used. This is done instead
of the request URL path values because many endpoints contain
IDs which end up making them unique for each resource.

When using the Path value, any path that contained a resource ID ended up being a unique label. I attempted to keep the path value and use regex to standardize the unique IDs, but it was a bit messy. I changed tactics and used the handler name instead, which is what Mattermost server does. I agree that it isn't quite as nice, but believe it's probably the best solution for us. What do you think?

gabrieljackson: Did a quick update to derive the func name via reflect. This removes the handler name duplication issue and maintains common metric labels.

Szymongib (approved):

PR #708, opened 2022-08-29
Upgrade kops to 1.22

stafot:
Issue: CLD-1543

#### Summary
Upgrade kops to 1.22

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-1543

#### Release Note

```release-note
Upgrade kops to 1.22
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -191,18 +191,6 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 
 	logger.WithField("name", kopsMetadata.Name).Info("Successfully deployed kubernetes")
 
-	logger.WithField("name", kopsMetadata.Name).Info("Updating VolumeBindingMode in default storage class")
-	k8sClient, err := k8s.NewFromFile(kops.GetKubeConfigPath(), logger)
-	if err != nil {
-		return err
-	}
-
-	_, err = k8sClient.UpdateStorageClassVolumeBindingMode("gp2")
-	if err != nil {
-		return err
-	}
-	logger.WithField("name", kopsMetadata.Name).Info("Successfully updated storage class")
-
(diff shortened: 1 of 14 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -190,18 +190,6 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
...
-	k8sClient, err := k8s.NewFromFile(kops.GetKubeConfigPath(), logger)
-	if err != nil {
-		return err
-	}
-
-	_, err = k8sClient.UpdateStorageClassVolumeBindingMode("gp2")
-	if err != nil {
-		return err
-	}
-	logger.WithField("name", kopsMetadata.Name).Info("Successfully updated storage class")
```
stafot, line 203: This block fails at the moment and in my understanding of things is totally obsolete and can be safely removed. But if your understanding is different @gabrieljackson and @Szymongib feel free to correct me.

internal/tools/kops/cluster.go
```diff
@@ -79,11 +79,11 @@ func (c *Cmd) CreateCluster(name, cloud string, kopsRequest *model.KopsMetadataR
 // Example setValue: spec.kubernetesVersion=1.10.0
 func (c *Cmd) SetCluster(name, setValue string) error {
```
gabrieljackson, line 80: There is another `SetInstanceGroup` method that should need this change: https://github.com/mattermost/mattermost-cloud/blob/master/internal/tools/kops/instance_groups.go#L221

internal/provisioner/kops_provisioner_cluster.go
gabrieljackson: Hmm, I can't remember what this was used for, but would agree that if there are no perceived issues then it should be okay to remove. The code was added by @stylianosrigas so let's check with him as well.

gabrieljackson (approved):
LGTM as long as both a cluster resize and upgrade were performed alongside creation and deletion tests.

PR #710, opened 2022-08-31
Fix sloth check on provision phase

stafot:
Issue: CLD-4062

#### Summary
Fix sloth check on provision phase

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-4062

#### Release Note

```release-note
Fix sloth check on provision phase
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -544,7 +544,7 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 
 	prom, _ := k8sClient.GetNamespace("prometheus")
 
-	if prom != nil {
+	if prom != nil && prom.Name != "" {
 		err = prepareSloth(k8sClient, logger)
 		if err != nil {
 			return errors.Wrap(err, "failed to prepare Sloth")


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -556,7 +556,7 @@ func (provisioner *KopsProvisioner) ProvisionCluster(cluster *model.Cluster, aws
 
 	prom, _ := k8sClient.GetNamespace("prometheus")
 
-	if prom != nil {
+	if prom.Name != "" {
```
Szymongib, line 559: We do not do error check from `k8sClient.GetNamespace` we also do not do `nil` check, so if `prom` ends up being `nil` this will panic.

stafot: The problem with the previous code is that prom is never nil as prom returned empty object instead of `nil`, this caused in rare cases where prometheus utility was not installed in a cluster sloth always to try to get installed while it can't as it sits in prometheus namespace. Current code fix this code as `prom.Name` can be either `prometheus` or empty string. @Szymongib

Szymongib: Yeah that makes sense but we still should do nil check since `GetNamespace` returns a pointer. The implementation of `GetNamespace` might change in the future where the `nil` will be returned instead of empty object which will lead to panic here. So something like `if prom != nil && prom.Name != "" {` should work.

stafot: Yes that makes sense

gabrieljackson (commented):
Just leaving a request to please retag me for review when the PR is ready. Thanks!

Szymongib (approved):

PR #725, opened 2022-09-28
Cleanup installation supervisor unit tests

gabrieljackson:
Much of the code for the supervisor tests is repeated for each test. This change adds some helper functions to minimize repetition and to reduce the chance for error.

```release-note
None
```


Szymongib (approved):
Very nice cleanup 👍

PR #726, opened 2022-09-30
EKS cluster creation with AWS API

Szymongib:
#### Summary

- Add new type `EKSMetadata` to `Cluster`.
- Extend AWS client with methods to create and delete EKS clusters and node groups.
- Refactor some methods to work both with Kops and EKS.
- Add some supporting methods.

#### Release Note

```release-note
EKS cluster creation with AWS API
```

### DIFF
--- a/model/cluster_request.go
+++ b/model/cluster_request.go
@@ -19,6 +19,11 @@ const (
 	NetworkingAmazon = "amazon-vpc-routed-eni"
 )
 
+var (
+	defaultEKSRoleARN       string
+	defaultNodeGroupRoleARN string
+)
+
 // CreateClusterRequest specifies the parameters for a new cluster.
 type CreateClusterRequest struct {
 	Provider               string                         `json:"provider,omitempty"`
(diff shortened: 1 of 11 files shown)


model/cluster_request.go
```diff
@@ -119,6 +143,12 @@ func (request *CreateClusterRequest) Validate() error {
 	}
 	// TODO: check zones and instance types?
 
+	if request.EKSConfig != nil {
+		if len(request.EKSConfig.NodeGroups) == 0 {
+			return errors.Errorf("at least 1 node group is required when using EKS")
```
gabrieljackson, line 148: nit
```suggestion
			return errors.New("at least 1 node group is required when using EKS")
```

gabrieljackson (commented):
Just checking that the intention is to persist the EKS metadata into the provisioner database in a follow-up PR?

Szymongib:
> Just checking that the intention is to persist the EKS metadata into the provisioner database in a follow-up PR?

Yes, that is the case. I have store part already implemented, but need to finish up handlers etc, so will include that togehter.

PR #730, opened 2022-10-07
Temporary disable calico-node daemonset check

stafot:
Issue: CLD-4268
Signed-off-by: Stavros Foteinopoulos <stafot@gmail.com>

#### Summary
Temporary disable calico-node daemonset check.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-4268

#### Release Note

```release-note
Temporary disable calico-node daemonset check
```


Mshahidtaj (approved):

stylianosrigas (approved):
We don't need this part at all. This was originally used for checking that non calico CNI clusters have calico node ready for network policies. In our case now this is not applicable so we can remove it until we think a better approach. 

Thanks @stafot 🚀

PR #744, opened 2022-10-19
Add schema count in installation report command

gabrieljackson:
If an installation is in a multitenant database we now report the number of schemas in the logical database where the installation resides.

```release-note
Add schema count in installation report command
```


fmartingr (approved):
LGTM

PR #746, opened 2022-10-21
Add server_lifetime and server_idle_timeout pgbouncer variables

stafot:
Issue: CLD-4380

#### Summary
Add `server_lifetime` and `server_idle_timeout` pgbouncer variables

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-4380

#### Release Note

```release-note
Add `server_lifetime` and `server_idle_timeout` pgbouncer variables
```


stylianosrigas (approved):
LGTM thanks :)

angeloskyr (approved):

fmartingr (approved):
GJ! LGTM.

One question though (unrelated but related to the PR). Why those variables are stored globally? (`reservePoolSize` and so on)

Also, would be nice to have a test checking that the template is generated with the correct parameters in the expected order, in case we keep adding in the future and we can mess up at some point. Plus for using Golang templates instead of string formatting. (I can take care of all this once I move on with what's already on my plate).

PR #748, opened 2022-10-21
[MM-44904] Adding installations to group based on conditions

fmartingr:
#### Summary

Refactor of the group allocation for new installations.

Previously, group allocation was performed randomly based on group annotations. This system is changed with this PR introducing the concept of a Group Allocator, which receives a list of groups and selects one based on it internal logic.

Two group allocations have been created, one Random, to preserve the original logic, and a new one to select a group based on the lowest count of total or updated installations.

Database utilization is quite heavy (since we need 6 queries to get the status for **one** group). If we only need the total installation count I could refactor the query or add a new method to the store/filter to retrieve the total installation count directly with the groups. (Which is probably the best way ahead for the time being).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44904

#### Release Note

```release-note
- Automatically add new installations to a group based on lowest installation count
```

### DIFF
--- a/model/group_allocator.go
+++ b/model/group_allocator.go
@@ -1,0 +1,63 @@
+package model
+
+import (
+	"math/rand"
+)
+
+// InstallationGroupAllocator defines a way to allocate installations into groups based on
+// each allocator internal logic.
+type InstallationGroupAllocator interface {
+	Choose(groups []*GroupDTO) (*GroupDTO, error)
+}
+
+// RandomInstallationGroupAllocator allocates the installation in a random group
+type RandomInstallationGroupAllocator struct{}
+
+func (iga *RandomInstallationGroupAllocator) Choose(groups []*GroupDTO) (*GroupDTO, error) {
+	return groups[rand.Intn(len(groups))], nil
+}
+
+// NewRandomInstallationGroupAllocator
+func NewRandomInstallationGroupAllocator() InstallationGroupAllocator {
+	return &RandomInstallationGroupAllocator{}
+}
+
+// LowestCountInstallationGroupAllocator allocates the installation in the group that has the lowest
+// count for the defined state
+type LowestCountInstallationGroupAllocator struct {
+	// the state to check
+	installationState string
+}
+
+// getStateValue gets the value for the selected state to check from the provided group
+// defaults to the total instalations silently if the provided state is not handled
+func (iga *LowestCountInstallationGroupAllocator) getStateValue(group *GroupDTO) int64 {
+	switch iga.installationState {
+	case "updated":
+		return group.Status.InstallationsUpdated
+	default:
+		return group.Status.InstallationsTotal
+	}
+}
+
+// getLowestInstallationCount retrieves the lowest installation count group based on the defined state field
+func (iga *LowestCountInstallationGroupAllocator) getLowestInstallationCount(groups []*GroupDTO) (group *GroupDTO) {
+	for _, g := range groups {
+		if group == nil || iga.getStateValue(g) < iga.getStateValue(group) {
+			group = g
+		}
+	}
+
+	return
+}
+
+func (iga *LowestCountInstallationGroupAllocator) Choose(groups []*GroupDTO) (*GroupDTO, error) {
+	return iga.getLowestInstallationCount(groups), nil
+}
+
+// NewLowestCountInstallationGroupAllocator
+func NewLowestCountInstallationGroupAllocator(state string) InstallationGroupAllocator {
+	return &LowestCountInstallationGroupAllocator{
+		installationState: state,
+	}
+}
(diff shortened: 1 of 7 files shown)


model/group_allocator.go
```diff
@@ -0,0 +1,63 @@
...
+
+// getLowestInstallationCount retrieves the lowest installation count group based on the defined state field
+func (iga *LowestCountInstallationGroupAllocator) getLowestInstallationCount(groups []*GroupDTO) (group *GroupDTO) {
+	for _, g := range groups {
+		if group == nil || iga.getStateValue(g) < iga.getStateValue(group) {
+			group = g
+		}
+	}
+
+	return
```
gabrieljackson, line 51: nit: I totally get using a named return here, but small request to always explicitly return values for code in this repo just to keep things consistent.

model/group_allocator.go
```diff
@@ -0,0 +1,63 @@
...
+// getStateValue gets the value for the selected state to check from the provided group
+// defaults to the total instalations silently if the provided state is not handled
+func (iga *LowestCountInstallationGroupAllocator) getStateValue(group *GroupDTO) int64 {
+	switch iga.installationState {
+	case "updated":
+		return group.Status.InstallationsUpdated
+	default:
+		return group.Status.InstallationsTotal
+	}
+}
```
gabrieljackson, line 41: After thinking this over a bit, let's keep things simple for now and just always use `Status.InstallationsTotal` for selecting the lowest count. I tried to think through if we may ever want to prioritize a certain state, but that seems pretty unlikely. If we ever need it, it would be simple to add.

gabrieljackson (commented):
This is coming along nicely. I left a few comments for you to review.

model/group_allocator.go
fmartingr: Perfect, that'll simplify things a lot.

fmartingr: After thinking about it a bit, I've removed the concept of the group allocator. if this logic is not going to change in the foreseable future this new concept only adds unnecessary complexity, so I've removed it and moved the logic to the `selectGroupFromAnnocations` method and added the test directly on the `api/installations/create` tests. There's still the thing with the SQLs. Not sure if you wanted me to approach that on this PR, another one or just disregard that entirely.

gabrieljackson (approved):
This looks great! 👍 

As you mentioned in your PR description we could change/add something to only return the total installation count to be a bit more performant. I think this is okay for now so I will leave that up to you. It would probably be nice to have, but we can also add it in a follow-up PR.

fmartingr:
I did some benchmarks only to scratch an itch and triple check that joins are way faster than anything else.

```
BenchmarkSelectOnlyAll-10                   1286            827112 ns/op
BenchmarkSelectWithinSelectAll-10              2         874080562 ns/op
BenchmarkLeftJoinAll-10                      142           8309563 ns/op
BenchmarkSelectThenSelectAll-10                7         191876863 ns/op
BenchmarkSelectOnly200-10                   6621            170949 ns/op
BenchmarkSelectWithinSelect200-10              6         175394083 ns/op
BenchmarkLeftJoin200-10                      176           6941688 ns/op
BenchmarkSelectThenSelect200-10               34          37195855 ns/op
```

The data:
- `SelectOnly` are selects without the count
- `SelectWithinSelect` are `SELECT ID, Name, (SELECT COUNT()...) Count FROM ...` queries
- `LeftJoin` are the left join queries
- `SelectThenSelect` are the `SELECT * FROM Group` then `SELECT COUNT() FROM Installations` on each group from the first query (this is the DTO approach).
- The numbers mean `All` for all rows (~2000) and `200` for `LIMIT 200` (since that’s the default page size).
- Data is not real enough, since the logic for selecting a group has more to it, but it is close enough.

As per our conversation, it makes more sense to have this on the DTO instead of modifying underlying logic, since this relates to two different tables. Even with the performance benefits, this piece of logic is only used at one place at the moment (I've left a comment in the code).

As a fun note, [this line of code](https://github.com/mattermost/mattermost-cloud/blob/master/internal/store/installation.go#L409) drove me crazy for some time while developing the tests (I was specifying a custom ID on the tests). */me laughs hysterically*

gabrieljackson (approved):
This looks great! Very nice work on the testing. 👍 

Only one small non-blocking comment.

cmd/cloud/group.go
```diff
@@ -47,6 +47,7 @@ func init() {
 
 	registerPagingFlags(groupListCmd)
 	groupListCmd.Flags().Bool("table", false, "Whether to display the returned group list in a table or not")
+	groupListCmd.Flags().Bool("retrieve-installation-count", false, "Whether to retrieve the installation count for the groups")
```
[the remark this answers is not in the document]
fmartingr: You're right, makes much more sense knowing that other flag, actually. Let me change that real quick.

gabrieljackson (approved):
LGTM

stafot (approved):
LGTM

PR #749, opened 2022-10-24
[MM-46849] Use rotator when resizing a cluster

fmartingr:
#### Summary

Add [rotator](https://github.com/mattermost/rotator) support to the resize operation of `cluster` in the same way we use it for `upgrade`. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46849

#### Release Note

```release-note
- Add rotator support to resize operation
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -17,6 +17,17 @@ import (
 	"github.com/mattermost/mattermost-cloud/model"
 )
 
+// Defaults
+const (
+	useRotatorDefault              = true
+	maxScalingDefault              = 5
+	maxDrainRetriesDefault         = 10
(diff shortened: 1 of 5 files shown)


fmartingr:
I think this is ready to review now. I have tested this by resizing my test cluster and it seemed to work, though most kops operations are missing tests and I'm not sure how to approach that. Checking that the Rotator is being sent around is fairly easy, but checking that is currently working should be checked in the actual rotator test suite I think. I would like to know your opinion on this.

gabrieljackson (approved):
Nice cleanup and reuse of code!

stylianosrigas (approved):
LGTM @fmartingr 🚀

PR #752, opened 2022-10-26
Refactor PGBouncer config and add auth query option

gabrieljackson:
 - Refactors PGBouncer configuration into a single object and removes the original Get and Set shared variable logic.
 - Adds validation logic and unit tests.
 - Adds option to define a custom auth query.

Fixes https://mattermost.atlassian.net/browse/MM-47932

```release-note
Refactor PGBouncer config and add auth query option
```


fmartingr (approved):
LGTM

PR #753, opened 2022-10-26
feat : dependabot workflow automation for updating dependency

Rajpratik71:
#### Summary

Dependabot workflow automation for updating dependency

#### Release Note

```release-note
feat : dependabot workflow automation for updating dependency
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+
+  # Dependencies listed in .github/workflows/*.yml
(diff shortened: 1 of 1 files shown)


fmartingr:
Hey @Rajpratik71, thanks for the contribution!

Sadly we do dependency management more on a manual basis since some of those usually require manual attention or pinning, thus having the kind of activity _dependabot_ creates is not useful in terms of maintenance.

Thanks for taking the time to go through this anyway, really appreciated!

Rajpratik71:
Manual control of dependency is fine but with a growing no. of distributed upstream dependencies, it becomes hard to manage. So, for that automation should be there to update dependencies. Further, CI Pipeline is there to test those changes. The current build is having many old dependencies due to that many vulnerabilities were found. i.e. why workflow automation will help here.

Further, this will not update the dependencies automatically, instead, a PR will be opened with changes that can be reviewed and tested with CI.

PR #761, opened 2022-11-04
[MM-48112] Dependencies update

fmartingr:
#### Summary

Update project dependencies

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48112

#### Release Note

```release-note
Updated most dependencies and go version
```


gabrieljackson (commented):
@fmartingr this is great.

There are two additional files that we should review as part of this: `.circleci/config.yml` and the `Makefile`. Each of these also contains some versioning information for Go that we should update.

Makefile
```diff
@@ -5,16 +5,18 @@
 ##                             VERSION PARAMS                                 ##
 ################################################################################
 
-## Docker Build Versions
-DOCKER_BUILD_IMAGE = golang:1.17
-DOCKER_BASE_IMAGE = alpine:3.16
-
 ## Tool Versions
+GOLANG_VERSION := $(shell cat go.mod | grep "^go " | cut -d " " -f 2)
```
fmartingr, line 9: @gabrieljackson  Would like feedback on this. I'm like 1/5, since I'd rather be explicit but on the other hand I can't find why we would use a different version on the Makefile than in the Go file.

gabrieljackson: Hey, yeah I think I am 0/5 on this. If we didn't have to still explicitly update the circleci config as well then I could see this making more sense. I am totally fine with either implementation so I will leave it up to you.

fmartingr:
Moved to #762

PR #764, opened 2022-11-09
[MM-46231] Migrate to AWS SDK v2 [acm]

fmartingr:
#### Summary

Start the migration to AWS SDK v2.

This PR contains the refactoring of the ACM logic in our AWS client in order to use the V2 client, which includes:
 - Initialisation of the AWS configuration the way v2 requires (using configuration instead of sessions)
 - Refactoring the logic of the ACM client methods to use SDK v2
 - Refactoring our AWS Client to use and expose our own objects instead of AWS objects (only for ACM). Mainly created a `model.Certificate` with the only field that we seemed to use from `aws.CertificateSummary`.
 - Refactoring the mock for the ACM tests, since the SDK v2 doesn't expose services interfaces and we needed to create our own.

This PR also includes an unrelated change that I made during my testing but I think is worth keeping (secret manager), refactoring the database username/password logic and moved it to models to a public method, since that is not tied to AWS itself but is our own choice. It can be removed and moved to a separate PR however.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48294

#### Release Note

```release-note
Migrate ACM to AWS-SDK v2
```


internal/tools/aws/acm_types.go
```diff
@@ -0,0 +1,26 @@
...
+// ACMAPI represents the series of calls we require from the AWS SDK v2 ACM Client
+type ACMAPI interface {
+	ListCertificates(ctx context.Context, params *acm.ListCertificatesInput, optFns ...func(*acm.Options)) (*acm.ListCertificatesOutput, error)
+	ListTagsForCertificate(ctx context.Context, params *acm.ListTagsForCertificateInput, optFns ...func(*acm.Options)) (*acm.ListTagsForCertificateOutput, error)
+}
+
+// newCertificateFromACMCertificateSummary converts an ACM's certificate summary into our own certificate type
+func newCertificateFromACMCertificateSummary(c types.CertificateSummary) *model.Certificate {
+	return &model.Certificate{
+		ARN: c.CertificateArn,
```
mirshahriar, line 24: Is it possible that **ARN** could be nil?

fmartingr: It's possible that `c.CertificateArn` would be nil ([docs](https://github.com/aws/aws-sdk-go-v2/blob/main/service/acm/types/types.go#L18)). 

Right now we search ACM Certs by the CertificateARN ([here](https://github.com/mattermost/mattermost-cloud/blob/048a9462589518fc7ae3066b3ab513814ee498f6/internal/tools/aws/acm.go#L33)) so it is supposed to return a certificate with an ARN.

Logic can be improved here but after thinking about it, since this is a transformation from an AWS Type to our own Type, I think the best way forward is to maintain the nullability in our types since we may require nullables in the future, so the method usages need to validate that on their end. (Just added it in https://github.com/mattermost/mattermost-cloud/pull/764/commits/f78ea8158bee5e2dabf67b0abc78f665797531b1, thanks for noticing)

internal/supervisor/installation.go
```diff
@@ -1244,11 +1244,11 @@ func (s *InstallationSupervisor) finalDeletionCleanup(installation *model.Instal
 		if record.DeleteAt > 0 {
 			continue
 		}
-		err = s.dnsProvider.DeleteDNSRecords([]string{record.DomainName}, logger)
-		if err != nil {
-			logger.WithError(err).Error("Failed to delete DNS record from Cloudflare")
-			return model.InstallationStateDeletionFinalCleanup
-		}
+		// err = s.dnsProvider.DeleteDNSRecords([]string{record.DomainName}, logger)
```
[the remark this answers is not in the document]
fmartingr: This is a oversight on my part, had a lot of things stashed and this slipped in. I commented out that code on my first week because I had a corrupted state, since I had installations pending deletion trying to delete DNS Records but I still had no access. This allowed me to safely pass that post deletion step, but slipped in this PR. Thanks a lot for catching it!

internal/provisioner/nginx.go
```diff
@@ -104,11 +104,15 @@ func (n *nginx) Migrate() error {
...
 
 func (n *nginx) NewHelmDeployment() (*helmDeployment, error) {
-	awsACMCert, err := n.awsClient.GetCertificateSummaryByTag(aws.DefaultInstallCertificatesTagKey, aws.DefaultInstallCertificatesTagValue, n.logger)
+	certificate, err := n.awsClient.GetCertificateSummaryByTag(aws.DefaultInstallCertificatesTagKey, aws.DefaultInstallCertificatesTagValue, n.logger)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to retrive the AWS ACM")
 	}
 
+	if certificate.ARN == nil {
+		return nil, fmt.Errorf("retrieved certificate does not have ARN")
```
gabrieljackson, line 113:
```suggestion
		return nil, errors.New("retrieved certificate does not have ARN")
```

internal/provisioner/nginx_internal.go
```diff
@@ -105,16 +105,20 @@ func (n *nginxInternal) Migrate() error {
...
 func (n *nginxInternal) NewHelmDeployment() (*helmDeployment, error) {
 
-	awsACMPrivateCert, err := n.awsClient.GetCertificateSummaryByTag(aws.DefaultInstallPrivateCertificatesTagKey, aws.DefaultInstallPrivateCertificatesTagValue, n.logger)
+	certificate, err := n.awsClient.GetCertificateSummaryByTag(aws.DefaultInstallPrivateCertificatesTagKey, aws.DefaultInstallPrivateCertificatesTagValue, n.logger)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to retrive the AWS Private ACM")
 	}
 
+	if certificate.ARN == nil {
+		return nil, fmt.Errorf("retrieved certificate does not have ARN")
```
gabrieljackson, line 114:
```suggestion
		return nil, errors.New("retrieved certificate does not have ARN")
```

gabrieljackson (approved):
Looks good! Just a few non-blocking suggestions on my end.

PR #765, opened 2022-11-09
test: fixed non deterministic result check

fmartingr:
#### Summary

Fixed a test that was expecting deterministic results.

#### Ticket Link

NONE

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/tools/aws/tags_test.go
+++ b/internal/tools/aws/tags_test.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
 package aws
 
 import (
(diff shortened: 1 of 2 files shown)


gabrieljackson (approved):
👍

mirshahriar:
I think this equal check doesn't consider order.

fmartingr:
> I think this equal check doesn't consider order.

I made some tests and [it does](https://go.dev/play/p/VYB1I1qSXwq), that's why they have `assert.Subset` for this scenarios.

mirshahriar:
> 

You are right. Could you please also check the length. Subset checks may pass while size is not same. `assert.Subset(t, [1, 2, 3], [1, 2]) is true`

fmartingr:
> > 
> 
> You are right. Could you please also check the length. Subset checks may pass while size is not same. `assert.Subset(t, [1, 2, 3], [1, 2]) is true`

You're totally right. 👍

internal/tools/aws/tags_test.go
```diff
@@ -53,8 +57,9 @@ func TestTagsASRDSTags(t *testing.T) {
 	key2 := "key2"
 	value2 := "value2"
 	tags, err := NewTags(key, value, key2, value2)
+	result := tags.ToRDSTags()
 	assert.NoError(t, err)
```
mirshahriar, line 61: It is preferable to check error immediately after it occurs.

In this case, just after
```
tags, err := NewTags(key, value, key2, value2)
```

mirshahriar (approved):

PR #766, opened 2022-11-11
[MM-46265] Uniform Cluster VPC resource tagging

fmartingr:
#### Summary

Use the same tagging approach for cluster creation independently if a VPC is specified.

Add some basic testing.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46265

#### Release Note

```release-note
Fixed cluster VPC resource tagging when specifying a custom VPC
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -81,20 +81,15 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 
 	var clusterResources aws.ClusterResources
 	if kopsMetadata.ChangeRequest.VPC != "" && provisioner.params.UseExistingAWSResources {
-		clusterResources, err = awsClient.GetVpcResourcesByVpcID(kopsMetadata.ChangeRequest.VPC, logger)
+		clusterResources, err = awsClient.ClaimVPC(kopsMetadata.ChangeRequest.VPC, cluster, provisioner.params.Owner, logger)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -81,9 +81,9 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 
 	var clusterResources aws.ClusterResources
 	if kopsMetadata.ChangeRequest.VPC != "" && provisioner.params.UseExistingAWSResources {
-		clusterResources, err = awsClient.GetVpcResourcesByVpcID(kopsMetadata.ChangeRequest.VPC, logger)
+		clusterResources, err = awsClient.ClaimVPC(kopsMetadata.ChangeRequest.VPC, cluster, provisioner.params.Owner, logger)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "Couldn't claim VPC")
```
gabrieljackson, line 86: nit, lowercase letters for wrapped error returns
```suggestion
			return errors.Wrap(err, "couldn't claim VPC")
```

gabrieljackson (approved):
Nice test additions. Thanks for adding that.

I quickly reviewed the `ClaimVPC` logic and that looks like the right move here. The only change we may need to make is to check if it is the only cluster in that VPC or not as well. You can tackle that in https://mattermost.atlassian.net/browse/MM-46277 though so this LGTM.

mirshahriar (approved):

PR #769, opened 2022-11-16
added support to set etcd manager envs

mirshahriar:
#### Summary

Added support to set etcd manager environment variables

Added two optional flags in server

```go
serverCmd.PersistentFlags().Int("etcd-quota-backend-bytes", 4294967296, "Raise alarms by cluster when backend size exceeds the given quota")
serverCmd.PersistentFlags().String("etcd-listen-metrics-url", "http://0.0.0.0:8081", "An additional URL to listen for metrics")
```

We have used  `--set strings` flag of `kops create cluster` to add these environment variables.

> ###  Note: For the time being, we have no choice but to set these environment variables.

Sample `etcdClusters` from `kops` **Cluster**

```yaml
  etcdClusters:
  - cpuRequest: 200m
    etcdMembers:
    - encryptedVolume: true
      instanceGroup: master-us-east-1c
      name: c
    manager:
      env:
      - name: ETCD_QUOTA_BACKEND_BYTES
        value: "4294967296"
      - name: ETCD_LISTEN_METRICS_URLS
        value: http://0.0.0.0:8081
    memoryRequest: 100Mi
    name: main
```

> Note: We need to set these environment variables only in main cluster.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48271

```release-note
Added environment variables in etcd cluster.
```


stafot:
out of interest @mirshahriar , why you didn't follow an approach like this https://github.com/mattermost/mattermost-cloud/blob/dbf4fdd6736bbc316ecbebb96d74bd4cd05adfa4/internal/provisioner/kops_provisioner_cluster.go#L147-L151 to be alligned with what we done in the past?

mirshahriar:
> out of interest @mirshahriar , why you didn't follow an approach like this
> 
> https://github.com/mattermost/mattermost-cloud/blob/dbf4fdd6736bbc316ecbebb96d74bd4cd05adfa4/internal/provisioner/kops_provisioner_cluster.go#L147-L151
> 
> to be alligned with what we done in the past?

I didn't notice. Thank you for pointing out.

internal/provisioner/kops_provisioner.go
```diff
@@ -31,6 +31,7 @@ type ProvisioningParams struct {
 	DeployMinioOperator     bool
 	NdotsValue              string
 	PGBouncerConfig         *PGBouncerConfig
+	EtcdManagerEnv          map[string]interface{}
```
fmartingr, line 34: nit: I'm not a fan of `interfaces{}` where we know the value types (even if the flag is `int`, internally we only use those as strings). 

And I understand that this is a map instead of a struct to ease the iteration later on, but still seems weird to me. In this case I would go with a `EtcdManager` struct with the fields we allow via flags, and an helper method to convert the structure to a manageable `[]string` to use with `kops.SetCluster`. This is personal preference!

fmartingr (commented):
Hey @mirshahriar , good work! I've left some comments there regarding structure creation and command invocation. I'm pretty sure you've tested this yourself but you wanted to double check!

Also, I'm assuming that adding a test for this scenario is a pain in regard to kops, right? Since we're calling the binary underneath.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -189,6 +189,25 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 			return errors.Wrapf(err, "failed to set %s", setValue)
 		}
 	}
+
+	if len(provisioner.params.EtcdManagerEnv) > 0 {
+		var override []string
+		var overrideIndex int
+		for key, val := range provisioner.params.EtcdManagerEnv {
+			override = append(override, "spec.etcdClusters[*].manager.env=")
+			envName := fmt.Sprintf("spec.etcdClusters[*].manager.env[%d].name=%s", overrideIndex, key)
```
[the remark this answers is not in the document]
mirshahriar: It works this way.

mirshahriar:
> 

We are unable to add an e2e test for this because e2e works with rest calls to the provisioner server, which does not return these values. Also, we can't test it with the `kops` binary as well because it requires a storage location in the `—state` flag, which we don't have in our e2e suite.

internal/provisioner/kops_provisioner.go
mirshahriar: Right. As values are always `string` in **Env**, we can use `map[string]string` here. But I'm not sure about `struct`. If we use struct, we need pointer types, or else we can't make any flag optional.

```go
type EtcdManager struct {
	etcdQuotaBackendBytes *int
	etcdListenMetricsURL *string
}
```

fmartingr: I'm like 0/5 on the struct, just felt more familiar. Leaving this up to you, since you've dug more in that particular piece on the provisioner.

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -189,6 +189,25 @@ func (provisioner *KopsProvisioner) CreateCluster(cluster *model.Cluster, awsCli
 			return errors.Wrapf(err, "failed to set %s", setValue)
 		}
 	}
+
+	if len(provisioner.params.EtcdManagerEnv) > 0 {
+		var override []string
+		var overrideIndex int
+		for key, val := range provisioner.params.EtcdManagerEnv {
+			override = append(override, "spec.etcdClusters[*].manager.env=")
```
fmartingr, line 197: After thinking a bit about this, If this line is to create the `env` dictionary under `manager` it should be outside of the `for` loop. Maybe is working right now because declaring it with an empty value doesn't remove existing elements as you told me yesterday.

If this line is to create a new index under the `.manager.env`, you can disregard this comment.

fmartingr (approved):
LGTM! Hope we get some feedback from the Cloud SRE team regarding how to handle the `append` via Kops better.

gabrieljackson (approved):
Looks good!

stafot (approved):
LGTM

PR #770, opened 2022-11-21
[MM-48301/MM-48299] Upgrade EC2/VPC AWS SDK to v2

fmartingr:
#### Summary

Refactor EC2 Service usage to AWS SDK v2

⚠️ Changed the signature for several EC2 methods that used slice pointers for input/output, since the SDK modified that underneath as well (and it makes much more sense this way in my opinion, don't want to have a `[]*string{nil}`).

📝 [mocks/ec2.go](https://github.com/mattermost/mattermost-cloud/pull/770/files#diff-baf07ffcce9c27c14c0991ea0e820b24d043681b31e884c4b212f0e5707092b4) has that number of deletions because we're only mocking the methods we actually use from the SDK, rather than the entire interface. Check `ec2_types.go`.

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-48301
- https://mattermost.atlassian.net/browse/MM-48299

#### Release Note

```release-note
Refactor AWS SDK EC2 usage to v2
```


mirshahriar (commented):
added a few comments. 

+ I think instead of declaring a new `ctx` variable, we can directly use `context.TODO()`

internal/tools/aws/ec2_types.go
```diff
@@ -0,0 +1,26 @@
...
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+)
+
+// EC2API represents the series of calls we require from the AWS SDK v2 EC2 Client
+type EC2API interface {
```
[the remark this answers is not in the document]
fmartingr: Because the v1 SDK provided interfaces for all services but the v2 does not, so we need to create our own with the methods we use so we can properly mock those in tests using `gomock`.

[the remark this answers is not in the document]
mirshahriar: Understood

mirshahriar (approved):
LGTM

internal/tools/aws/route53.go
```diff
@@ -546,7 +546,7 @@ func prettyRoute53Response(resp *route53.ChangeResourceRecordSetsOutput) string
 
 // parseHostedZoneResourceID removes prefix from hosted zone ID.
 func parseHostedZoneResourceID(hostedZone *route53.HostedZone) (string, error) {
-	id := strings.TrimLeft(*hostedZone.Id, hostedZonePrefix)
+	id := strings.TrimLeft(*hostedZone.Id, hostedZonePrefix[0:len(hostedZonePrefix)-1])
```
gabrieljackson, line 549: Did you find a bug here or was this just a general improvement?

gabrieljackson (approved):
Very nice! Just one small question below.

internal/tools/aws/route53.go
fmartingr: Linter complained because `TrimLeft` uses the second argument as a character list to remove from the first argument, but `hostedZonePrefix` starts and ends with a slash.

gabrieljackson: Hmm, yeah I think either would work, but this seems fine to me as is.

gabrieljackson (approved):

PR #771, opened 2022-11-24
refinement of flag parsing (cloud server)

mirshahriar:
## Sample

```go
type InstallationGetFlags struct {
	server           string
	lockInstallation bool
}

func (flags *InstallationGetFlags) AddFlags(cmd *cobra.Command) {
	cmd.Flags().StringVar(&flags.server, "server", "http://localhost:8075", "")
	cmd.Flags().BoolVar(&flags.lockInstallation, "lock-installation", true, "")
}
```

#### Ticket Link

[Jira Ticket](https://mattermost.atlassian.net/browse/MM-48628)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/cmd/cloud/server_flag.go
+++ b/cmd/cloud/server_flag.go
@@ -27,7 +27,7 @@ type supervisorOption struct {
 }
 
 func (flags *supervisorOption) addFlags(command *cobra.Command) {
-	command.PersistentFlags().BoolVar(&flags.disableAllSupervisors, "disable-all-supervisors", false, "disable all supervisorOption (API-only functionality)")
+	command.PersistentFlags().BoolVar(&flags.disableAllSupervisors, "disable-all-supervisors", false, "disable all supervisors (API-only functionality)")
 
 	command.PersistentFlags().BoolVar(&flags.clusterSupervisor, "cluster-supervisor", true, "Whether this server will run a cluster supervisor or not.")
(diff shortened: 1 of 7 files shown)


fmartingr (approved):
Nice work! I like it way more this way, with everything on the same place :D

If I had to complain about something here is that we do not need many assignments that come from the older flags usage, for example we don't need `slowPoll := sf.slowPoll` anymore since we can just use `sf.slowPoll`, but this seems good to go :D

mirshahriar:
> Nice work! I like it way more this way, with everything on the same place :D
> 
> If I had to complain about something here is that we do not need many assignments that come from the older flags usage, for example we don't need `slowPoll := sf.slowPoll` anymore since we can just use `sf.slowPoll`, but this seems good to go :D

You are right. I will update them all. Thank you for highlighting.

cmd/cloud/server_flag.go
```diff
@@ -0,0 +1,213 @@
+package main
+
+import (
+	"time"
+
+	toolsAWS "github.com/mattermost/mattermost-cloud/internal/tools/aws"
+	"github.com/spf13/cobra"
+)
+
+type supervisorOption struct {
```
gabrieljackson, line 10: nit: `supervisorOptions` is probably a bit better along with the other options, but not a big deal. 0/5

gabrieljackson (approved):
I like this quite a bit! Nice work. I didn't go through every flag string to make sure that nothing was changed so I will trust that your copy-paste skills did the trick. 😄

PR #773, opened 2022-11-29
[MM-48554] Performance improvements on provisioner queries

fmartingr:
#### Summary

Performance improvements on provisioner queries:

- Added indexes in several columns:
  - `Installation.DeleteAt` since that's broadly used for filtering
  - `Installation.State` since that's also broadly used for filtering (will get better if we use an `ENUM` here)
  - `StateChangeEvent.EventID` since we use that for filtering in `JOIN`s
  - `StateChangeEvent(ResourceID, NewState)` since that's used together as a filter that was not triggering any indexes.
  - `Event.Timestamp` used for sorting events.

This is a first step in an always-ongoing performance improvements.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48554

#### Release Note

```release-note
Added indexes to several table columns to improve SQL performance
```

### DIFF
--- a/internal/store/migrations.go
+++ b/internal/store/migrations.go
@@ -1,5 +1,5 @@
-// See LICENSE.txt for license information.
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
 //
 
 package store
--- a/internal/store/migrations.go
+++ b/internal/store/migrations.go
@@ -1,5 +1,5 @@
-// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 //
 
 package store
@@ -910,7 +910,7 @@ var migrations = []migration{
 					ID TEXT PRIMARY KEY,
 					RawInstallationIDs BYTEA NOT NULL,
 					CreateAt BIGINT NOT NULL,
-					DeleteAt BIGINT NOT NULL,             
+					DeleteAt BIGINT NOT NULL,
 					LockAcquiredBy CHAR(26) NULL,
 					LockAcquiredAt BIGINT NOT NULL
 				);
@@ -1086,7 +1086,7 @@ var migrations = []migration{
 				DatabaseType TEXT NOT NULL,
 				InstallationsRaw BYTEA NOT NULL,
 				CreateAt BIGINT NOT NULL,
-				DeleteAt BIGINT NOT NULL,             
+				DeleteAt BIGINT NOT NULL,
 				LockAcquiredBy CHAR(26) NULL,
 				LockAcquiredAt BIGINT NOT NULL
 			);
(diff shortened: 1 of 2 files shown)


stafot:
@fmartingr While I agree with the indexes addition here I don't fully follow why `leftJoin` would be better here instead of `join` could you elaboarate a bit your reasoning/choice?

fmartingr:
> @fmartingr While I agree with the indexes addition here I don't fully follow why `leftJoin` would be better here instead of `join` could you elaborate a bit your reasoning/choice?

I've been digging my SQL test cases and I opted in for `left join` because there where no fields being used in the initial query for the joint table, but either I copied the query wrong (most likely) or I tested the wrong query (also likely). In my tests I remember that being a few ms quicker than using `INNER JOIN`. In this scenario, the only benefit would be the ability to retrieve `StatusChangeEvent` rows even if they wouldn't have an `Event`, but that can't be the case since `EventID` is not nullable.

Thanks for spotting this @stafot, I need to find out which query I was testing, hope the undo history reaches that far 😅

stafot (approved):
@fmartingr LGTM. 
I am pretty confident that this PR will help on CPU front. I have some worries for the impact on memory usage but let's see.

fmartingr:
> I am pretty confident that this PR will help on CPU front. I have some worries for the impact on memory usage but let's see.

Yeah, the indexes for the most time-consuming query is going to be as big as a primary key index, but previously that same query had zero index hits. We need to find balance once we see what happens with some real usage.

> I'm just curious about the size of the improvement. Have you tried the command explain? 

I used a combination of `explain` and `pg_stat_` tables. If you have some feedback on something I we can do to get more information out of queries just let me know.

> What has changed since the addition of the index? If possible, I would like you to include a few changes in the description.

What exactly did you want me to address in the description?

mirshahriar:
> What exactly did you want me to address in the description?

Like, before adding index It used to be a full table scan, but now it's an index scan. (just an example)

fmartingr:
> Like, before adding index It used to be a full table scan, but now it's an index scan. (just an example)

That's the case for all indexes, specially the joint index on StateChangeEvent which was triggering full table scans all the time and now it uses indexes, but all of them improved scans on my tests, and since most of them are used to sort/filter most of the time I think there would be improvements in other places as well even if those queries are not a problem nowadays. If you want me to get some actual numbers I can replicate some of the queries with a before/after.

internal/store/migrations.go
```diff
@@ -1,5 +1,5 @@
-// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
gabrieljackson, line 2: Hmm, guessing you didn't mean to move this?

gabrieljackson (approved):
Good conversation on this! These changes look reasonable to me. We can see how they impact things and go from there.

internal/store/migrations.go
fmartingr: That's weird. I have a plugin that takes care of that for me. 🤔  Let me revert that.

PR #776, opened 2022-12-01
[MM-48593] Create Prometheus SLOs per cluster

fmartingr:
#### Summary

Create the Sloth Prometheus Server Level object on cluster creation.

Removed a duplicated piece of code that wasn't being used: [kops_provisioner_installation_sli.go](https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/kops_provisioner_installation_sli.go) (logic is in [provisioner_installation_sli.go](https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/provisioner_installation_sli.go))

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48593

#### Release Note

```release-note
Create Prometheus SLOs on cluster creation
```

PR #777, opened 2022-12-02
[MM-48583] Create SLIs/SLOs per ring via provisioner

fmartingr:
#### Summary

Create Ring SLOs on cluster provision.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48583

#### Release Note

```release-note
Create Ring SLOs on cluster provision
```

Blocked by https://github.com/mattermost/mattermost-cloud/pull/776


model/installation_database.go
```diff
@@ -87,6 +87,7 @@ type InstallationDatabaseStoreInterface interface {
 	GetProxyDatabaseResourcesForInstallation(installationID string) (*DatabaseResourceGrouping, error)
 	GetOrCreateProxyDatabaseResourcesForInstallation(installationID, multitenantDatabaseID string) (*DatabaseResourceGrouping, error)
 	DeleteInstallationProxyDatabaseResources(multitenantDatabase *MultitenantDatabase, databaseSchema *DatabaseSchema) error
+	GetGroupDTOs(filter *GroupFilter) ([]*GroupDTO, error)
```
fmartingr, line 90: I'm not enjoying this interface's name. Would like suggestions. `ProvisionerStore` ?

mirshahriar: Are you going to add implementation of `GetGroupDTOs` later?

internal/provisioner/cluster_sli.go
```diff
@@ -12,6 +12,43 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const objective = 99.5
```
mirshahriar, line 15: Merely an opinion. Do we wish to retrieve this value from the ENV? Because, if we modify this value in the future, a new version will be required.

model/installation_database.go
fmartingr: Implementation already exists in [internal/store/group_dto.go](https://github.com/mattermost/mattermost-cloud/blob/master/internal/store/group_dto.go#L30)

internal/provisioner/cluster_sli.go
fmartingr: My guess is that if this value is modified, we will need to re-provision every cluster's CRDs and that will be made either: 1) a script by the SRE team 2) a reprovision of all clusters. And even if we use an environment variable, we'll need to restart/redeploy the provisioner cnc for it to take effect (unsure if we have something like reloader in place).

I don't see much value adding a new flag for this configuration at this point tbh, but I'd want to hear SRETeam opinion on the subject, since this request comes from them and will have more context on the matter.

internal/provisioner/cluster_sli.go
```diff
@@ -35,7 +83,7 @@ func makeClusterSLOs(cluster *model.Cluster) slothv1.PrometheusServiceLevel {
 					},
 					Description: "Availability metric for mattermost API on cluster " + cluster.ID,
 					Name:        "requests-availability-cluster",
-					Objective:   99.5,
+					Objective:   objective,
```
stafot, line 86: since you made this effort you could make it also a flag.

internal/provisioner/cluster_sli.go
stafot: Since you made this a variable take the full path and make it also a flag imho.

fmartingr: Two against one, huh? 😛

fmartingr: Pease check: https://github.com/mattermost/mattermost-cloud/pull/777/commits/28bdec31ad7e21fac134c7064ceeebbbf8c06c5e

internal/provisioner/cluster_sli.go
```diff
@@ -12,14 +12,62 @@ import (
...
+			},
+			Service: resourceName,
+			SLOs: []slothv1.SLO{
+				{
+					Alerting: slothv1.Alerting{
+						PageAlert:   slothv1.Alert{Disable: true},
+						TicketAlert: slothv1.Alert{Disable: true},
+					},
+					Description: "Availability metric for mattermost API on " + group.Name,
+					Name:        "requests-availability-cluster",
```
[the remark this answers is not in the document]
fmartingr: Hey @angeloskyratzakos, any comments here?

mirshahriar (approved):
LGTM

gabrieljackson (approved):
No blockers from me! 👍

internal/provisioner/cluster_sli.go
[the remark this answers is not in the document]
stafot: @fmartingr proceed with this naming.

PR #785, opened 2022-12-09
[MM-46451] Avoiding unnecessary CloudFlare requests

mirshahriar:
#### Summary

We will only update DNS records when changes necessitate it.

Steps:
- Get DNS records
- If there are no existing records, create a new one.
- If record already exists, only update if record is modified.

#### Ticket Link

Resolves [this ticket](https://mattermost.atlassian.net/browse/MM-46451)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/tools/cloudflare/dns.go
+++ b/internal/tools/cloudflare/dns.go
@@ -154,7 +154,7 @@ func (c *Client) upsertDNS(zoneNameList []string, dnsName, dnsEndpoint string, l
 		if *recordToUpdate.Proxied != *record.Proxied {
 			doUpdate = true
 		}
-	} else if recordToUpdate.Proxied != nil || record.Proxied != nil {
+	} else if recordToUpdate.Proxied != record.Proxied {
 		doUpdate = true
 	}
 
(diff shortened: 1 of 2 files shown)


internal/tools/cloudflare/dns.go
```diff
@@ -114,42 +114,64 @@ func (c *Client) CreateDNSRecords(dnsNames []string, dnsEndpoints []string, logg
...
+		doUpdate := false
+
+		if recordToUpdate.Content != record.Content || recordToUpdate.TTL != record.TTL {
+			doUpdate = true
+		}
+		if recordToUpdate.Proxied != nil && record.Proxied != nil {
+			if *recordToUpdate.Proxied != *record.Proxied {
+				doUpdate = true
+			}
+		} else if recordToUpdate.Proxied != nil || record.Proxied != nil {
```
fmartingr, line 148: Shouldn't this be: ?

```suggestion
		} else if recordToUpdate.Proxied != record.Proxied {
```

fmartingr (approved):
LGTM! 👍 I don't particularly like the anonymous function, but I like that all logic is still present on the same method with it.

mirshahriar:
>  I don't particularly like the anonymous function

We require a function to address memory leaks caused by `defer` calls.  This function has no other callers. That is why I made it anonymous. But you are right.

stafot (approved):

gabrieljackson (approved):
👍

PR #787, opened 2022-12-14
Set priorities to 'creation-requested' and 'creation-in-progress'

mirshahriar:
#### Summary

We are setting priorities on the following states while executing

- creation-requested
- creation-in-progress

#### Ticket Link

  Fixes [MM-33114](https://mattermost.atlassian.net/browse/MM-33114)

#### Release Note
```release-note
Set priorities to 'creation-requested' and 'creation-in-progress'
```

### DIFF
--- a/internal/supervisor/cluster.go
+++ b/internal/supervisor/cluster.go
@@ -75,15 +75,9 @@ func (s *ClusterSupervisor) Do() error {
 		return nil
 	}
 
-	// Preference of state (high to low).
-	preferredOrder := map[string]int{
-		model.ClusterStateCreationRequested:  2,
-		model.ClusterStateCreationInProgress: 1,
-	}
-
 	// Sort the clusters by state preference. Relative order is preserved.
 	sort.SliceStable(clusters, func(i, j int) bool {
-		return preferredOrder[clusters[i].State] > preferredOrder[clusters[j].State]
+		return model.PendingWorkPriority[clusters[i].State] > model.PendingWorkPriority[clusters[j].State]
 	})
 
 	for _, cluster := range clusters {
--- a/internal/supervisor/cluster.go
+++ b/internal/supervisor/cluster.go
@@ -5,6 +5,8 @@
 package supervisor
 
 import (
+	"sort"
+
 	"github.com/mattermost/mattermost-cloud/internal/tools/aws"
 	"github.com/mattermost/mattermost-cloud/model"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


internal/supervisor/cluster.go
```diff
@@ -73,6 +75,11 @@ func (s *ClusterSupervisor) Do() error {
 		return nil
 	}
 
+	// Sort the clusters by state preference. Relative order is preserved.
+	sort.SliceStable(clusters, func(i, j int) bool {
+		return model.PendingWorkPriority[clusters[i].State] > model.PendingWorkPriority[clusters[j].State]
```
fmartingr, line 80: Wouldn't this crash if we try to access an unset State key on the `PendingWorkPriority` map? If we try to `PendingWorkPriority[ClusterStateResizeRequested]`, shouldn't this crash? But I see you added tests and those are passing. I'm confused.

model/cluster_states.go
```diff
@@ -71,6 +71,13 @@ var AllClusterStatesPendingWork = []string{
 	ClusterStateDeletionRequested,
 }
 
+// PendingWorkPriority is a map of states to their priority. Default priority is 0.
+// States with higher priority will be processed first.
+var PendingWorkPriority = map[string]int{
```
fmartingr, line 76: nit: `ClusterStateWorkPriority` ? 0/5

internal/supervisor/cluster_test.go
```diff
@@ -140,6 +149,61 @@ func TestClusterSupervisorDo(t *testing.T) {
...
+			mockEventProducer,
+			"instanceID",
+			logger,
+		)
+		err := supervisor.Do()
+		require.NoError(t, err)
+
+		clusterListByWorkOrder := mockEventProducer.clusterListByEventOrder
+		require.Equal(t, len(preferredClusterOrder), len(clusterListByWorkOrder))
+		for i, clusterID := range clusterListByWorkOrder {
```
fmartingr, line 202: I think you could use `Equal` directly to compare that the two slices and it's order are the same.

fmartingr (commented):
Looks good! I just have a question about the `map` and what it stores below, to avoid crashes.

internal/supervisor/cluster.go
mirshahriar: No. It will return zero value. 

https://go.dev/play/p/koT9sxmFsgP

fmartingr: I may have been thinking on something else then. Good good! Thanks!

internal/supervisor/cluster_test.go
mirshahriar: Thanks. That will be nice

fmartingr (approved):

PR #790, opened 2022-12-14
Migrate from CircleCI to Github Actions

spirosoik:
#### Summary
We gradually moving away from CirleCI to Github Actions as we are consolidating the CIs we use. In parallel we are introducing `trivy` in the PR level and we block the PR about the number of critical vulnerabilities.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/CLD-4697

#### Release Note
```release-note
NONE
```


.github/workflows/cd.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@dc323e67f16fb5f7663d20ff7941f27f5809e9b6 # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: cd/build-docker
+        run: make build-image
+
+      - name: ci/build-docker-e2e
+        run: make build-image-e2e
```
phoinixgrr, line 27: Would propose in introducing a makefile into the root of the project, that encapsulates building all affected projects, into a single target.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: ci/scan-docker-security-e2e
+        uses: aquasecurity/trivy-action@9ab158e8597f3b310480b9a69402b419bc03dbd5 # v0.8.0
+        with:
+          image-ref: "mattermost/mattermost-cloud-e2e"
+          format: "table"
+          exit-code: "1"
+          ignore-unfixed: true
+          vuln-type: "os,library"
+          severity: "CRITICAL"
+          output: "trivy-results-critical-e2e.sarif"
```
phoinixgrr, line 49: Since we choose to encapsulate all functionality in Makefiles, should these be part of the makefile? 

`make tests`  ?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Upload trivy scan E2E
+        uses: github/codeql-action/upload-sarif@896079047b4bb059ba6f150a5d87d47dde99e6e5 # v2.11.6
+        with:
+          sarif_file: "trivy-results-critical-e2e.sarif"
+
+      - name: cd/docker-push
+        run: ./scripts/push-docker.sh
+        env:
+          TAG: ${GITHUB_REF_NAME}
```
phoinixgrr, line 64: Suggest moving the push functionality to a Makefile, and drop all related bash scripts laying around.

Example; 
https://github.com/mattermost/rtcd/blob/master/Makefile#L151-L163

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.19"
+          cache: true
```
phoinixgrr, line 27: Should we make this part of a Makefile target instead?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+        run: make check-style
+
+      - name: ci/check-modules
+        run: make check-modules
+
+      - name: ci/check-boilerplate
+        run: make check-modules
+
+      - name: ci/check-mocks
+        run: make verify-mocks
```
phoinixgrr, line 39: Would suggest in encapsulating this, in a single makefile target
And then ghactions can call something like `make lint` to perform all related tests. 

Thus we decapsulate the specifics of the project completely from the actual CI. To keep the pipeline clean and readable and easily to be imported to other projects.

scripts/push-docker.sh
```diff
@@ -3,18 +3,9 @@
...
-else
-  echo "Pushing release $CIRCLE_TAG..."
-  TAG="$CIRCLE_TAG"
-fi
-
-echo $DOCKER_PASSWORD | docker login --username $DOCKER_USERNAME --password-stdin
+echo $DOCKER_TOKEN | docker login --username $DOCKER_USERNAME --password-stdin
 
 docker tag mattermost/mattermost-cloud:test mattermost/mattermost-cloud:$TAG
 docker tag mattermost/mattermost-cloud-e2e:test mattermost/mattermost-cloud-e2e:$TAG
```
phoinixgrr, line 11: Would vote in favor of killing all bash scripts laying around the project. and encapsulate this functionality on a root Makefile.

.github/workflows/cd.yml
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

.github/workflows/cd.yml
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

.github/workflows/ci.yml
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

.github/workflows/ci.yml
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

scripts/push-docker.sh
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

.github/workflows/cd.yml
phoinixgrr: Resolving since it is out of the scope. We are only migrating to ghactions in this iteration. 
Makefile concerns to be addressed in next iteration.

internal/events/delivery_test.go
```diff
@@ -205,7 +205,7 @@ func failureEventHandler(t *testing.T, deliveryChan chan<- *model.StateChangeEve
 }
 
 func awaitEvents(t *testing.T, deliveryChan <-chan *model.StateChangeEventPayload, expectedID string, eventsCount int) {
-	timeout := time.After(25 * time.Second)
+	timeout := time.After(50 * time.Second)
```
phoinixgrr, line 208: Is this change related to the scope of the PR?

phoinixgrr: Ok, I now have context https://community.mattermost.com/private-core/pl/8gycrgzoij8hdm6ogiines7zxe

fmartingr:
I have pushed two commits:
- One upgrades helm: It seems that there's no problem upgrading to the latest release, I've checked the changes and the code and it seems that everything we're currently using is still present in helm. If there's any obscure thing I should be aware of, let me know.
- The other one upgrades `go-restful` **transitive** dependency to trivy's suggested release. I would like @mirshahriar and @gabrieljackson 's 👀 on this as well, since I don't like upgrading transitive dependencies even if it seems that nothing has broken. I've made some tests and everything seems to work and apparently the API hasn't changed much and provisioner compiles, runs and provisions without noticeable issues (at least on my tests), but... this is sort of a dirty workaround. I do not want to ask for an exception regarding critical dependencies at the same time we're actually fixing them, but since this is not a publicly exposed service I'm not sure. Would like extra feedback on this.

mirshahriar (approved):
LGTM

fmartingr (approved):

PR #792, opened 2022-12-14
fix logger using pointer

fmartingr:
#### Summary

Dereferencing VpcID used in logger

#### Release Note

```release-note
NONE
```


gabrieljackson (approved):
Thanks!

PR #795, opened 2022-12-19
[MM-49155] Set priorities for creation-specific actions.

mirshahriar:
#### Summary

We are setting priorities on the following states while executing

**ClusterInstallation**
- creation-requested
- reconciling
- ready

**Installation**
- creation-requested
- creation-no-compatible-clusters
- creation-pre-provisioning
- creation-in-progress
- creation-configuring-dns

#### Ticket Link

  Fixes [MM-49155](https://mattermost.atlassian.net/browse/MM-49155)

#### Release Note
```release-note
Set priorities for creation-specific actions for Installation & ClusterInstallation.
```

### DIFF
--- a/internal/supervisor/cluster_installation_test.go
+++ b/internal/supervisor/cluster_installation_test.go
@@ -5,15 +5,14 @@
 package supervisor_test
 
 import (
+	"fmt"
 	"testing"
 
-	"github.com/mattermost/mattermost-cloud/internal/testutil"
-
 	"github.com/mattermost/mattermost-cloud/internal/provisioner"
-
 	"github.com/mattermost/mattermost-cloud/internal/store"
 	"github.com/mattermost/mattermost-cloud/internal/supervisor"
 	"github.com/mattermost/mattermost-cloud/internal/testlib"
+	"github.com/mattermost/mattermost-cloud/internal/testutil"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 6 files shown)


internal/supervisor/cluster_installation_test.go
```diff
@@ -23,23 +22,51 @@ type mockClusterInstallationStore struct {
...
+	Installations                           []*model.Installation
 	ClusterInstallations                    []*model.ClusterInstallation
 
 	UnlockChan                     chan interface{}
 	UpdateClusterInstallationCalls int
 }
 
 func (s *mockClusterInstallationStore) GetCluster(clusterID string) (*model.Cluster, error) {
-	return s.Cluster, nil
+	if s.Cluster != nil {
```
fmartingr, line 34: I know that this is the mock, but isn't necessary here to check that `s.Cluster.ID == clusterID`? Just wondering.

fmartingr (approved):
LGTM

internal/supervisor/cluster_installation_test.go
mirshahriar: Previously, this method returned the cluster regardless of `clusterID`.

```go
func (s *mockClusterInstallationStore) GetCluster(clusterID string) (*model.Cluster, error) {
	return s.Cluster, nil
}
```

`S.Cluster` is set if the test contains a single cluster. So `clusterID` does not require verification. 

However, for this new test, the list of clusters is set in `s.Clusters` ignoring `s.Cluster`

fmartingr: Gotcha 👍

gabrieljackson (commented):
This is a nice improvement, but the original ticket was actually for `Installation` resources. Do you think you could add those in here as well?

mirshahriar:
> This is a nice improvement, but the original ticket was actually for `Installation` resources. Do you think you could add those in here as well?

Thank you. I have added installation-specific changes here.

fmartingr (approved):

gabrieljackson (approved):
LGTM. Thanks for the update!

PR #796, opened 2022-12-20
[CLD-4714] Delete helm chart to clean up LB resources

mirshahriar:
#### Summary

We are manually deleting the helm chart for Nginx. This will clean up LB resources from the cloud provider.

#### Ticket Link

  Fixes [CLD-4714](https://mattermost.atlassian.net/browse/CLD-4714)

#### Release Note

```release-note
NONE
```

### DIFF
--- a/internal/provisioner/nginx_internal.go
+++ b/internal/provisioner/nginx_internal.go
@@ -18,6 +18,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	namespaceNginxInternal = "nginx-internal"
+)
+
 type nginxInternal struct {
 	awsClient      aws.AWS
 	kubeconfigPath string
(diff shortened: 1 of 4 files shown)


internal/provisioner/nginx_internal.go
```diff
@@ -89,6 +93,25 @@ func (n *nginxInternal) ActualVersion() *model.HelmUtilityVersion {
 }
 
 func (n *nginxInternal) Destroy() error {
+	k8sClient, err := k8s.NewFromFile(n.kubeconfigPath, n.logger)
+	if err != nil {
+		return errors.Wrap(err, "failed to set up the k8s client")
+	}
+
+	services, err := k8sClient.Clientset.CoreV1().Services("nginx-internal").List(context.TODO(), metav1.ListOptions{})
```
fmartingr, line 101: I think `"nginx-internal"` is stored somewhere as a `const`. Would be nice to use that on every instance now that you're on it.

fmartingr (approved):
👍

internal/provisioner/nginx_internal.go
mirshahriar: Right. We do not have any `const` now. But I will set one.

mirshahriar:
> Hey, I am curious if you tried deleting the helm chart for nginx instead of removing the LBs manually. I figured it may perform LB cleanup for us.

Indeed It's a good idea. I have changed to delete with helm. Thank you.

gabrieljackson (approved):
👍

PR #799, opened 2022-12-20
[MM-48302] Upgrade DynamoDB to SDK v2

fmartingr:
#### Summary

Upgrade DynamoDB to SDK v2

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48302

#### Release Note

```release-note
Upgraded AWS SDK for DynamoDB to V2
```


gabrieljackson (approved):
LGTM!

internal/tools/aws/dynamodb.go
```diff
@@ -5,29 +5,37 @@
...
-		if aerr.Code() == dynamodb.ErrCodeResourceNotFoundException {
+	_, err := a.Service().dynamodb.DescribeTable(
+		context.TODO(),
+		&dynamodb.DescribeTableInput{
+			TableName: aws.String(tableName),
+		})
+
+	if err != nil {
+		var awsErr *types.ResourceNotFoundException
+		if errors.As(err, &awsErr) {
```
mirshahriar, line 28: Just curious. What if It is a different type of error?

mirshahriar (approved):

internal/tools/aws/dynamodb.go
fmartingr: I'm assuming (from the previous logic) that if any error raises here, we will try to delete the database anyway (since that's expected), and any possible error would raise there instead. We could however have some kind of warning here (without ending the function) so is better traceable. What do you think?

mirshahriar: A warning message will be helpful.

PR #806, opened 2022-12-22
[MM-49345] AWS SDK v2 migration - RDS, KMS, GT

fmartingr:
#### Summary

Upgraded RDS, KMS and GT to AWS SDK v2 version

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49345

#### Release Note

```release-note
Upgraded AWS SDK for RDS, KMS & GT to v2
```


mirshahriar (approved):
LGTM.

gabrieljackson (commented):
Looks like we need to resolve some conflicts. Please re-request my review when that is done.

fmartingr:
> Looks like we need to resolve some conflicts. Please re-request my review when that is done.

My bad, I'll wait on the merge until the two pending services are reviewed.

gabrieljackson (approved):
As always, great work on this. I didn't catch any issues, but wanted to check to see if you ran any manual tests with this code such as creating an installation with a single tenant RDS cluster? The reason I ask is this is the most delicate part of the AWS code and just want to make sure we did some spot checks first.

fmartingr:
> As always, great work on this. I didn't catch any issues, but wanted to check to see if you ran any manual tests with this code such as creating an installation with a single tenant RDS cluster? The reason I ask is this is the most delicate part of the AWS code and just want to make sure we did some spot checks first.

I wanted to do some end testing using #810 that contains all AWS migration branches. More so looking at the problems that some of the refactors caused. Is there some particular configurations that I should be aware of while testing? Since those take its time to run, so I could try automate them somehow, and then run them on big changes like this.

gabrieljackson:
Yeah, great question. The most straightforward manual tests I run when I want to double-check things is to create and delete installations with different database and filestore options. I normally don't test them all, but will hit a few just to verify nothing obvious is wrong. Keep in mind that some database options may require the databases to be created first by database factory to be successful. We don't run all the database types in the dev account so you will need to check your cluster VPC if you are unsure if you can test the multitenant options.

PR #807, opened 2022-12-23
Revert command functions to their earlier names

mirshahriar:
#### Summary

Example

`newCmdCluster()` -> `clusterCmd()`

#### Release Note

```release-note
None
```

### DIFF
--- a/cmd/cloud/backup.go
+++ b/cmd/cloud/backup.go
@@ -10,21 +10,21 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func newCmdInstallationBackup() *cobra.Command {
+func installationBackupCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "backup",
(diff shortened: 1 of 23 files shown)


fmartingr (approved):
PR itself it's ok, though I like the `newXXX` functions more, since they actually creating objects. The old naming was like that because those were the actual cobra commands, not a factory of them. But I'll leave it up what you think is best. 1/5.

mirshahriar:
> The old naming was like that because those were the actual cobra commands

I completely agree.  These functions are creating new cobra command, so good to have `newCmd` prefix.  But, It is up to @gabrieljackson.

gabrieljackson (approved):
I see where you guys are coming from on this and am 0/5 now on the names. I will approve and you can decide whether to merge or close the PR. Totally fine with either.

mirshahriar:
Lets keep command name to `newXXX`

PR #815, opened 2022-12-27
[MM-49373] Bucket and Helm deletion fixes

fmartingr:
#### Summary

- Use `HeadBucket` instead of `GetBucketPolicy` to check if a bucket exists or not.
- Check chart is present before trying to delete it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49373

https://mattermost.atlassian.net/browse/MM-49377

#### Release Note

```release-note
None
```

### DIFF
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -68,12 +68,36 @@ func (d *helmDeployment) Update() error {
 func (d *helmDeployment) Delete() error {
 	logger := d.logger.WithField("helm-delete", d.chartDeploymentName)
 
-	logger.Infof("Deleting helm chart %s", d.chartDeploymentName)
-	err := deleteHelmChart(*d, d.kubeconfigPath, logger)
+	// Ensure the chart is present before deletion
+	exists, err := d.Exists()
 	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("got an error trying to delete the helm chart %s", d.chartDeploymentName))
+		return err
+	} else if !exists {
+		logger.Warnf("chart %s not present, assuming already deleted", d.chartDeploymentName)
+		return nil
 	}
+
+	err = deleteHelmChart(*d, d.kubeconfigPath, logger)
+	if err != nil {
+		return errors.Wrapf(err, "got an error trying to delete the helm chart %s", d.chartDeploymentName)
+	}
+
 	return nil
+}
+
+func (d *helmDeployment) Exists() (bool, error) {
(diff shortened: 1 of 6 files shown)


stafot (approved):

internal/provisioner/nginx_internal.go
```diff
@@ -95,15 +95,10 @@ func (n *nginxInternal) ActualVersion() *model.HelmUtilityVersion {
...
-	helm := newHelmDeployment(
-		chartNameNginxInternal,
-		chartDeploymentNameNginxInternal,
-		namespaceNginxInternal,
-		n.kubeconfigPath,
-		nil,
-		defaultHelmDeploymentSetArgument,
-		n.logger,
-	)
+	helm, err := n.NewHelmDeployment()
```
mirshahriar, line 98: Why are we adding this? `NewHelmDeployment` does some extra tasks. I think we do not need that to delete.

mirshahriar (approved):

internal/provisioner/helm_utils.go
```diff
@@ -68,14 +68,38 @@ func (d *helmDeployment) Update() error {
...
 	logger := d.logger.WithField("helm-delete", d.chartDeploymentName)
 
-	logger.Infof("Deleting helm chart %s", d.chartDeploymentName)
-	err := deleteHelmChart(*d, d.kubeconfigPath, logger)
+	// Ensure the chart is present before deletion
+	exists, err := d.Exists()
 	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("got an error trying to delete the helm chart %s", d.chartDeploymentName))
+		return err
+	} else if !exists {
```
mirshahriar, line 75:
```suggestion
	}
        if !exists {
```

mirshahriar (approved):

internal/provisioner/helm_utils.go
mirshahriar: nit: we do not need `else` here

fmartingr: I don't know what I was thinking 😂

internal/provisioner/nginx_internal.go
fmartingr: I wanted to be sure that we use the same helm deployment in both creation and deletion, this was the only provisioner that used a different one in the two methods.

fmartingr: @mirshahriar Take a look at the changes

PR #817, opened 2022-12-27
[MM-29718] Allow env var to set every cloud flag

mirshahriar:
#### Summary

Environment variables with the prefix `CP_` can be used to set any flag.

**Some examples**

`size-node-min-count` → `CP_SIZE_NODE_MIN_COUNT`
`debug` → `CP_DEBUG`
`slo-target-availability` → `CP_SLO_TARGET_AVAILABILITY`

The following order will be used

- flag value
- env value
- default value

#### Ticket Link

  Jira [Ticket](https://mattermost.atlassian.net/browse/MM-29718)

#### Release Note

```release-note
Allow ENV variables to set every cloud flag
```

### DIFF
--- a/cmd/cloud/main.go
+++ b/cmd/cloud/main.go
@@ -60,7 +60,7 @@ func main() {
 func populateEnv(cmd *cobra.Command) {
 	v := viper.New()
 
-	v.SetEnvPrefix("mm")
+	v.SetEnvPrefix("cp")
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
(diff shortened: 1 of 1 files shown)


gabrieljackson (commented):
Looks good, but let's change the prefix. `MM` is already used in mattermost server config so let's go with `CP` for Cloud Provisioner to avoid confusion.

fmartingr (approved):
Only a few lines of code and now we can use environment variables to set configuration options? What is this sorcery?

PR #818, opened 2022-12-28
Use value-type data to fix an erroneous pointer reference. 

mirshahriar:
#### Summary

```go
for _, zone := range zoneList.HostedZones {
    zones = append(zones, &zone)
}
```

Here, adding _pointer of zone_ causes invalid reference. Because this variable is being updated with each list items.

#### Ticket Link

  Fixes [MM-49380](https://mattermost.atlassian.net/browse/MM-49380)

#### Release Note

```release-note
Fixed an invalid pointer reference that led to an incorrect zone name.
```

### DIFF
--- a/internal/tools/aws/route53.go
+++ b/internal/tools/aws/route53.go
@@ -488,7 +488,7 @@ func (a *Client) getHostedZonesWithTag(tag Tag, firstOnly bool) ([]*types.Hosted
 			return nil, errors.Wrapf(err, "failed to list all hosted zones")
 		}
 
-		for _, zone := range zoneList.HostedZones {
+		for i, zone := range zoneList.HostedZones {
 			id, err := parseHostedZoneResourceID(zone)
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to parse hosted zone ID: %s", *zone.Name)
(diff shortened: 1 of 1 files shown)


stafot (approved):

stafot:
@mirshahriar Please add a release note

agnivade:
These bugs are caught by the [exportloopref](https://github.com/kyoh86/exportloopref) linter that we use in .golangci.yml in our other repos. I'd suggest to consider adding golangci in the CI pipeline.

fmartingr:
> These bugs are caught by the [exportloopref](https://github.com/kyoh86/exportloopref) linter that we use in .golangci.yml in our other repos. I'd suggest to consider adding golangci in the CI pipeline.

I suggested moving ourselves to golangci-lint as well, since we have some outdated linters around. Is there a consensus across mattermost or a template we could use? (so we keep all projects equally in terms of formatting and lint requirements)

agnivade:
Good point on the consensus, and probably something we should consider with our "developer experience" theme for next year. For now, there is no "official" statement. Although you will see 99% of repos under mattermost using golangci.

I don't think all projects need to be 100% equal in terms of formatting or lint requirements. Different projects can have different requirements with security, performance etc. And then there's also the matter of keeping the linter version up to date.

But if you are looking for a base template, you can use: https://github.com/mattermost/perseus/blob/main/.golangci.yml

PR #825, opened 2023-01-06
[MM-49445] Added stacktrace in Error log

mirshahriar:
#### Summary

To add stacktrace to logs, we must first set the flag.
```
$ cloud server [flags]

Flags:
    --enable-log-stacktrace             Add stacktrace in error logs.
```

This will add the stacktrace if the error type is `error`, `fatal`, or `panic`.

Sample:

```
ERRO[2023-01-06T07:55:57+06:00] Failed to lock installation, status: 404      handler=handleUpdateInstallation installation=f7qjd88xpibxfpf9w63156aeuh instance=cpc81je3sp8txbmit8sse7b94a method=PUT path=/api/installation/f7qjd88xpibxfpf9w63156aeuh/mattermost request=qx4p9ppsk78rzds3ixy1bhgpje stacktrace="internal/api/installation.go:816; internal/api/installation.go:348; internal/api/handler.go:38"
```

Added stacktrace
`stacktrace="internal/api/installation.go:816; internal/api/installation.go:348; internal/api/handler.go:38"`

#### Ticket Link

Resolves [MM-49445](https://mattermost.atlassian.net/browse/MM-49445)

#### Release Note

```release-note
None
```

### DIFF
--- a/cmd/cloud/logger.go
+++ b/cmd/cloud/logger.go
@@ -5,7 +5,10 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"runtime"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


fmartingr:
Can’t check properly at the moment, but we already have the stacktrace on the err (since we use github.com/pkg/errors), though I’m not sure we have that available in the log entry, but it may simplify things maybe? [ref](https://pkg.go.dev/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper)

mirshahriar:
> we already have the stacktrace on the err  (since we use github.com/pkg/errors)

You are right. We can use a **stacktrace** that is already available. However, stacktrace is not always available. For example
```
c.Logger.Errorf("Failed to lock installation, status: %d", status)
```

We need something like `logrus.WithError(errors.New("Foo"))` to get stacktrace.  There can be another way.

Even if we are able to get stacktrace, it will flood with lines in logs. Like below
```
github.com/mattermost/mattermost-cloud/internal/api.getInstallationForTransition
	/Users/mir/go/src/github.com/mattermost/mattermost-cloud/internal/api/installation.go:817
github.com/mattermost/mattermost-cloud/internal/api.handleUpdateInstallation
		/Users/mir/go/src/github.com/mattermost/mattermost-cloud/internal/api/installation.go:348
github.com/mattermost/mattermost-cloud/internal/api.contextHandler.ServeHTTP
		/Users/mir/go/src/github.com/mattermost/mattermost-cloud/internal/api/handler.go:38
github.com/gorilla/mux.(*Router).ServeHTTP
		/Users/mir/go/src/github.com/mattermost/mattermost-cloud/vendor/github.com/gorilla/mux/mux.go:210
net/http.serverHandler.ServeHTTP
	/usr/local/go/src/net/http/server.go:2947
net/http.(*conn).serve
	/usr/local/go/src/net/http/server.go:1991
runtime.goexit\n\t/usr/local/go/src/runtime/asm_arm64.s:1172
```

That's why I came up with a simple way that will only print like this.

`internal/api/installation.go:816; internal/api/installation.go:348; internal/api/handler.go:38"`

mirshahriar:
> 
> That's why I came up with a simple way that will only print like this.
> 

I am nonetheless open to discussion. @fmartingr , what are your thoughts?

fmartingr:
For me its ok like this, I only remembered that we had the ‘errors’ package, but its true that this are log lines and we may not have the actual error (though we should, if available). I think we can move forward with this and make any improvements in the future if necessary.

gabrieljackson (approved):
Oh wow, this looks to produce a lot cleaner logs than the typical full full stacktrace dumps that I am used to seeing. Very nice. As Felipe said, we can iterate on this if we find that something is missing.

fmartingr (approved):

PR #826, opened 2023-01-09
[MM-49396] Add Name tag in ELB resources

mirshahriar:
#### Summary

Adding Name tag in ELB resources

#### Ticket Link

  Resolves [MM-49396](https://mattermost.atlassian.net/browse/MM-49396)

#### Release Note

```release-note
None
```


internal/tools/aws/elb_types.go
```diff
@@ -0,0 +1,34 @@
...
+	GetLoadBalancerAPI(string) ELB
+}
+
+// ELB is an interface to access AWS resources
+type ELB interface {
+	GetLoadBalancerResource(name string) (string, error)
+	TagLoadBalancer(arn string, tags map[string]string) error
+}
+
+// ELBV1 represents the series of calls we require from the AWS SDK v1 ELB Client
```
fmartingr, line 25:
```suggestion
// ELBV1 represents the series of calls we require from the AWS SDK v2 ELB Client v1
```

internal/tools/aws/elb_types.go
```diff
@@ -0,0 +1,34 @@
...
+type ELBV1 interface {
+	AddTags(ctx context.Context, params *elbv1.AddTagsInput, optFns ...func(*elbv1.Options)) (*elbv1.AddTagsOutput, error)
+}
+
+// ELBV2 represents the series of calls we require from the AWS SDK v2 ELB Client
```
fmartingr, line 30:
```suggestion
// ELBV2 represents the series of calls we require from the AWS SDK v2 ELB Client v2
```

fmartingr (approved):
LGTM. Only minor naming issues. Seems an elegant way to solve the problem of AWS providing two different packages so we can still operate over classic load balancers. But that poses a question. Should we migrate classic load balancers? If AWS is stopping providing support this could be a worse problem in the future if the APIs get deprecated. IIRC for now is safe to have CLBs on a VPC, but who knows what could happen. cc: @gabrieljackson @stylianosrigas

internal/tools/aws/elb.go
```diff
@@ -0,0 +1,99 @@
...
+			elbv1.NewFromConfig(cfg),
+		},
+		elasticLoadbalancerV2: elasticLoadbalancerV2{
+			elbv2.NewFromConfig(cfg),
+		},
+	}
+}
+
+// GetLoadBalancerAPI returns the correct ELB API based on elb type
+func (c *Client) GetLoadBalancerAPI(elbType string) ELB {
```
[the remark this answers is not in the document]
mirshahriar: It can be `GetLoadBalancerAPIByType`?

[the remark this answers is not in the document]
fmartingr: Sure, that seems fine as well.

stafot (approved):

internal/tools/aws/elb.go
```diff
@@ -0,0 +1,99 @@
...
+	return name, nil
+}
+
+// TagLoadBalancer adds tags to the ELB
+func (e elasticLoadbalancerV1) TagLoadBalancer(name string, tags map[string]string) error {
+	var elbTags []elbv1Type.Tag
+	for key, value := range tags {
+		elbTags = append(elbTags, elbv1Type.Tag{
+			Key:   pointer.String(key),
+			Value: pointer.String(value),
```
gabrieljackson, line 53: nit: you can use `aws.String()` for stuff like this to avoid importing another package. Not a big deal though.

gabrieljackson (approved):
Looks great. Just need to resolve a few more merge conflicts.

PR #827, opened 2023-01-10
Add Perseus database type

gabrieljackson:
This change includes the basic functionality needed to deploy the Perseus database proxy service and to use it for new installations.

High Level Perseus Architecture:
Screenshot 2023-01-09 at 10 53 28 PM

Reviewer Notes:
 - Although this PR refactors some of the shared code with PGBouncer, there is a lot more we can do. Out of a desire to not break anything some code was duplicated for now with some minor tweaks which can be improved later.
 - Some PGBouncer provisioning happens as part of Perseus database setup. This is intended to allow for migrations to and from PGBouncer in the future. It's easier to do this stuff proactively instead of needing to do extensive updates later.
 - Sorry, I tried to keep this change smaller, but this is all needed for basic functionality.

```release-note
Add Perseus database type
```


internal/tools/aws/database_perseus.go
```diff
@@ -0,0 +1,1251 @@
...
+		return errors.Wrapf(err, "failed to ensure the pgbouncer auth user secret was created for %s", rdsID)
+	}
+
+	close, err := d.connectToRDSCluster(rdsPostgresDefaultSchema, *rdsCluster.Endpoint, DefaultMattermostDatabaseUsername, *masterDBSecret.SecretString)
+	if err != nil {
+		return errors.Wrapf(err, "failed to connect to the multitenant proxy cluster %s", rdsID)
+	}
+	defer close(logger)
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(DefaultMySQLContextTimeSeconds*time.Second))
```
fmartingr, line 473: `DefaultMySQLContextTimeSeconds` ?

internal/tools/aws/database_perseus.go
```diff
@@ -0,0 +1,1251 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create Mattermost database user")
+	}
+
+	err = d.ensurePerseusUserHasRole(ctx, installationSecret.MasterUsername)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to grant permissions to perseus database user")
+	}
+
+	err = d.ensureMasterUserHasRole(ctx, DefaultPerseusDatabaseUsername)
```
fmartingr, line 632: Argument is role, but we're sending username. Assuming same usrname and role names, but its confusing from this point of view.

I also find confusing how we create "DatabaseUser", "PerseusUser" and "MasterUser". Are all those three different? I'm asking because we are using the same values/db for some of them.

fmartingr (approved):
Following logic was a bit tricky at first but this looks good to me, my only worries are with naming (commented out below) on parts that I did not fully understand or found confusing. Some tests for the critical methods would also be nice.

> Although this PR refactors some of the shared code with PGBouncer, there is a lot more we can do. Out of a desire to not break anything some code was duplicated for now with some minor tweaks which can be improved later.

Can you put comments on the ones you already duplicated so we can easily track those?

internal/tools/aws/database_perseus.go
```diff
@@ -0,0 +1,1251 @@
...
+		return errors.Wrapf(err, "failed to get KMS key for perseus")
+	}
+	encBytes, err := d.client.kmsEncrypt(*key.Arn, perseusSecret.MasterPassword)
+	if err != nil {
+		return errors.Wrap(err, "failed to encrypt destination password")
+	}
+	encryptedDestinationPassword := base64.StdEncoding.EncodeToString(encBytes)
+
+	// Create Perseus auth entries.
+	close, err := d.connectToPerseusAuthDatabase(dbResources.MultitenantDatabase.VpcID, logger)
```
mirshahriar, line 198: Can we rename `close`? It collides with builtin `close` func.

internal/tools/aws/database_perseus.go
gabrieljackson: Yeah, it's more of a general timeout that we use, but let's create a new one to clear up naming confusion. 👍

internal/tools/aws/database_perseus.go
gabrieljackson: Yeah, I know this part is hard to follow. There are indeed 4 users: master DB user, perseus user to connect to the auth database, perseus user to connect to the shared DBs for connection pooling, and the individual schemas+users for each Mattermost installation. I have plans to clean this up a bunch for both perseus and pgbouncer in a follow-up PR.

internal/tools/aws/database_perseus.go
```diff
@@ -0,0 +1,1251 @@
...
+	if len(multitenantDatabases) == 0 {
+		return nil, nil, nil
+	}
+
+	unlockFn, err := lockMultitenantDatabase(multitenantDatabases[0].ID, d.instanceID, store, logger)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to lock multitenant database")
+	}
+	databaseResources, err := store.GetOrCreateProxyDatabaseResourcesForInstallation(d.installationID, multitenantDatabases[0].ID)
+	if err != nil {
```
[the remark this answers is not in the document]
gabrieljackson: Very nice find! This was actually a bug with PGBouncer as well so I will correct it in both places.

mirshahriar (approved):

PR #829, opened 2023-01-10
Update github ref name with context instead of env

fmartingr:
#### Summary

Fixes an incorrect environment reference on an action attribute, replacing it for a [context access](https://docs.github.com/en/actions/learn-github-actions/contexts).

From: https://github.com/mattermost/mattermost-cloud/actions/runs/3885573452/jobs/6629615255

#### Ticket Link

--

#### Release Note

```release-note
None
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,3 +121,11 @@ jobs:
         env:
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+
+  cd:
+    needs:
+      - lint
+      - test
+      - build
+    uses: ./.github/workflows/cd.yml
+    if: startsWith(github.ref, 'refs/tags/v')
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,12 +1,6 @@
 name: cd
-on:
-  workflow_run:
-    workflows: ["ci"]
-    types:
-      - completed
-  push:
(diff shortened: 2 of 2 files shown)


gabrieljackson (approved):
👍

spirosoik (approved):

fmartingr:
@spirosoik I think this PR is fine, but does not solve the problem we're facing in PRs with CD, I can take care of that as well.

fmartingr:
Let me know if [this](https://github.com/mattermost/mattermost-cloud/pull/829/commits/d07cc0c22c444ee56deb5bca9b4fb97988a5251b) makes sense.

.github/workflows/ci.yml
```diff
@@ -121,3 +121,11 @@ jobs:
         env:
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+
+  cd:
```
spirosoik, line 125: could you elaborate a bit more?

fmartingr: This is how we do it at the open source project I help maintain, we have the flows with the actual logic get invoked from the flows that has the triggers. Here basically we invoke the CD if lint/test/build passes and where the reference comes from a version tag (not 100% sure about this one, since that could be `master` as well, but I don't know if we want it to trigger on each push to master and/or version creations)

spirosoik: we want only tags actually to trigger the CD part.

fmartingr:
Replaced by #832

PR #832, opened 2023-01-13
Fix latest tag docker push and ignore CD on master

spirosoik:
#### Summary
We removed the latest tag during the migration and now we return it back for docker images. In parallel we ignore CD during master branch.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-4840

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,6 +2,7 @@ name: cd
 on:
   workflow_run:
     workflows: ["ci"]
+    branches-ignore: [master]
     types:
       - completed
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -2,6 +2,7 @@ name: cd
 on:
   workflow_run:
     workflows: ["ci"]
+    branches-ignore: [master]
```
fmartingr, line 5: So we do not want to run this workflow after CI has finished successfully? That's what I was trying to do on #829

fmartingr (approved):
👍

.github/workflows/cd.yml
spirosoik: yeah we don't need to run it on master.

phoinixgrr (approved):
Just a small note. Non blocking. 

Do you think we could  remove the bash script "push-docker.sh"? 
This functionality, could be part of the actual ghaction. Thus removing extra clutter and moving to a more simple approach.

spirosoik:
> Just a small note. Non blocking.
> 
> Do you think we could remove the bash script "push-docker.sh"? This functionality, could be part of the actual ghaction. Thus removing extra clutter and moving to a more simple approach.

ideally no, I would like to have repetitive. I would consider us though to be part of Makefile as in rtcd

PR #833, opened 2023-01-13
Tag docker images with latest tag

gabrieljackson:
```release-note
None
```

### DIFF
--- a/scripts/push-docker.sh
+++ b/scripts/push-docker.sh
@@ -7,12 +7,25 @@ set -euox
 
 echo $DOCKERHUB_TOKEN | docker login --username $DOCKERHUB_USERNAME --password-stdin
 
-if [ "$TAG" = "master" ] || [ "$TAG" = "main" ]; then
-    TAG=latest
+if [ "$COMMIT_SHA" = "" ]; then
+    echo "COMMIT_SHA was not provided"
+    exit 1
 fi
 
-docker tag mattermost/mattermost-cloud:test mattermost/mattermost-cloud:$TAG
-docker tag mattermost/mattermost-cloud-e2e:test mattermost/mattermost-cloud-e2e:$TAG
+echo "Tagging images with SHA $COMMIT_SHA"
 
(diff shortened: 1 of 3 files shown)


scripts/push-docker.sh
```diff
@@ -7,12 +7,25 @@ set -euox
...
 
-docker push mattermost/mattermost-cloud:$TAG
-docker push mattermost/mattermost-cloud-e2e:$TAG
+docker tag mattermost/mattermost-cloud:test mattermost/mattermost-cloud:$COMMIT_SHA
+docker tag mattermost/mattermost-cloud-e2e:test mattermost/mattermost-cloud-e2e:$COMMIT_SHA
+
+docker push mattermost/mattermost-cloud:$COMMIT_SHA
+docker push mattermost/mattermost-cloud-e2e:$COMMIT_SHA
+
+if [ "$REF_NAME" = "master" ] || [ "$REF_NAME" = "main" ]; then
```
fmartingr, line 23: Shouldn't this be `==`?

```suggestion
if [ "$REF_NAME" == "master" ] || [ "$REF_NAME" == "main" ]; then
```

fmartingr (commented):
LGTM except for this one comment

mirshahriar (approved):

scripts/push-docker.sh
mirshahriar: For single [, we can use =. For [[, we can use ==

gabrieljackson: Yeah, this is my understanding as well. For the record I hate shell scripting and could be wrong.

mirshahriar: https://linuxize.com/post/how-to-compare-strings-in-bash/#comparison-operators

fmartingr: My bash is a bit rusted as well, that's why I wasn't 100% sure.

fmartingr: Was chcking more information about it: https://unix.stackexchange.com/a/306115

PR #840, opened 2023-01-18
[MM-49713] Enable rotator on resize operations when kops require rolling-updates

fmartingr:
#### Summary

Determine when to use rotator on cluster resize by performing a `kops rolling-update cluster` in dry-run mode first and analyzing the output. If the output says: "_No rolling-update required._" then we can avoid enabling and running the rotator and just invoke `kops` as usual. If that is not present, we enable and run the rotator previous to the `kops` call so the nodes are handled by the rotator instead of by kops.

This is an ugly way of checking if kops is about to rotate the nodes, but since that command doesn't support JSON output (or any machine-friendly output), this is the easiest way I found to get that information out of it.

Unsure if this also applies to [upgrade](https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/kops_provisioner_cluster.go#L293) as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49713

#### Release Note

```release-note
Fixes rotator usage on cluster resize operations
```

### DIFF
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -548,16 +548,23 @@ func (provisioner *KopsProvisioner) ResizeCluster(cluster *model.Cluster, awsCli
 		return err
 	}
 
-	// TODO: Find out where to put this logic correctly
-	// if cluster.ProvisionerMetadataKops.RotatorRequest.Config != nil {
-	// 	if *cluster.ProvisionerMetadataKops.RotatorRequest.Config.UseRotator {
-	// 		logger.Info("Using node rotator for node resize")
-	// 		err = provisioner.RotateClusterNodes(cluster)
-	// 		if err != nil {
-	// 			return err
-	// 		}
-	// 	}
-	// }
+	requiresRollingUpdate, err := kops.RollingUpdateClusterRequired(kopsMetadata.Name)
+	if err != nil {
+		return err
+	}
+
+	if requiresRollingUpdate {
+		logger.Info("Rolling update is required")
+		if cluster.ProvisionerMetadataKops.RotatorRequest.Config != nil {
+			if *cluster.ProvisionerMetadataKops.RotatorRequest.Config.UseRotator {
+				logger.Info("Using node rotator for node resize")
+				err = provisioner.RotateClusterNodes(cluster)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
 
 	err = kops.RollingUpdateCluster(kopsMetadata.Name)
 	if err != nil {
--- a/internal/provisioner/kops_provisioner_cluster.go
+++ b/internal/provisioner/kops_provisioner_cluster.go
@@ -550,12 +550,12 @@ func (provisioner *KopsProvisioner) ResizeCluster(cluster *model.Cluster, awsCli
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


fmartingr:
I just tested a resize operation from 1 to 2, then 2 to 3 min nodes, and it didn't require a rolling release (so the rotator is not called):

Command (I had previously invoked this to min 1):
```
$ cloud cluster resize --cluster CLUSTERID --size-node-min-count 3 --size-node-max-count 3
```

Server: 
```
...
INFO[2023-01-20T14:35:05+01:00] Invoking command                              args="[/opt/homebrew/bin/kops rolling-update cluster --name=CLUSTERID-kops.k8s.local --state=s3://BUCKET]" cluster=CLUSTERID cmd=/opt/homebrew/bin/kops instance=INSTANCEID provisioner=kops run=RUNID
INFO[2023-01-20T14:35:05+01:00] [kops] FeatureFlag "SpecOverrideFlag"=true    cluster=CLUSTERID instance=INSTANCEID provisioner=kops run=RUNID
INFO[2023-01-20T14:35:10+01:00] [kops] NAME                     STATUS  NEEDUPDATE      READY   MIN     TARGET  MAX     NODES  cluster=CLUSTERID instance=INSTANCEID provisioner=kops run=RUNID
INFO[2023-01-20T14:35:10+01:00] [kops] master-us-east-1a        Ready   0               1       1       1       1       1   cluster=CLUSTERID instance=INSTANCEID provisioner=kops run=RUNID
INFO[2023-01-20T14:35:10+01:00] [kops] nodes-us-east-1a         Ready   0               3       3       3       3       2    cluster=CLUSTERID instance=INSTANCEID provisioner=kops run=RUNID
INFO[2023-01-20T14:35:10+01:00] [kops] No rolling-update required.            cluster=CLUSTERID instance=INSTANCEID provisioner=kops run=RUNID
...
```

gabrieljackson (approved):
The code for this looks solid. I know it was already covered, but we are at the mercy of kops not changing that text in future releases. It will have to be something that we try to test when we update the kops binary we use.

fmartingr:
> The code for this looks solid. I know it was already covered, but we are at the mercy of kops not changing that text in future releases. It will have to be something that we try to test when we update the kops binary we use.

Since updating kops is something we need to do with the most care possible anyway (since this is not the only thing that could break), I'd say just testing for a rollling-update command to check the text when we upgrade our kops version should be enough.

I took a quick look at latest kops and that string [is still there](https://github.com/kubernetes/kops/blob/master/cmd/kops/rolling-update_cluster.go#L438).

internal/provisioner/kops_provisioner_cluster.go
```diff
@@ -548,16 +548,23 @@ func (provisioner *KopsProvisioner) ResizeCluster(cluster *model.Cluster, awsCli
...
-	// 			return err
-	// 		}
-	// 	}
-	// }
+	requiresRollingUpdate, err := kops.RollingUpdateClusterRequired(kopsMetadata.Name)
+	if err != nil {
+		return err
+	}
+
+	if requiresRollingUpdate {
```
mirshahriar, line 556: nit: looks like this boolean `requiresRollingUpdate` should be something like `requiresClusterRotation`.  Because whether this value is `true` or `false`, we always do `RollingUpdateCluster`.

fmartingr: I like that name more, yeah.

PR #841, opened 2023-01-18
Add installation deletion pending expiration control

gabrieljackson:
This change introduces logic to control when installations that are pending deletion are fully deleted. Installations now have individual timestamps when they will exit the pending deletion state.

Installations that are currently pending deletion with no expiration value will fall back and use the existing logic based on installation change events.

When installation deletion is requested, the expiry time is set to the default future time as defined in the provisioning server. The value can then be updated if needed to happen sooner or later. As before, the installation deletion can be cancelled if it is still pending.

Fixes https://mattermost.atlassian.net/browse/MM-47409

```release-note
Add installation deletion pending expiration control
```

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -235,7 +235,6 @@ func newCmdInstallationUpdateDeletion() *cobra.Command {
 		},
 		PreRun: func(cmd *cobra.Command, args []string) {
 			flags.clusterFlags.addFlags(cmd)
-			return
 		},
 	}
 	flags.addFlags(cmd)
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -214,7 +214,6 @@ func newCmdInstallationUpdateDeletion() *cobra.Command {
 		Short: "Updates the pending deletion parameters of an installation.",
 		RunE: func(command *cobra.Command, args []string) error {
 			command.SilenceUsage = true
-			flags.installationDeletionPatchRequestOptionsChanged.addFlags(command)
 			client := model.NewClient(flags.serverAddress)
 
 			request := &model.PatchInstallationDeletionRequest{}
@@ -236,6 +235,7 @@ func newCmdInstallationUpdateDeletion() *cobra.Command {
 		},
 		PreRun: func(cmd *cobra.Command, args []string) {
 			flags.clusterFlags.addFlags(cmd)
+			flags.installationDeletionPatchRequestOptionsChanged.addFlags(cmd)
 		},
 	}
 	flags.addFlags(cmd)
@@ -831,7 +831,7 @@ func executeInstallationDeploymentReportCmd(flags installationDeploymentReportFl
 			return err
 		}
 		for _, event := range events {
-			output += fmt.Sprintf("%s - %s > %s\n", model.CompleteTimeStringFromMillis(event.Event.Timestamp), event.StateChange.OldState, event.StateChange.NewState)
+			output += fmt.Sprintf("%s - %s > %s\n", model.DateTimeStringFromMillis(event.Event.Timestamp), event.StateChange.OldState, event.StateChange.NewState)
 		}
 	}
 
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -214,10 +214,11 @@ func newCmdInstallationUpdateDeletion() *cobra.Command {
 		Short: "Updates the pending deletion parameters of an installation.",
 		RunE: func(command *cobra.Command, args []string) error {
 			command.SilenceUsage = true
+			flags.installationDeletionPatchRequestOptionsChanged.addFlags(command)
 			client := model.NewClient(flags.serverAddress)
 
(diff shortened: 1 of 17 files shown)


model/installation_states.go
```diff
@@ -221,6 +221,9 @@ var (
 			InstallationStateDeletionPendingRequested,
 			InstallationStateDeletionPendingInProgress,
 		},
+		InstallationStateDeletionPending: {
+			InstallationStateDeletionPending,
```
fmartingr, line 225: Is this the only state this can move to?

model/time.go
```diff
@@ -8,19 +8,29 @@ import "time"
...
 
-// GetDateString returns a standard date string from millis.
-func GetDateString(millis int64) string {
+// DateStringFromMillis returns a standard date string from millis.
+func DateStringFromMillis(millis int64) string {
 	return TimeFromMillis(millis).Format("Jan 2 2006")
 }
 
+// CompleteTimeStringFromMillis returns a standard complete time string from millis.
+func CompleteTimeStringFromMillis(millis int64) string {
```
fmartingr, line 30: `DateTimeStringFromMillis` ?

model/installation_request_test.go
```diff
@@ -724,3 +821,7 @@ func TestNewAssignInstallationGroupFromReader(t *testing.T) {
 func sToP(s string) *string {
 	return &s
 }
+
+func iToP(i int64) *int64 {
```
fmartingr, line 825: We need to refactor this functions to a common place, we have them spread all over test files.

model/installation_states.go
gabrieljackson: Yep! That's right.

model/installation_request_test.go
gabrieljackson: Yeah, agreed. I will create a ticket to cleanup everything.

cmd/cloud/server.go
```diff
@@ -418,15 +418,16 @@ func executeServerCmd(flags serverFlags) error {
...
-		Logger:        logger,
+		Store:                             sqlStore,
+		Supervisor:                        standardSupervisor,
+		Provisioner:                       clusterProvisioner,
+		DBProvider:                        resourceUtil,
+		EventProducer:                     eventsProducer,
+		Environment:                       awsClient.GetCloudEnvironmentName(),
+		AwsClient:                         awsClient,
+		Metrics:                           cloudMetrics,
+		InstallationDeletionExpiryDefault: flags.installationDeletionPendingTime,
```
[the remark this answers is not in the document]
gabrieljackson: We want this value to be configured with the server startup flag so I could create a var with a Setter and Getter in the `model` package. Do you think that would be a cleaner solution?

internal/supervisor/installation_deletion.go
```diff
@@ -165,34 +165,44 @@ func (s *InstallationDeletionSupervisor) transitionInstallation(installation *mo
...
-	logger = logger.WithField("time-spent-pending-deletion", timeSincePending.String())
-	if timeSincePending < s.deletionPendingTime {
-		timeUntilDeletion := s.deletionPendingTime - timeSincePending
-		logger.WithField("time-until-deletion", timeUntilDeletion.String()).Debug("Installation is not ready for deletion")
-		return model.InstallationStateDeletionPending
+	if installation.DeletionPendingExpiry != 0 {
+		// Primary deletion pending check.
+		if model.GetMillis() < installation.DeletionPendingExpiry {
+			timeUntilDeletion := time.Until(model.TimeFromMillis(installation.DeletionPendingExpiry))
+			logger.WithField("time-until-deletion", timeUntilDeletion.Round(time.Second).String()).Debug("Installation is not ready for deletion")
```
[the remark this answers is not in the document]
gabrieljackson: Hmm, I would actually prefer to keep them the same for now so we can easily review all the logs for pending deletions. The log line below has an extra field which we can use to differentiate if we really need to. How does that sound?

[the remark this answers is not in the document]
fmartingr: But is the same field, right?

cmd/cloud/server.go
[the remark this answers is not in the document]
fmartingr: If this is something configurable on startup I understand that is built that way, since I didn't see that I wasn't sure. 👍

cmd/cloud/installation.go
```diff
@@ -204,6 +206,43 @@ func newCmdInstallationDelete() *cobra.Command {
...
 
+func newCmdInstallationUpdateDeletion() *cobra.Command {
+	var flags installationUpdateDeletionFlags
+
+	cmd := &cobra.Command{
+		Use:   "update-deletion",
+		Short: "Updates the pending deletion parameters of an installation.",
+		RunE: func(command *cobra.Command, args []string) error {
+			command.SilenceUsage = true
+			flags.installationDeletionPatchRequestOptionsChanged.addFlags(command)
```
mirshahriar, line 217: This will work here too. But I have added this in `PreRun`

mirshahriar (approved):
LGTM

PR #844, opened 2023-01-20
[MM-49778] Fix kops creating a temporary file instead of a directory for the kops cache

fmartingr:
#### Summary

Caused by: #820 

The removal of deprecation warnings refactored a `ioutil.TempDir` into a `os.CreateTemp` by mistake, instead of using `os.MkdirTemp` as it was expected. This was causing the kops client to error all the time due to it being unable to write files into a file, instead of a directory which is the thing it expects.

I also added more verbose errors for the kops command invokation, so we can easily spot what's happening directly in the logs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49778

#### Release Note

```release-note
Fixed kops unable to locate kubecfg files
```

### DIFF
--- a/internal/tools/kops/cmd.go
+++ b/internal/tools/kops/cmd.go
@@ -33,7 +33,7 @@ func New(s3StateStore string, logger log.FieldLogger) (*Cmd, error) {
 		return nil, errors.Wrap(err, "failed to find kops installed on your PATH")
 	}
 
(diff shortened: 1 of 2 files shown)


stafot:
@fmartingr please add a release note. In my view it is a bug fix here.

stylianosrigas (approved):

PR #847, opened 2023-01-30
[MM-49747] Check installation update flags changes

fmartingr:
#### Summary

Added flag change check to installation update command.

Checked other update commands and those already check for changes except for cluster update which only accepts one change (allow installations).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49747

#### Release Note

```release-note
Check specified installation update flags before creating an update request
```

### DIFF
--- a/cmd/cloud/installation_flag.go
+++ b/cmd/cloud/installation_flag.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
 package main
 
 import (
@@ -73,7 +77,24 @@ func (flags *installationCreateFlags) addFlags(command *cobra.Command) {
 	_ = command.MarkFlagRequired("dns")
 }
 
+type installationPatchRequestChanges struct {
+	ownerIDChanged bool
+	versionCHanged bool
+	imageChanged   bool
+	sizeChanged    bool
+	licenseChanged bool
+}
+
+func (flags *installationPatchRequestChanges) addFlags(command *cobra.Command) {
+	flags.ownerIDChanged = command.Flags().Changed("owner")
+	flags.versionCHanged = command.Flags().Changed("version")
+	flags.imageChanged = command.Flags().Changed("image")
+	flags.sizeChanged = command.Flags().Changed("size")
+	flags.licenseChanged = command.Flags().Changed("license")
+}
+
 type installationPatchRequestOptions struct {
+	installationPatchRequestChanges
 	ownerID            string
 	version            string
 	image              string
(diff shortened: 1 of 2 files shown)


gabrieljackson (approved):
Awesome!

cmd/cloud/installation_flag.go
```diff
@@ -89,10 +110,34 @@ func (flags *installationPatchRequestOptions) addFlags(command *cobra.Command) {
 	command.Flags().StringVar(&flags.image, "image", "mattermost/mattermost-enterprise-edition", "The Mattermost container image to use.")
 	command.Flags().StringVar(&flags.size, "size", model.InstallationDefaultSize, "The size of the installation. Accepts 100users, 1000users, 5000users, 10000users, 25000users, miniSingleton, or miniHA. Defaults to 100users.")
 	command.Flags().StringVar(&flags.license, "license", "", "The Mattermost License to use in the server.")
-
 	command.Flags().StringArrayVar(&flags.mattermostEnv, "mattermost-env", []string{}, "Env vars to add to the Mattermost App. Accepts format: KEY_NAME=VALUE. Use the flag multiple times to set multiple env vars.")
 	command.Flags().BoolVar(&flags.mattermostEnvClear, "mattermost-env-clear", false, "Clears all env var data.")
+}
+
+func (flags *installationPatchRequestOptions) GetPatchInstallationRequest() *model.PatchInstallationRequest {
+	request := model.PatchInstallationRequest{}
```
mirshahriar, line 118: Don't we want to use `default` flag values? In this case, If **CLI** doesn't set a flag explicitly, the pointer will be `nil` which doesn't match with previous behavior.

cmd/cloud/installation_flag.go
```diff
@@ -113,6 +158,25 @@ func (flags *installationUpdateFlags) addFlags(command *cobra.Command) {
 	_ = command.MarkFlagRequired("installation")
 }
 
+func (flags *installationUpdateFlags) GetPatchInstallationRequest() (*model.PatchInstallationRequest, error) {
+	request := flags.installationPatchRequestOptions.GetPatchInstallationRequest()
+
+	envVarMap, err := parseEnvVarInput(flags.mattermostEnv, flags.mattermostEnvClear)
```
mirshahriar, line 164: Can we make it `mattermostEnv` to match it with `priorityEnv`?

cmd/cloud/installation_flag.go
fmartingr: That's the purpose of this PR. Since this is a patch operation, we do not want to use all the default values since we may only want to update one attribute.

mirshahriar: Thats right. Thanks

mirshahriar (approved):

cmd/cloud/installation_flag.go
mirshahriar: Oh. I mean the variable name. `envVarMap` -> `mattermostEnv `

fmartingr: Oh, I was wondering if you referred to the logic itself. Yeah, that make sense.

PR #856, opened 2023-02-08
[MM-50390] fix: keep trying to claim available VPCs

fmartingr:
#### Summary

Keep trying to Claim a VPC in case of errors.

I'm assuming this was set up like it was for a reason, so I hope @gabrieljackson knows. My only concern with this is that if something related to tagging fails in between the claiming, we will leave bogus tags on all VPCs, instead of only one. We could use some cancel function to leave everything as it was if any error happens? (making claiming somehow atomic)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50390

#### Release Note

```release-note
fix: keep trying to claim VPCs in case of errors
```

### DIFF
--- a/internal/tools/aws/cluster_management.go
+++ b/internal/tools/aws/cluster_management.go
@@ -333,27 +333,63 @@ func (a *Client) claimVpc(clusterResources ClusterResources, cluster *model.Clus
 		}
 	}
 
+	cleanupTagResources := []TagResourceInput{}
+	currentTags := NewTagsFromEC2Tags(vpcs[0].Tags)
+
+	cleanup := func() error {
+		cleanupLogger := logger.WithField("vpc_id", clusterResources.VpcID)
+		cleanupLogger.Info("Cleaning up VPC and related resources after failing to claim it")
+		for _, op := range cleanupTagResources {
+			if err := a.TagResource(op.ResourceID, op.Key, op.Value, logger); err != nil {
+				cleanupLogger.WithField("resource_id", op.ResourceID).WithError(err).Error("failed to tag resouce to its original value")
+				return errors.Wrap(err, "failed to tag resource to its original value")
+			}
+		}
+
+		return nil
+	}
+
 	if claimSecondaryCluster {
 		err = a.TagResource(clusterResources.VpcID, trimTagPrefix(VpcSecondaryClusterIDTagKey), cluster.ID, logger)
 		if err != nil {
 			return errors.Wrapf(err, "unable to update %s", VpcClusterIDTagKey)
 		}
+		cleanupTagResources = append(cleanupTagResources, TagResourceInput{
+			ResourceID: clusterResources.VpcID,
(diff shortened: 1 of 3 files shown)


mirshahriar:
@fmartingr, I was also thinking about this same solution. But in this case, that particular VPC will never be claimed unless we fix that manually.

fmartingr:
> @fmartingr, I was also thinking about this same solution. But in this case, that particular VPC will never be claimed unless we fix that manually.

Depends on the error. My only experience with failed VPC claiming was when the resource was cleaned up manually and tags wasn't properly set (for example, an empty cluster ID instead of `none`).

mirshahriar:
> > @fmartingr, I was also thinking about this same solution. But in this case, that particular VPC will never be claimed unless we fix that manually.
> 
> Depends on the error. My only experience with failed VPC claiming was when the resource was cleaned up manually and tags wasn't properly set (for example, an empty cluster ID instead of `none`).

I am thinking about this particular scenario. Do we want to fix that manually after cleanup?

gabrieljackson (approved):
I think this change is okay for now. While we did purposefully return with an error before if claim failed to prevent multiple VPC tag issues from occurring, this will still help us know which VPCs are not in a valid state for us to act on. We should probably discuss a more robust VPC claim mechanism for the future. I have never been completely happy with this so we could definitely improve it in a future PR.

mirshahriar (approved):

fmartingr:
Let me think a bit more about a proper cleanup, just in case any errors came up, so we don't came up with properly untagged VPCs.

fmartingr:
@mirshahriar @gabrieljackson  Thinking about this one. Would be enough to simply `releaseVpc` the VPC that gave an error when claiming it? I was thinking on a way to return a cleanup function to untag everything that has been tagged up until the error ocurred, but this may be simpler and more efficient I think? Since the same untagging should be done in both scenarios.

internal/tools/aws/cluster_management.go
```diff
@@ -235,13 +235,17 @@ func (a *Client) GetAndClaimVpcResources(cluster *model.Cluster, owner string, l
...
-			logger.Warn(err)
+			logger.WithError(err).WithField("vpc", *vpc.VpcId).Warn("Couldn't get cluster resources for VPC")
 			continue
 		}
 
 		err = a.claimVpc(clusterResources, cluster, owner, logger)
 		if err != nil {
-			return clusterResources, err
+			logger.WithError(err).WithField("vpc", *vpc.VpcId).Warn("Couldn't claim VPC")
+			if err := a.releaseVpc(cluster, logger); err != nil {
```
mirshahriar, line 245: This will not work because of the same reason. The `CloudClusterID` tag is missing.

gabrieljackson: Yeah, we should test this out at least first in the dev account by manually messing with the tags. I am thinking there are a few scenarios where this will fail, but testing will validate that.

PR #858, opened 2023-02-13
Fix workbench command removing workbench directory

gabrieljackson:
This change prevents the workbench command from silently cleaning up the workbench directory as it exits.

```release-note
Fix workbench command removing workbench directory
```


stylianosrigas (approved):
Thanks :)

fmartingr (approved):
I will just trust you on this.

PR #860, opened 2023-02-15
[MM-50487] Enable EKS cluster support without server flag

mirshahriar:
#### Summary

Now we support both provisioners running at the same time.

#### Ticket Link

Fixes [MM-50487](https://mattermost.atlassian.net/browse/MM-50487) & [MM-50729](https://mattermost.atlassian.net/browse/MM-50729)

#### Release Note

```release-note
Support KOPS and EKS running in the Server at the same time.
```


fmartingr (commented):
Hey, awesome work on this. Way more changes than I expected to be honest, which makes me wonder. 

I like how it is right now since it's a good translation over the codebase we already have, but all those interfaces in each supervisor looks like extra steps for the same goal to me. Why don't we define a main `Provisioner` interface and make every supervisor have a `GetSupervisor` method that returns the required one? i understand that this is the "go way" of doing things, but we are constraining ourselves in each supervisor when the actual logic is elsewhere in the same module.

This is more a general comment/complain, nothing wrong with your changes! I'd like to know @gabrieljackson insights on this.

mirshahriar:
> Why don't we define a main `Provisioner` interface and make every supervisor have a `GetSupervisor` method that returns the required one?

@fmartingr, Something like this?

```
type ClusterSupervisor struct {
	provisioner    ProvisionerOption
}

// This will only return the ClusterProvisioner interface either for kops or eks
func (s *ClusterSupervisor) GetProvisioner(provisioner string) ClusterProvisioner {
	return s.provisioner.getProvisioner(provisioner)
}

// This will call for the provisioner either kops or eks
func (s *ClusterSupervisor) createCluster(cluster *model.Cluster, logger log.FieldLogger) {
	if s.GetProvisioner(cluster.Provisioner).PrepareCluster(cluster) {}
}
```

gabrieljackson (commented):
Yep, I think I am on the same page as you two on this. This change looks good to me, but we can also go the other direction if we think it's cleaner. 0/5 from me so just tag me for review when you want me to take another look.

mirshahriar:
We can keep this on your priority list so that we can refactor some needless `interface{}`. For this PR, we are good to merge. This will unblock some other tasks.

@fmartingr & @gabrieljackson

fmartingr (approved):
LGTM. My only nit is the attributes that are named `XXXOperator` when it should be provisioner, but we can change that later on, not an issue atm.

gabrieljackson (approved):
👍

cmd/cloud/server.go
```diff
@@ -274,23 +273,25 @@ func executeServerCmd(flags serverFlags) error {
...
 		provisioningParams,
-		resourceUtil,
 		awsClient,
+		sqlStore,
 		logger,
 	)
 
-	provisionerOptionForSupervisor := supervisor.GetProvisionerOption(eksProvisioner, kopsProvisioner)
-	provisionerOptionForAPI := api.GetProvisionerOption(eksProvisioner, kopsProvisioner)
+	provisionerInterface := provisioner.NewProvisioner(
```
fmartingr, line 287: Pretty sure this is not an interface, just a provisioner.

internal/provisioner/kops_provisioner.go
```diff
@@ -210,8 +200,8 @@ func (provisioner *KopsProvisioner) k8sClient(clusterName string, logger log.Fie
 	var k8sClient *k8s.KubeClient
 	k8sClient, err = k8s.NewFromFile(configLocation, logger)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "failed to create k8s client from file")
+		return nil, errors.Wrap(err, "failed to create k8s client from file")
 	}
 
-	return k8sClient, invalidateOnError, nil
```
fmartingr, line 216: Was `invalidateOnError` unnecessary?

internal/provisioner/provisioner.go
```diff
@@ -0,0 +1,70 @@
...
+	ClusterProvisionerOption
+	params         ProvisioningParams
+	resourceUtil   *utils.ResourceUtil
+	backupOperator *BackupOperator
+	store          *store.SQLStore
+	logger         log.FieldLogger
+	kubeOption     ClusterProvisionerOption
+}
+
+var _ supervisor.ClusterInstallationProvisioner = (*Provisioner)(nil)
```
fmartingr, line 34: What am I looking at here?

fmartingr (approved):
I think that the changes are good for the purpose, and you did a great job streamlining the provisioner usage for Kops and EKS here. The individual interfaces for the supervisors are still an itch for me, but they do no harm. I've left some questions around.

cmd/cloud/server.go
mirshahriar: Right. Thanks. I will rename it

internal/provisioner/kops_provisioner.go
mirshahriar: `invalidateOnError` was never used.  I mean It was called by defer but defer was set after the error check. So It was necessary. Thats why I have removed.

internal/provisioner/provisioner.go
mirshahriar: This is just a nice way to assert that `struct` implements expected set of interfaces.

mirshahriar:
> The individual interfaces for the supervisors are still an itch for me

I am considering the flow as a hexagonal architecture. `Supervisor` is in the inner layer, whilst `Provisioner` is in the outer layer. To eliminate outward flow _(which is forbidden in Hex Arch)_, we only need to employ `dependency inversion`. Hence, each supervisor utilizes an `interface{}`.

mirshahriar:
@gabrieljackson, I have made some changes.

internal/provisioner/provisioner.go
gabrieljackson: Huh, interesting. Never seen this before.

internal/api/cluster.go
```diff
@@ -108,29 +107,23 @@ func handleCreateCluster(c *Context, w http.ResponseWriter, r *http.Request) {
...
-		w.WriteHeader(http.StatusBadRequest)
-		return
-	}
-	if c.Provisioner.ProvisionerType() == provisioner.EKSProvisionerType && createClusterRequest.EKSConfig == nil {
-		c.Logger.Error("invalid request: EKSConfig is required when provisioner type is 'eks")
-		w.WriteHeader(http.StatusBadRequest)
-		return
+	provisioner := "kops"
+	if createClusterRequest.Provisioner != "" {
+		provisioner = createClusterRequest.Provisioner
```
gabrieljackson, line 112: Maybe I missed it, but is this validated anywhere so that I can't send a random value? We should maybe do that and also maybe used constant values for the strings?

gabrieljackson (approved):
Thanks for taking the time to show this in code. When I saw that your PR had more deleted lines than added I knew you probably were onto something. This looks great!

I only have one question that I won't block on.

internal/api/cluster.go
mirshahriar: Thank you for your review! I appreciate your suggestion to include validation, which I intentionally deferred to the next PR. However, I want to assure you that the current usage is limited to `eks` or `kops` (default), so sending a random value should not cause any issue. Thank you again for taking the time to review!

PR #862, opened 2023-02-17
Increase time of command running

stafot:
2 minutes aren't enough for DB migration operations

#### Summary
Increase time of command running

#### Release Note

```release-note
Increase time of command running
```


stafot:
@fmartingr If there aren't any objections due to other implications that we cannot think of please approve and cut another pathc realease with this fix.

fmartingr (approved):

fmartingr:
> @fmartingr If there aren't any objections due to other implications that we cannot think of please approve and cut another pathc realease with this fix.

Would it be any weird scenario of me cutting a 0.69.5 release where there is a 0.70.0 already present? I suppose nothing wrong is going to happen, but just wanted to double check.

angeloskyratzakos (approved):

stafot:
I think the release should be 0.69.4 @fmartingr

PR #863, opened 2023-02-17
[MM-47885] Check subnet types are correct after a kops update

fmartingr:
#### Summary

This PR adds a new AWS Client method to fix kops switching the subnet `SubnetType` tag to its own values after running `kops update` commands and before claiming a VPC.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47885

#### Release Note

```release-note
check subnet types are correct when claiming vpc and performing kops updated
```


internal/tools/aws/cluster_management.go
```diff
@@ -52,12 +52,74 @@ func (cr *ClusterResources) IsValid() error {
...
+}
+
+func (a *Client) fixSubnetTypeTag(subnet ec2Types.Subnet, tagKey, tagIncorrectValue, tagCorrectValue, vpcID string, logger log.FieldLogger) error {
+	for _, tag := range subnet.Tags {
+		if *tag.Key == tagKey && *tag.Value == tagIncorrectValue {
+			logger.WithFields(log.Fields{
+				"vpc_id":    vpcID,
+				"subnet_id": *subnet.SubnetId,
+			}).Warnf("Incorrect SubnetType for Subnet, setting it to '%s' from '%s'", tagCorrectValue, tagIncorrectValue)
+			return a.TagResource(*subnet.SubnetId, "SubnetType", tagCorrectValue, logger)
```
mirshahriar, line 107:
```suggestion
			return a.TagResource(*subnet.SubnetId, tagKey, tagCorrectValue, logger)
```

fmartingr:
> LGTM. Only one question. Do we need this fix every time after every kops update?

Some kops commands are the one that ruins the tagging, notably the update command that we call several times. I think we are mostly good only on the claiming (fixing it before we claim), but some subnet filtering will fail releasing due to having incorrect tags as well, that's why I'd rather go safe and check the subnets when we call kops.

stafot:
@mirshahriar Yes every kops update messes them up as these values are hardcoded on kops sourcecode :(

stafot (approved):

PR #864, opened 2023-02-20
Resolve issues with the EKS cluster creation.

mirshahriar:
#### Summary

We are creating a Launch Template for the NodeGroup with provided AMI which takes User Data like `max-pods` per Node.

---

By default, an EKS cluster uses **VPC CNI**, which isn't suitable for us  because it can only handle a limited number of IP addresses. Thats why we need alternatives and  for now, **Calico networking** has been added as default. But it comes with another EKS's limitation. 

> Calico networking cannot currently be installed on the EKS control plane nodes. As a result the control plane nodes will not be able to initiate network connections to Calico pods. (This is a general limitation of EKS's custom networking support, not specific to Calico.) As a workaround, trusted pods that require control plane nodes to connect to them, such as those implementing admission controller webhooks, can include hostNetwork:true in their pod spec

We have multiple admission controller setups such as `nginx`, `nginx-internal`.  We can set `hostNetwork:true` in pod spec. But again, we must consider how to resolve port disputes.. That's why I have disabled these admission controllers for EKS cluster.

---

We have added two more ClusterState `waiting-for-node` & `provision-in-progress`. Between cluster creation and node creation, we need to complete a few tasks. To prevent repetition, we have added these states.

#### Ticket Link

  Resolved https://mattermost.atlassian.net/browse/MM-50705

#### Release Note

```release-note
None
```


internal/api/helpers.go
```diff
@@ -91,3 +91,7 @@ func parseGroupConfig(u *url.URL) (bool, bool, error) {
 
 	return includeGroupConfig, includeGroupConfigOverrides, nil
 }
+
+func int32Ptr(i int32) *int32 {
```
fmartingr, line 95: You can use `ptr.Int()` from `smithy-go`.

internal/supervisor/cluster.go
```diff
@@ -292,7 +297,21 @@ func (s *ClusterSupervisor) checkClusterCreated(cluster *model.Cluster, logger l
...
+}
+
+func (s *ClusterSupervisor) checkNodesCreated(cluster *model.Cluster, logger log.FieldLogger) string {
+	ready, err := s.provisioner.CheckNodesCreated(cluster, s.aws)
+	if err != nil {
+		logger.WithError(err).Error("Failed to check if node creation finished")
+		return model.ClusterStateCreationFailed
+	}
+	if !ready {
+		logger.Info("Cluster not yet ready")
```
fmartingr, line 310:
```suggestion
		logger.Info("Cluster nodes are not ready yet")
```

fmartingr (approved):
This looks good, only some minor comments. Also, should we document this elsewhere or just having the comments in code is enough? At first I thought about adding it to a document but then we would have two maintain two different pieces of text. What do you think?

internal/api/helpers.go
mirshahriar: Right.

mirshahriar:
>  Should we document this elsewhere 

When the working flow is complete, we need documentation of it. We still have a lot of pieces missing, and the process could be changed.

model/cluster_states.go
```diff
@@ -11,6 +11,10 @@ const (
 	ClusterStateCreationRequested = "creation-requested"
 	// ClusterStateCreationInProgress is a cluster that is being actively created.
 	ClusterStateCreationInProgress = "creation-in-progress"
+	// ClusterStateWaitingForNode is a cluster that is waiting for nodes to be ready
+	ClusterStateWaitingForNode = "waiting-for-node"
```
gabrieljackson, line 15: nit: I think this sounds slightly better, but we can also change it later
```suggestion
	// ClusterStateWaitingForNodes is a cluster that is waiting for nodes to be ready
	ClusterStateWaitingForNodes = "waiting-for-nodes"
```

gabrieljackson (approved):
LGTM! Very nice.

model/cluster_states.go
mirshahriar: Right. Thanks

PR #874, opened 2023-03-02
Remove redundant or unneeded AWS client access

mirshahriar:
#### Summary

We are not need to pass an AWS client when the client is already present in the receiver.

#### Release Note

```release-note
NONE
```


fmartingr (approved):
Nice cleanup!

gabrieljackson (approved):
👍 Nice!

PR #876, opened 2023-03-02
rollback namespace deletion logic

mirshahriar:
#### Summary

Undo logic for namespace deletion with finalizer 

#### Release Note

```release-note
None
```


fmartingr (approved):
LGTM. But is the waiting for finalisers not needed in the end?

mirshahriar:
> LGTM. But is the waiting for finalisers not needed in the end?

We do but need more time to figure out the correct solution.

gabrieljackson (approved):

PR #877, opened 2023-03-03
[MM-51053] delete helm to cleanup LBs during cluster teardown

mirshahriar:
#### Summary

We need to delete helm to cleanup LBs while deleting cluster.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51053

#### Release Note

```release-note
None
```

### DIFF
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -94,7 +94,7 @@ func (d *helmDeployment) Exists() (bool, error) {
 	}
 
 	for _, chart := range list.asSlice() {
-		if chart.Name == d.chartDeploymentName && chart.Chart == d.chartName && chart.Namespace == d.namespace {
+		if chart.Name == d.chartDeploymentName && chart.Namespace == d.namespace {
 			return true, nil
 		}
 	}
--- a/internal/provisioner/helm_utils.go
+++ b/internal/provisioner/helm_utils.go
@@ -213,11 +213,12 @@ func upgradeHelmChart(chart helmDeployment, configPath string, logger log.FieldL
 // deleteHelmChart is used to delete Helm charts.
 func deleteHelmChart(chart helmDeployment, configPath string, logger log.FieldLogger) error {
 	arguments := []string{
-		"--debug",
-		"delete",
+		"uninstall",
+		chart.chartDeploymentName,
 		"--kubeconfig", configPath,
 		"--namespace", chart.namespace,
-		chart.chartDeploymentName,
+		"--wait",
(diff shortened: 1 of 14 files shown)


internal/provisioner/nginx.go
```diff
@@ -116,7 +116,11 @@ func (n *nginx) ActualVersion() *model.HelmUtilityVersion {
 }
 
 func (n *nginx) Destroy() error {
-	return nil
+	helm, err := n.NewHelmDeployment()
+	if err != nil {
+		return err
+	}
+	return helm.Delete()
```
gabrieljackson, line 123: I believe the helm client would return immediately after calling delete, but k8s may need some time to clean things up in the background afterwards. Do you know if that is correct or not? I am wondering if we need to add some wait logic to ensure there is enough time for a full cleanup.

internal/provisioner/helm_utils.go
```diff
@@ -94,7 +94,7 @@ func (d *helmDeployment) Exists() (bool, error) {
 	}
 
 	for _, chart := range list.asSlice() {
-		if chart.Name == d.chartDeploymentName && chart.Chart == d.chartName && chart.Namespace == d.namespace {
```
fmartingr, line 97: What was the problem with this check? just curious.

fmartingr (approved):
Changes look good to me, just wondering what implies waiting for all charts to get removed now. This mostly affects test/dev on a daily basis, but still.

internal/provisioner/helm_utils.go
mirshahriar: They are different entities. `d.chartName` holds `ingress-nginx/ingress-nginx` but `chart.Chart` holds `ingress-nginx-4.2.0`

mirshahriar:
> what implies waiting for all charts to get removed now

In current version, It takes 6 minutes to delete a **dev** cluster. If we delete all helms, it takes 10 minutes. However, deleting all of these helms concurrently (**goroutine**) only takes 5 minutes.

@fmartingr do you think we should go for `goroutine`?

fmartingr:
> > what implies waiting for all charts to get removed now
> 
> In current version, It takes 6 minutes to delete a **dev** cluster. If we delete all helms, it takes 10 minutes. However, deleting all of these helms concurrently (**goroutine**) only takes 5 minutes.
> 
> @fmartingr do you think we should go for `goroutine`?

With the numbers you provided, I don't think is worth it to be honest.

gabrieljackson (approved):
LGTM

PR #878, opened 2023-03-06
[MM-50968] Small fixes for e2e tests

fmartingr:
#### Summary

- Added a Debug configuration to E2E test output. `DEBUG=true`
- Refactored cluster cleanup (no more errors on trying to delete an already deleted cluster)
- Wait for subscription messages for 1 seconds before evaluating the test steps that came from subscriptions.
  - This (hopefully) fixes a race condition where the cluster has already been deleted (correctly) but the subscription message has not arrived yet, making the step-checking fail at the end of the tests.
  - I tried to switch this to a "broadcast" from the webhook (we only register the webhook, and we leave two channels, one for the suite and one for the events recorder to collect what's going on) but that added more complexity so I wanted to try this first. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50968

#### Release Note

```release-note
fix: e2e tests fixes and improvements
```

### DIFF
--- a/e2e/tests/cluster/cluster_test.go
+++ b/e2e/tests/cluster/cluster_test.go
@@ -73,7 +73,7 @@ func Test_ClusterLifecycle(t *testing.T) {
 			if err != nil {
 				test.Logger.WithError(err).Error("Error cleaning up installation")
 			}
-			err := test.ClusterSuite.DeleteCluster(context.Background())
+			err := test.ClusterSuite.Cleanup(context.Background())
 			if err != nil {
 				test.Logger.WithError(err).Error("Error cleaning up cluster")
 			}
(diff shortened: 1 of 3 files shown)


e2e/tests/cluster/cluster_test.go
```diff
@@ -86,6 +86,9 @@ func Test_ClusterLifecycle(t *testing.T) {
 	err = test.Run()
 	require.NoError(t, err)
 
+	// Make sure we wait for al subscription events
```
mirshahriar, line 89:
```suggestion
	// Make sure we wait for all subscription events
```

mirshahriar (approved):
Nicely Done.

gabrieljackson (approved):
LGTM

PR #880, opened 2023-03-08
[MM-50968] More information on E2E test logs

fmartingr:
#### Summary

- Added a simple way to capture if e2e tests receive any signals

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50968

#### Release Note

```release-note
feat: more verbose e2e tests
```


fmartingr:
Adding some eyes from @mattermost/cloud-sre just to be sure that the `HOSTNAME` thing will report something in our infra.

stylianosrigas (approved):

gabrieljackson (approved):

e2e/tests/cluster/suite.go
```diff
@@ -61,8 +62,9 @@ func SetupClusterLifecycleTest() (*Test, error) {
 	testID := model.NewID()
 	state.TestID = testID
 	logger := logrus.WithFields(map[string]interface{}{
-		"test":   "cluster-lifecycle",
-		"testID": testID,
+		"test":     "cluster-lifecycle",
+		"testID":   testID,
+		"hostname": os.Getenv("HOSTNAME"),
```
mirshahriar, line 67: What is the use case of this `hostname`?

e2e/tests/cluster/cluster_test.go
```diff
@@ -39,6 +41,15 @@ func TestMain(m *testing.M) {
 	code := m.Run()
 	state.EndTime = time.Now()
 
+	// Notify if we receive any signal
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
```
mirshahriar, line 46: Looks like these signals have no impact. Is it just for logs?

mirshahriar (approved):
I have two questions.

mirshahriar:
> `HOSTNAME` thing will report something in our infra.

What are we expecting from `os.Getenv("HOSTNAME")`? `Pod Name`?

e2e/tests/cluster/cluster_test.go
fmartingr: Yeah, just to register in logs if we received any signal, potentially from k8s during a node removal or something like that.

e2e/tests/cluster/suite.go
mirshahriar: I'm not sure whether we can get the Pod Name unless we include the following in the Pod Spec.

```yaml
      env:
        - name: HOSTNAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

mirshahriar: BTW, Pod name is already available in logs; `label: pod`.

fmartingr: > I'm not sure whether we can get the Pod Name unless we include the following in the Pod Spec.

Just checked and `HOSTNAME` env var is there without any configuration.

> BTW, Pod name is already available in logs; label: pod.

Oh my you're right. I assumed it won't be there since we didn't add it as a field, but prometheus gets it from the k8s cluster anyway. Have been reading logs all week and didn't notice 😓  Reverting this then!

PR #881, opened 2023-03-09
Refining Code Organization: moving methods, renaming and grouping files

mirshahriar:
#### Release Note

```release-note
None
```


fmartingr (approved):
Awesome work. I have only one nit on one filename, but so far it seems that no logic was harmed in the process 👍

internal/provisioner/installation_provisioner.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
mirshahriar: These three methods implement `InstallationProvisioner` interface which is used by **Installation Supervisor**. That's why I have kept them together in this file `installation_provisioner`.

mirshahriar:
> you have removed/not added the header boilerplate in several files

I have added "header" text to these new files. However, I can see that some existing files lack this text. I'm not touching those files in this PR.

model/provisioner_param.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
+package model
+
+// ProvisioningParams represent configuration used during various provisioning operations.
+type ProvisioningParams struct {
```
gabrieljackson, line 8: nit: I can see why you moved this config object to `model`, but it may still be better suited staying in `internal`. My general guideline for whether something should be in `model` or not is if there is a situation where an outside service may want to reference or use the object. In this case, ProvisioningParams is only really used in our internal supervisor/provisioner code so not sure it would be useful.

As I said, this is just a general guideline and not something I feel strongly on. Just figured I would mention it.

gabrieljackson (approved):
LGTM! Just one non-blocking comment.

model/provisioner_param.go
mirshahriar: That is right. To solve the cycle dependency, I must package it separately. What about a model `package` inside the `/internal/provisioner/model` directory?

mirshahriar: I think I have found a solution

PR #886, opened 2023-03-17
[MM-50721] Provisioner: Update webhooks and subscriptions resources to store custom client header config

fmartingr:
#### Summary

Allow webhooks and subscriptions to send headers long with their requests, providing headers in the format `Key=Value`, where `Value` is the header value or an environment variable name to get the value from.

Usage: `cloud webhook create --owner xx --url xx --header Key=Value --header-from-env Key2=ENV_VAR_NAME`

Struct used in storage:

``` js
[
  {
    "key": "key",
    "value": "value", // optional
    "valueFromEnv": "ENV_VAR_NAME", //optional
  }
]
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50721

#### Release Note

```release-note
feat: allow webhooks and subscriptions to send custom headers
```

### DIFF
--- a/cmd/cloud/events_subscription_flag.go
+++ b/cmd/cloud/events_subscription_flag.go
@@ -19,6 +19,7 @@ type subscriptionCreateFlags struct {
 	eventType        string
 	failureThreshold time.Duration
 	headers          map[string]string
+	headersFromEnv   map[string]string
 }
 
 func (flags *subscriptionCreateFlags) addFlags(command *cobra.Command) {
@@ -27,7 +28,8 @@ func (flags *subscriptionCreateFlags) addFlags(command *cobra.Command) {
 	command.Flags().StringVar(&flags.owner, "owner", "", "OwnerID of the subscription.")
 	command.Flags().StringVar(&flags.eventType, "event-type", string(model.ResourceStateChangeEventType), "Event type of the subscription.")
 	command.Flags().DurationVar(&flags.failureThreshold, "failure-threshold", 0, "Failure threshold of the subscription.")
-	command.Flags().StringToStringVar(&flags.headers, "headers", nil, "headers that should be sent with the request")
+	command.Flags().StringToStringVar(&flags.headers, "header", nil, "a header that should be sent with the request")
+	command.Flags().StringToStringVar(&flags.headersFromEnv, "header-from-env", nil, "a header that should be sent with the request, with values read from environment variables")
 	_ = command.MarkFlagRequired("url")
 	_ = command.MarkFlagRequired("owner")
 	_ = command.MarkFlagRequired("event-type")
--- a/cmd/cloud/events_subscription_flag.go
+++ b/cmd/cloud/events_subscription_flag.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
 package main
 
 import (
@@ -14,6 +18,7 @@ type subscriptionCreateFlags struct {
 	owner            string
 	eventType        string
 	failureThreshold time.Duration
+	headers          map[string]string
 }
 
 func (flags *subscriptionCreateFlags) addFlags(command *cobra.Command) {
@@ -22,6 +27,7 @@ func (flags *subscriptionCreateFlags) addFlags(command *cobra.Command) {
 	command.Flags().StringVar(&flags.owner, "owner", "", "OwnerID of the subscription.")
 	command.Flags().StringVar(&flags.eventType, "event-type", string(model.ResourceStateChangeEventType), "Event type of the subscription.")
 	command.Flags().DurationVar(&flags.failureThreshold, "failure-threshold", 0, "Failure threshold of the subscription.")
+	command.Flags().StringToStringVar(&flags.headers, "headers", nil, "headers that should be sent with the request")
 	_ = command.MarkFlagRequired("url")
 	_ = command.MarkFlagRequired("owner")
 	_ = command.MarkFlagRequired("event-type")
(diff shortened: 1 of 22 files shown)


model/webhook.go
```diff
@@ -6,9 +6,43 @@ package model
...
+func ParseHeadersFromStringMap(sm *StringMap) (map[string]string, error) {
+	if sm == nil {
+		return map[string]string{}, nil
+	}
+	headers := make(map[string]string, len(*sm))
+	var err error
+	for key, value := range *sm {
+		headerValue := value
+		if strings.HasPrefix(value, WebhookHeaderEnvironmentValuePrefix) {
+			headerValue = os.Getenv(strings.TrimPrefix(value, WebhookHeaderEnvironmentValuePrefix))
```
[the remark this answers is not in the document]
fmartingr: I ended up going this way because it wasn't so difficult to understand (IMO) and it made easier using the CLI (single flag) and manually editing the database (less payload). I'm not against the other way, but since this is more internal than external I'd rather use `--headers A=B,C=D` than `--header A=B --header-from-env C=D` or something like that. I could also use the same input and just internally marshall into that other format, but it would only add complexity and I don't see much benefit at the moment. 

If we already have some plans on introduce other ways of setting header values, having a new structure is more extensible and I would go for that (before we start using this one).

Any thoughts?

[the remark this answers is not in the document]
gabrieljackson: No, I really can't foresee other use-cases so I think this will work just fine. I just wanted to check in to see what trade-offs you ran into. Let's go with the simpler implementation!

cmd/cloud/events_subscription_flag.go
```diff
@@ -22,6 +27,7 @@ func (flags *subscriptionCreateFlags) addFlags(command *cobra.Command) {
 	command.Flags().StringVar(&flags.owner, "owner", "", "OwnerID of the subscription.")
 	command.Flags().StringVar(&flags.eventType, "event-type", string(model.ResourceStateChangeEventType), "Event type of the subscription.")
 	command.Flags().DurationVar(&flags.failureThreshold, "failure-threshold", 0, "Failure threshold of the subscription.")
+	command.Flags().StringToStringVar(&flags.headers, "headers", nil, "headers that should be sent with the request")
```
gabrieljackson, line 30: nit: maybe add something about `env:` values to the flag description?

cmd/cloud/webhook_flag.go
```diff
@@ -18,11 +24,13 @@ type webhookCreateFlag struct {
 	webhookFlags
 	ownerID string
 	url     string
+	headers map[string]string
 }
 
 func (flags *webhookCreateFlag) addFlags(command *cobra.Command) {
 	command.Flags().StringVar(&flags.ownerID, "owner", "", "An opaque identifier describing the owner of the webhook.")
 	command.Flags().StringVar(&flags.url, "url", "", "The callback URL of the webhook.")
+	command.Flags().StringToStringVar(&flags.headers, "headers", nil, "headers that should be sent with the request")
```
gabrieljackson, line 33: Same here

gabrieljackson (changes requested):
Just have a few comments below

mirshahriar (approved):
LGTM

internal/events/delivery.go
```diff
@@ -202,7 +202,7 @@ func (s *sender) processSubscriptionEvents(claimFunc func(instanceID string) (*m
 	log := s.logger.WithField("subscription", subscription.ID)
 	defer s.unlockSubscription(subscription.ID, log)
 
-	log.Debug("Processing events delivery for subscription")
+	s.logger.Debug("Processing events delivery for subscription")
```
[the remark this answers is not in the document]
fmartingr: I thought we were using the global logger 🤦

gabrieljackson (approved):
LGTM!

model/webhook_request.go
```diff
@@ -17,6 +17,7 @@ import (
 type CreateWebhookRequest struct {
 	OwnerID string
 	URL     string
+	Headers []WebhookHeader
```
mirshahriar, line 20:
```suggestion
	Headers Headers
```

model/types.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+		if header.Value != nil && header.ValueFromEnv != nil {
+			return fmt.Errorf("header %s cannot have both a value and a value_from_env", header.Key)
+		}
+	}
+	return nil
+}
+
+func (wh *Headers) GetHeaders() map[string]string {
+	headers := make(map[string]string, len(*wh))
```
mirshahriar, line 55: `wh` can be `nil` here. `panic`

model/types.go
```diff
@@ -0,0 +1,64 @@
...
+	ValueFromEnv *string `json:"value_from_env,omitempty"`
+}
+
+type Headers []WebhookHeader
+
+func (wh Headers) Value() (driver.Value, error) {
+	return json.Marshal(wh)
+}
+
+func (wh *Headers) Scan(databaseValue interface{}) error {
```
mirshahriar, line 26: `databaseValue` can also be `nil` here. `panic`

mirshahriar (changes requested):

model/types.go
fmartingr: This will fall down to the `default` case.

model/types.go
fmartingr: Yeah I don't know why I put a pointer in this method.

model/types.go
mirshahriar: Right, but with error. If we do not have headers, we want empty list, right? The whole DB scan will return error here.

fmartingr: Oh yeah, you're right. I added at some point a test case for that scenario and since test were passing I thought the database driver took care of that (just using `nil` from `NULL`). Adding the test case (which is mising, obviously) and handling nulls explicitly. Good catch.

mirshahriar (approved):
LGTM

PR #887, opened 2023-03-17
chore: allow custom test flags

fmartingr:
#### Summary

Slight modification to the `Makefile` so we can provide our own `TEST_FLAGS`, useful when we do not want the `-v` flag to easily see what's going on or `-count=1` to avoid test cache.

#### Ticket Link

None

#### Release Note

```release-note
chore: allow custom test flags
```


gabrieljackson (approved):
👍

mirshahriar (approved):

PR #888, opened 2023-03-17
refactor: use generics for repeating code on DTO decoding

fmartingr:
#### Summary

This replaces all `XXXXDTO(s)FromReader` with a generics version of it.

Already had this going of for some tests I did with #886, so why not just open a PR for it. If you consider it cool enough we can merge it.

#### Ticket Link

none

#### Release Note

```release-note
refactor: use generics for repeating code on DTO decoding
```


mirshahriar (approved):
Like this a lot. The use of generics here is excellent.

PR #889, opened 2023-03-18
[MM-51509] Upgrade helm version to support --wait flag

mirshahriar:
#### Summary

We added the `--wait` flag in `helm uninstall`.  But this flag is only available after `v3.7.0`.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51509

#### Release Note

```release-note
None
```


fmartingr (approved):
This seems ok to me but I don't know if we are pinned to that version for any reason. Adding sre to review this as well.

PR #893, opened 2023-03-20
Fix workflow not to run CD on PR

mirshahriar:
#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51533

#### Release Note

```release-note
None
```


.github/workflows/cd.yml
```diff
@@ -1,13 +1,11 @@
...
 on:
-  workflow_run:
-    workflows: ["ci"]
-    branches-ignore: [master]
-    types:
-      - completed
-  push:
-    tags:
-      - "v[0-9]+.[0-9]+.[0-9]+"
+  workflow_call:
```
spirosoik, line 4: you still need the `workflow_run` but instead for `branches-ignore` add `["*"]`.

mirshahriar: `workflow_run` always works on default branch. So It will `build` & `push` for default branch.

.github/workflows/notify-release.yml
fmartingr, line 1: Are we replacing this with something else?

.github/workflows/ci.yml
```diff
@@ -75,44 +76,26 @@ jobs:
...
-          image-ref: "mattermost/mattermost-cloud"
-          format: "table"
-          ignore-unfixed: true
-          vuln-type: "os,library"
-          severity: "CRITICAL"
-
-      - name: ci/build-docker-e2e
-        run: make build-image-e2e
+  master:
+    if: ${{ github.ref_name  == 'master' }}
```
fmartingr, line 88: To make this easily reusable across projects, can we set up `master` as a variable? You can put :

```
env:
  DEFAULT_BRANCH: master
```

at the top of the file and then use is as `env.DEFAULT_BRANCH`. 

This will make easier move this same flows across projects :)

[doc ref](https://docs.github.com/en/github-ae@latest/actions/learn-github-actions/variables)

.github/workflows/notify-release.yml
mirshahriar: I should not delete this. mistake.

spirosoik:
@phoinixgrr @toninis have a look here, let's follow what we discussed with the design doc of CI Forks too.

toninis (changes requested):
I would use a different approach . Check [mattermost-plugin-playbooks](https://github.com/mattermost/mattermost-plugin-playbooks) repo . 

On `ci.yaml` I would use :
```yaml
on:
  pull_request:
  push:
    branches:
      - master
```
and on `cd.yaml` 
```yaml
on:
  push:
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+"
  workflow_run:
    workflows:
      - ci
    types:
      - completed
    branches:
      - master
```

You do not have to complicate things with composite actions here . We are rolling out a documentation in order to make the transition smoother . 

WIth this way you limit the conditions and reduce complexity . If you need any help ping us .

.github/workflows/ci.yml
```diff
@@ -1,29 +1,30 @@
...
+      - name: ci/checkout-repo
+        uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
         with:
           fetch-depth: 0
 
       - name: Setup Go
         uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
         with:
-          go-version: "1.19"
+          go-version: "${{ env.GO_VERSION }}"
```
fmartingr, line 27: If we're not doing a matrix, I'd say go with `go-version-file: go.mod`

PR #896, opened 2023-03-27
Add catalog-info.yaml for Backstage

spirosoik:
#### Summary
Registering provisioner to backstage portal

#### Release Note
```release-note
NONE
```


fmartingr (approved):
Approving though there's not much for us to review here.

One question though:`owner` should be `cloud` or `cloud/platform` ? Are we using Cloud directly without sub teams in backstage?

gabrieljackson (approved):

spirosoik:
@fmartingr you are right. We are in process to create a team for platform as right now doesn't exist.

PR #897, opened 2023-03-27
[MM-51737] Cluster Installation Status endpoint

mirshahriar:
#### Summary

This endpoint `/api/cluster_installation/%s/status` will return the status of running the mm servers.

```
type ClusterInstallationStatus struct {
	InstallationFound bool   `json:"installation_found"`
	Replicas          *int32 `json:"replicas,omitempty"`
	TotalPod          *int32 `json:"total_pod,omitempty"`
	RunningPod        *int32 `json:"running_pod,omitempty"`
	ReadyPod          *int32 `json:"ready_pod,omitempty"`
	StartedPod        *int32 `json:"started_pod,omitempty"`
	ReadyLocalServer  *int32 `json:"ready_local_server,omitempty"`
}
```

The `cloud` command `cloud cluster installation status --cluster-installation=` will fetch from that endpoint.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51737

#### Release Note

```release-note
None
```

### DIFF
--- a/e2e/workflow/installation.go
+++ b/e2e/workflow/installation.go
@@ -117,6 +117,10 @@ func (w *InstallationSuite) GetCI(ctx context.Context) error {
 	return nil
 }
 
+func (w *InstallationSuite) CheckClusterInstallationStatus(ctx context.Context) error {
+	return pkg.WaitForClusterInstallationReadyStatus(w.client, w.Meta.ClusterInstallationID, w.logger)
+}
+
 // GetConnectionStrAndExport saves db connection string currently configured for installation and export statistics.
 func (w *InstallationSuite) GetConnectionStrAndExport(ctx context.Context) error {
 	connectionString, err := pkg.GetConnectionString(w.client, w.Meta.ClusterInstallationID)
--- a/e2e/workflow/installation.go
+++ b/e2e/workflow/installation.go
@@ -117,6 +117,7 @@ func (w *InstallationSuite) GetCI(ctx context.Context) error {
 	return nil
 }
 
+// CheckClusterInstallationStatus checks if ClusterInstallation is in a good condition.
 func (w *InstallationSuite) CheckClusterInstallationStatus(ctx context.Context) error {
 	return pkg.WaitForClusterInstallationReadyStatus(w.client, w.Meta.ClusterInstallationID, w.logger)
 }
--- a/internal/provisioner/cluster_installation_provisioner.go
+++ b/internal/provisioner/cluster_installation_provisioner.go
@@ -538,7 +538,6 @@ func (provisioner Provisioner) GetClusterInstallationStatus(cluster *model.Clust
 			return nil, errors.Wrap(err, "failed to query mattermost deployment")
 		}
 
-		status.InstallationFound = false
 		return &status, nil
 	}
 
--- a/internal/provisioner/cluster_installation_provisioner.go
+++ b/internal/provisioner/cluster_installation_provisioner.go
@@ -538,11 +538,11 @@ func (provisioner Provisioner) GetClusterInstallationStatus(cluster *model.Clust
 			return nil, errors.Wrap(err, "failed to query mattermost deployment")
 		}
 
-		status.InstallationFound = ptr.Bool(false)
+		status.InstallationFound = false
 		return &status, nil
 	}
 
-	status.InstallationFound = ptr.Bool(true)
+	status.InstallationFound = true
 	status.Replicas = deployment.Spec.Replicas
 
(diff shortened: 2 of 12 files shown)


model/cluster_installation.go
```diff
@@ -33,6 +35,16 @@ type ClusterInstallationFilter struct {
...
 
+type ClusterInstallationStatus struct {
+	InstallationFound bool   `json:"installation_found"`
+	Replicas          *int32 `json:"replicas,omitempty"`
+	TotalPod          *int32 `json:"total_pod,omitempty"`
+	RunningPod        *int32 `json:"running_pod,omitempty"`
+	ReadyPod          *int32 `json:"ready_pod,omitempty"`
+	StartedPod        *int32 `json:"started_pod,omitempty"`
+	ReadyLocalServer  *int32 `json:"ready_local_server,omitempty"`
+}
```
gabrieljackson, line 46: Historically, we have used the non-underline JSON naming scheme for our various structs: https://github.com/mattermost/mattermost-cloud/blob/master/model/group_status.go#L12-L20

This wasn't really an intentional decision, but sort of became the standard for the codebase. Do you think we should change that going forward or keep the original style. @fmartingr have any thoughts as well?

mirshahriar: Let's stick with the initial design. If we want to change, we also need to consider how to handle DB objects.

e2e/workflow/installation.go
```diff
@@ -117,6 +117,10 @@ func (w *InstallationSuite) GetCI(ctx context.Context) error {
 	return nil
 }
 
+func (w *InstallationSuite) CheckClusterInstallationStatus(ctx context.Context) error {
+	return pkg.WaitForClusterInstallationReadyStatus(w.client, w.Meta.ClusterInstallationID, w.logger)
```
fmartingr, line 121: Maybe we can merge `GetCI` and `CheckClusterInstallationStatus` into the  same step?

internal/provisioner/cluster_installation_provisioner.go
```diff
@@ -522,6 +524,86 @@ func (provisioner Provisioner) getMattermostCustomResource(cluster *model.Cluste
...
+
+	var status model.ClusterInstallationStatus
+
+	deployment, err := k8sClient.Clientset.AppsV1().Deployments(clusterInstallation.Namespace).Get(context.TODO(), makeClusterInstallationName(clusterInstallation), metav1.GetOptions{})
+	if err != nil {
+		if !k8sErrors.IsNotFound(err) {
+			return nil, errors.Wrap(err, "failed to query mattermost deployment")
+		}
+
+		status.InstallationFound = false
```
fmartingr, line 541: This should default to `false`.

model/cluster_installation.go
fmartingr: My personal preference is `snake_case` for APIs, but I'd rather stick with `CamelCase` in the provisioner than switch everything to snakes, I'd rather be homogeneus than picky about this. 1/5

model/cluster_installation.go
```diff
@@ -124,3 +136,13 @@ func MigrateClusterInstallationResponseFromReader(reader io.Reader) (*MigrateClu
 
 	return &migrateClusterInstallationResponse, nil
 }
+
+func NewClusterInstallationStatusFromReader(reader io.Reader) (*ClusterInstallationStatus, error) {
```
fmartingr, line 140: Shouldn't we get a lint error when a public function does not have a comment? cc: @gabrieljackson

fmartingr (approved):
Looks good to me. Ingenius way to check the server status by the way.

While I agree this make our e2e more robust, I'd say we may have found a _something_ in either the operator or the server. We should trust that our ready pods are truly ready, so if the server is reporting that is ready when it could not be the case this could lead to other problems in the future. So the end goal here should be not requiring to manually check if the server is properly running since it should report that itself correctly.

e2e/workflow/installation.go
mirshahriar: In that case, do we need a better name than `GetCI`?

model/cluster_installation.go
mirshahriar: We need to configure that in linters. Looks like it's missing.

gabrieljackson: Huh, that used to be working. We must have lost it along the way somehow.

gabrieljackson (approved):
👍

e2e/workflow/installation.go
mirshahriar: I'm leaving it now as it it.

PR #898, opened 2023-03-28
feat: failing doers don't interrupt other doers

fmartingr:
#### Summary

Currently the `MultiDoer` stucks the cluster if some of the enabled `Doer`s does not work or has a error. This change allows the continued operation of all `Doer`s even if some of them fail in any way, to prevent the provisioner to get stuck.

**Note:** This was like that **by design**, so I'll leave this as a Draft for now until we know or decide how to move forward.

#### Ticket Link

None

#### Release Note

```release-note
feat: failing doers don't interrupt other doers
```


mirshahriar (approved):
LGTM. One question. Is there any effect of a failed doer on the subsequent doers if we continue?

fmartingr:
> LGTM. One question. Is there any effect of a failed doer on the subsequent doers if we continue?

Depends on the doer that is failing and the current reconciliation phase of the installation being worked on I guess. If you trigger a backup and the backup doer is not working properly (awat down), the installation would be scaled down to zero and left that way. But we don't have failover either way. At least handling the doers independently an outage would be partial, since we could still create new installations since the installation doer would still work properly. Mostly a problem locally when enabling all the doers, but trying to minimize any problem in production as well.

My main concern about this is that it was designed to be the other way around (even with tests) and we don't know why, actually. That's the only thing that refrain me from moving this forward because I may have missed something.

fmartingr:
After discussing it with @gabrieljackson we reached the conclusion that we are going to keep this as it was for two main reasons:
- Since this was done it by design and we cannot figure out why at the moment, modifying it may create unnecessary/unknown problems; 
- while doer failures are a mild inconvenience while working locally, a doer failing in production should not happen and should be monitored and mitigated either way. Not only that, but a doer failing that is a dependency for others may block them either way

Closing.

PR #903, opened 2023-04-09
Fix invalid Grafana link template

mirshahriar:
#### Summary

Grafana link is not working for TestID. 

#### Release Note

```release-note
None
```


fmartingr (approved):
Good catch

PR #910, opened 2023-04-25
[MM-51736] Delete Nodegroup from an EKS cluster

mirshahriar:
#### Summary

```bash
$ cloud cluster nodegroup delete --cluster=mw3xfmepaffefbuwees6ynbqjo --nodegroup=calls
```

This command will delete an existing Nodegroup from an existing EKS cluster.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51736

#### Release Note

```release-note
None
```


internal/supervisor/cluster.go
```diff
@@ -33,8 +33,9 @@ type ClusterProvisioner interface {
 	PrepareCluster(cluster *model.Cluster) bool
 	CreateCluster(cluster *model.Cluster) error
 	CheckClusterCreated(cluster *model.Cluster) (bool, error)
-	CreateNodes(cluster *model.Cluster) error
-	CheckNodesCreated(cluster *model.Cluster) (bool, error)
+	CreateNodegroups(cluster *model.Cluster) error
+	CheckNodegroupsCreated(cluster *model.Cluster) (bool, error)
+	DeleteNodegroup(cluster *model.Cluster) error
```
fmartingr, line 38: My only nitpick here is that if this can remove more than one nodegroup it should be plural, even if the CLI only allows for one at a time?

fmartingr (approved):

internal/supervisor/cluster.go
mirshahriar: Right. Thanks

internal/provisioner/eks_provisioner.go
```diff
@@ -343,6 +343,71 @@ func (provisioner *EKSProvisioner) CheckNodesCreated(cluster *model.Cluster) (bo
 	return true, nil
 }
 
+// DeleteNodegroups deletes the EKS nodegroup.
+func (provisioner *EKSProvisioner) DeleteNodegroups(cluster *model.Cluster) error {
+	logger := provisioner.logger.WithField("cluster", cluster.ID)
+
+	eksMetadata := cluster.ProvisionerMetadataEKS
+	if eksMetadata == nil {
+		return errors.New("error: EKS metadata not set when deleting EKS NodeGroup")
```
gabrieljackson, line 352: nit: we don't normally prepend returned errors with `error:`

internal/provisioner/eks_provisioner.go
```diff
@@ -343,6 +343,71 @@ func (provisioner *EKSProvisioner) CheckNodesCreated(cluster *model.Cluster) (bo
...
+	}
+
+	changeRequest := eksMetadata.ChangeRequest
+	if changeRequest == nil || changeRequest.NodeGroups == nil {
+		return errors.New("error: EKS NodeGroup change request not set when deleting EKS NodeGroup")
```
gabrieljackson, line 357: Same here

internal/supervisor/cluster.go
```diff
@@ -264,6 +267,17 @@ func (s *ClusterSupervisor) createNodegroups(cluster *model.Cluster, logger log.
 	return s.refreshClusterMetadata(cluster, logger)
 }
 
+func (s *ClusterSupervisor) deleteNodegroup(cluster *model.Cluster, logger log.FieldLogger) string {
```
gabrieljackson, line 270:
```suggestion
func (s *ClusterSupervisor) deleteNodegroups(cluster *model.Cluster, logger log.FieldLogger) string {
```

gabrieljackson: There's a few other logs and function names around here that use mixed naming of nodegroups and nodegroup. I would just use nodegroups for everything as it looks like this would cleanup multiple nodegroups if requested.

model/cluster_states.go
```diff
@@ -36,6 +36,10 @@ const (
 	ClusterStateNodegroupsCreationRequested = "nodegroups-creation-requested"
 	// ClusterStateNodegroupsCreationFailed is a cluster that failed to create nodegroups.
 	ClusterStateNodegroupsCreationFailed = "nodegroups-creation-failed"
+	// ClusterStateNodegroupDeletionRequested is a cluster in the process of deleting nodegroup.
+	ClusterStateNodegroupDeletionRequested = "nodegroup-deletion-requested"
+	// ClusterStateNodegroupDeletionFailed is a cluster that failed to delete nodegroup.
+	ClusterStateNodegroupDeletionFailed = "nodegroup-deletion-failed"
```
gabrieljackson, line 42: Same here with the states. I think these should also be `nodegroups`?

gabrieljackson (commented):
Just marked a few name mismatches for `nodegroup` vs `nodegroups` that we should probably cleanup first.

mirshahriar:
@gabrieljackson, In next PR, I'll fix logs/error messages. Fixing those here will cause several unrelated changes.

gabrieljackson (approved):
👍

PR #915, opened 2023-04-30
[MM-52578] Resolved panic for missing VPC

mirshahriar:
#### Summary

If the VPC is claimed by the secondary cluster, deployment of the NGINX utility causes a panic.
We need to get the VPC either claimed by primary or secondary cluster.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-52578

#### Release Note

```release-note
None
```


gabrieljackson (approved):
👍

PR #917, opened 2023-05-03
Use VPC CNI as Networking

mirshahriar:
#### Summary

We are only supporting AWS VPC CNI for now.

#### Release Note

```release-note
None
```

### DIFF
--- a/model/cluster_request.go
+++ b/model/cluster_request.go
@@ -17,6 +17,8 @@ const (
 	NetworkingCalico = "calico"
 	// NetworkingAmazon is Amazon networking plugin.
 	NetworkingAmazon = "amazon-vpc-routed-eni"
+	// NetworkingVpcCni is Amazon VPC CNI networking plugin.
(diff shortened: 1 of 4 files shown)


fmartingr (approved):

model/cluster_request.go
```diff
@@ -17,6 +17,8 @@ const (
 	NetworkingCalico = "calico"
 	// NetworkingAmazon is Amazon networking plugin.
 	NetworkingAmazon = "amazon-vpc-routed-eni"
+	// NetworkingVpcCni is Amazon VPC CNI networking plugin.
+	NetworkingVpcCni = "vpc-cni"
```
gabrieljackson, line 21:
```suggestion
	NetworkingVpcCni = "amazon-vpc-cni"
```

I think this would be a bit clearer for anyone reviewing cluster configuration. Just want to be sure that there is no confusion about the nature of the CNI.

gabrieljackson (approved):
LGTM!

PR #918, opened 2023-05-03
[MM-51246] fix: recover old utility version parsing fallback

fmartingr:
#### Summary

Recover code to parse old utility version parsing for old clusters included in environment databases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51246

#### Release Note

```release-note
fix: recover old utility version parsing fallback
```


model/cluster_utility.go
```diff
@@ -169,6 +172,76 @@ func (h *UtilityGroupVersions) AsMap() map[string]*HelmUtilityVersion {
 	}
 }
 
+// UnmarshalJSON is a custom JSON unmarshaler that can handle both the
+// old Version string type and the new type. It is entirely
+// self-contained, including types, so that it can be easily removed
+// when no more clusters exist with the old version format.
+// TODO DELETE THIS
+func (h *UtilityGroupVersions) UnmarshalJSON(bytes []byte) error {
+	type innerUtilityGroupVersions struct {
```
mirshahriar, line 181: I think we do not need this new `struct`. We can use `UtilityGroupVersions` here.

model/cluster_utility.go
```diff
@@ -169,6 +172,76 @@ func (h *UtilityGroupVersions) AsMap() map[string]*HelmUtilityVersion {
...
+		Rtcd                *HelmUtilityVersion
+		NodeProblemDetector *HelmUtilityVersion
+		MetricsServer       *HelmUtilityVersion
+		Velero              *HelmUtilityVersion
+		Cloudprober         *HelmUtilityVersion
+	}
+
+	var versions *innerUtilityGroupVersions = &innerUtilityGroupVersions{}
+
+	err := json.Unmarshal(bytes, versions)
```
mirshahriar, line 199: Can we Unmarshal directly into the receiver `h`? In that case, If `err == nil`, we do not need to assign these values into  the receiver.

model/cluster_utility.go
```diff
@@ -169,6 +172,76 @@ func (h *UtilityGroupVersions) AsMap() map[string]*HelmUtilityVersion {
...
+	if err != nil && strings.Contains(err.Error(), "cannot unmarshal string into Go struct field") {
```
mirshahriar, line 200: Just a thought. If we do not need to know the version of the utilities of the deleted cluster, we can simply ignore the error and return `nil`.

fmartingr: I'm ok with this, but previous to the discussion I'd rather submit the fix in case we want the data for some strange reason.

model/cluster_utility.go
fmartingr: We can use `h` directly because we'll end in a stack overflow, since we will end up calling this method each time we call `json.Unmarshal`.

mirshahriar: Right. My bad. Can we try this?

```go
type innerUtilityGroupVersions UtilityGroupVersions
var versions = &innerUtilityGroupVersions{}
```

And to assign this `versions` into `h`, we can use `*h = *(*UtilityGroupVersions)(versions)`

mirshahriar (approved):
LGTM

fmartingr:
The alternatives for this are:
- Creating a migration that will convert from the old format to the new one or even remove it if we don't want it anymore.
- Clean the database from unnecessary rows
- Ignore the error and just move along, since this is just for old metadata.
- Parse it the way it was working before (this PR).

gabrieljackson (approved):
LGTM. I think this makes a lot of sense compared to the other options.

PR #920, opened 2023-05-05
[MM-52658] fix: raised e2e installation timeout to 15m

fmartingr:
#### Summary

With the change of the E2E tests to single tentant postgresql databases, E2E were failing because the current timeout of 5 minutes is too low. In my tests it usually take around 8-9 minutes now, so I've raised the timeout to 15m just to be sure, on in case we use other database engines that take longer as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52658

#### Release Note

```release-note
test(e2e): raised installation timeout to 15m
```

### DIFF
--- a/e2e/pkg/installation.go
+++ b/e2e/pkg/installation.go
@@ -20,6 +20,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// installationWaitTimeout is the maximum time to wait for an installation to be ready.
+const installationWaitTimeout = 15 * time.Minute
+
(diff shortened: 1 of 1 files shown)


gabrieljackson (approved):

mirshahriar (approved):
LGTM. Only one question. As this is taking a long time, can we increase retry interval to 30s?

mirshahriar:
> Can we increase retry interval to 30s?

This is only used to wake up hibernated installation. This is okay. Ignore my comment

PR #921, opened 2023-05-08
[MM-51707] Drop SQLite support

fmartingr:
#### Summary

- Remove SQLite relevant logic from the code and migrations
- Added QoL to allow running unittests using PostgreSQL directly (using docker)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51707

#### Release Note

```release-note
feat: drop sqlite support
```


internal/store/migrations.go
```diff
@@ -84,58 +84,6 @@ var migrations = []migration{
 			if err != nil {
 				return err
 			}
-		} else if e.DriverName() == driverSqlite {
```
gabrieljackson, line 87: I believe we should leave the migrations as they were for now even though it's highly unlikely it will matter. My thinking here is if we are reviewing master code to look for an issue that may have been introduced in migrations that we should have the full history even for SQLite just in case.

It looks like we are keeping the `driverSqlite` variable for now so that should be okay.

Same for the others below. If you ran into something that led to wanting to remove these just let me know and we can discuss.

internal/store/migrations.go
```diff
@@ -2146,6 +1577,14 @@ var migrations = []migration{
 			return err
 		}
 
+		return nil
+	}},
+	{semver.MustParse("0.42.0"), semver.MustParse("0.43.0"), func(e execer) error {
+		// No-op migration to ensure SQLite is no longer used.
+		if e.DriverName() == driverSqlite {
+			panic("SQLite is no longer supported as a database backend")
```
gabrieljackson, line 1585: I never thought I would see the day! 🎉

gabrieljackson (commented):
Very nice work. Please look over my migration-specific request. LGTM other than that.

gabrieljackson (approved):
This is great! Thanks

PR #922, opened 2023-05-09
Set logger output to os.Stdout

mirshahriar:
#### Summary

```go
logger = log.New()
```

We use this `logger` in our code. So, to set output to stdout, we need to set it in `logger`

```
logger.SetOutput(os.Stdout)
```

#### Release Note

```release-note
Set logger output to os.Stdout
```

### DIFF
--- a/cmd/cloud/logger.go
+++ b/cmd/cloud/logger.go
@@ -21,7 +21,7 @@ func init() {
 		FullTimestamp: true,
 	})
 	// Output to stdout instead of the default stderr.
-	log.SetOutput(os.Stdout)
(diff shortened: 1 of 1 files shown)


fmartingr (approved):
This seems natural to me. I'm adding SRE as reviewers because the default output for logrus is `stderr`, and while I don't think there could be any problem with the log collecting I'd rather make sure.

mirshahriar:
> the default output for logrus is `stderr`.

Right. I have changed this based on this comment

```go
// Output to stdout instead of the default stderr.
log.SetOutput(os.Stdout)
```

PR #927, opened 2023-05-26
Upgrade k8s rotator dependency

gabrieljackson:
```release-note
None
```


stafot:
@gabrieljackson had similar lint issues in rotator you can check how we refactored these there

gabrieljackson:
> @gabrieljackson had similar lint issues in rotator you can check how we refactored these there

Yep thanks, I wanted to keep our versioning mostly aligned with the operator so I opted to use some go mod replace statements for now to do this instead of updating the code. We will remove these later when we are able to update the operator decencies as well.

stafot (approved):
LGTM

nickmisasi (approved):

PR #931, opened 2023-06-06
Add installation deletion report command

gabrieljackson:
This new command is used to summarize installation deletion times that are upcoming for all installations. Installations are counted in configurable groups with the cutoff time being any time in the future.

Fixes https://mattermost.atlassian.net/browse/CLD-5779

```release-note
Add installation deletion report command
```

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -7,11 +7,13 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


fmartingr (approved):
LGTM. Though I didn't dive into this `tablewritter` thing. 
Unsure how difficult would've been to just use `Printf` with fixed widths.

fmartingr (approved):
LGTM. Though I didn't dive into this `tablewritter` thing. 
Unsure how difficult would've been to just use `Printf` with fixed widths.

Double clicking is the best.

nickmisasi (approved):

PR #932, opened 2023-06-08
Onboarding adjustments

nickmisasi:
#### Summary
Having gone through the onboarding I've made a few adjustments to the README for first timers. This PR also includes a docker-compose file to spin up a postgres database, since the recent deprecation of sqlite requires a postgres instance of some kind. The postgres db can be spun up with either `docker-compose up` (optionally passing `-d`), or with `make dev-start`. The postgres db will create the `cloud` database on first start, so all that's required is to run the schema migrations against it.

#### Ticket Link
N/a

#### Release Note

```release-note
None
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,26 @@
+version: "3.7"
+services:
+  postgres:
+    image: "postgres:12.5"
+    container_name: "cloud-postgres"
+    restart: on-failure
+    networks:
+    - cloud
+    ports:
+    - "5430:5432"
+    environment:
+      POSTGRES_USER: provisioner
+      POSTGRES_PASSWORD: provisionerdev
+    volumes:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: "3.7"
+services:
+  postgres:
+    image: "postgres:12.5"
+    container_name: "cloud-postgres"
+    restart: on-failure
+    networks:
+    - cloud
+    ports:
+    - "5430:5432"
```
nickmisasi, line 10: I think Gabe uses 5433 here, but that's something that is taken by CWS for me so I've gone for 5430 instead. 0/5 on what we want here.

gabrieljackson: Yep, that totally works. 👍

gabrieljackson (approved):
LGTM!

PR #934, opened 2023-06-09
[CLD-5774] Update S3 bucket cleanup logic to handle versioning

nickmisasi:
#### Summary
Adds logic to S3 storage cleanup that will suspend versioning, then iterate through all versions and delete them before then deleting all objects and the bucket itself. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-5774
#### Release Note

```release-note
Updated S3 bucket cleanup logic to handle versioned objects
```

### DIFF
--- a/internal/tools/aws/s3.go
+++ b/internal/tools/aws/s3.go
@@ -79,6 +79,59 @@ func (a *Client) s3EnsureBucketCreated(bucketName string, logger log.FieldLogger
 	return nil
 }
 
+func (a *Client) S3BatchDeleteVersions(bucketName string, prefix *string) error {
+	ctx := context.TODO()
+
+	versionsPaginator := s3.NewListObjectVersionsPaginator(
+		a.service.s3,
+		&s3.ListObjectVersionsInput{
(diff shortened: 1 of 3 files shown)


internal/tools/aws/s3.go
```diff
@@ -150,6 +234,25 @@ func (a *Client) S3EnsureBucketDeleted(bucketName string, logger log.FieldLogger
 		logger.WithField("s3-bucket-name", bucketName).WithError(err).Warn("Could not determine if S3 bucket exists")
 	}
 
+	isVersioned, err := a.S3IsVersioningEnabled(bucketName)
+
+	if err != nil {
+		logger.WithField("s3-bucket-name", bucketName).WithError(err).Warn("Could not determine if S3 bucket is versioned")
+		// TODO: Continue anyways?
+		return err
+	}
```
nickmisasi, line 243: @gabrieljackson thoughts here? Should we just assume not versioned in this case?

stafot (approved):

gabrieljackson (approved):
LGTM!

PR #936, opened 2023-06-13
Update notify-release.yml

stafot:
#### Summary
Current checkout is out of date and uses deprecated node12

#### Release Note

```release-note
Current checkout is out of date and uses deprecated node12
```


gabrieljackson (approved):
👍

PR #940, opened 2023-06-20
Fix CI/CD mock check

gabrieljackson:
This change separates out the mock check to use Ubuntu 20.04 until 22.04 is fixed.

Fixes https://mattermost.atlassian.net/browse/CLD-5865

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,8 @@ unittest:
 .PHONY: verify-mocks
 verify-mocks: mocks
 	@if !(git diff --quiet HEAD); then \
+		git status \
+		git diff \
 		echo "generated files are out of date, run make mocks"; exit 1; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -234,6 +234,8 @@ unittest:
 .PHONY: verify-mocks
 verify-mocks: mocks
 	@if !(git diff --quiet HEAD); then \
+		git status \
+		git diff \
```
nickmisasi, line 238: Can remove these now?
```suggestion
```

nickmisasi (approved):

Makefile
gabrieljackson: I think they are useful so we can see what files are not matching on future failures

PR #942, opened 2023-07-10
[CLD-5850] Refactor E2E tests to be more flexible

nickmisasi:
#### Summary
This PR refactors the existing E2E tests to allow them to be more flexible. 

I've shifted a bunch of logic that was previously isolated to the cluster suite into a "shared" package that can be leveraged by others. I have introduced an installation specific package in the e2e tests for lifecycle testing more specific to that of installations (like creating and deleting an installation with a versioned AWS S3 bucket, for example)

The installation tests leverage the idempotency of the workflow library, and will re-use a cluster that was created by a previous step if there is one (creating one otherwise). This allows the installation suite to run quicker, as each step doesn't require a cluster to be created and destroyed. With this refactor, adding new cases to the installation suite is as simple as creating a new function that returns a list of steps, and adding it to the case list in the installation_test.go file.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-5850

#### Release Note

```release-note
None
```

### DIFF
--- a/e2e/tests/shared/shared.go
+++ b/e2e/tests/shared/shared.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+//
+
 //go:build e2e
 // +build e2e
 
--- a/Makefile
+++ b/Makefile
@@ -262,11 +262,14 @@ e2e-cluster:
 	@echo Starting cluster e2e test.
 	go test ./e2e/tests/cluster -tags=e2e -v -timeout 90m
 
-
-.PHONE: e2e-installation
+.PHONY: e2e-installation
 e2e-installation:
 	@echo Starting installation e2e test
 	go test ./e2e/tests/installation -tags=e2e -v -timeout 90m
+
+
+.PHONY: e2e
+e2e: e2e-cluster e2e-installation
 
 
 ## --------------------------------------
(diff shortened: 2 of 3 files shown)


e2e/tests/shared/shared.go
```diff
@@ -1,14 +1,17 @@
-// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
-// See LICENSE.txt for license information.
-//
-
```
gabrieljackson, line 4: Probably want to leave this intact

e2e/workflow/installation.go
```diff
@@ -106,6 +106,36 @@ func (w *InstallationSuite) CreateInstallation(ctx context.Context) error {
...
+		DB(w.Params.DBType).
+		FileStore(model.InstallationFilestoreAwsS3).
+		Annotations(w.Params.Annotations)
+
+	installation, err := w.client.CreateInstallation(installationBuilder.CreateRequest())
+	if err != nil {
+		return errors.Wrap(err, "while creating installation")
+	}
+	w.logger.Infof("Installation created: %s", installation.ID)
+	w.Meta.InstallationID = installation.ID
```
gabrieljackson, line 122: One side effect of this is the test webhook that is sent after the run will probably only contain the log link for this installation and not the first installation in the suite. This isn't a big deal. I want to refactor the webhooks anyway so just noting it here if we end up using the log links after this is merged.

gabrieljackson (approved):
LGTM! Just had two non-blocking comments that can be addressed later.

gabrieljackson (approved):
👍

PR #943, opened 2023-07-12
[MM-5945] Add support for installation deletion locks

nickmisasi:
#### Summary
Adds support for deletion locks, which will prevent an installation from being deleted if the lock is set. If the lock is set, the lock must be unset before deletion can occur.  Deletion locks are _not_ enabled by default and must be done so explicitly. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-5945

#### Release Note

```release-note
Support for "deletion lock", preventing an installation from being deleted if set
```

### DIFF
--- a/internal/api/security.go
+++ b/internal/api/security.go
@@ -25,6 +25,8 @@ func initSecurity(apiRouter *mux.Router, context *Context) {
 	securityInstallationRouter := securityRouter.PathPrefix("/installation/{installation:[A-Za-z0-9]{26}}").Subrouter()
 	securityInstallationRouter.Handle("/api/lock", addContext(handleInstallationLockAPI)).Methods("POST")
 	securityInstallationRouter.Handle("/api/unlock", addContext(handleInstallationUnlockAPI)).Methods("POST")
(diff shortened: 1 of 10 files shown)


nickmisasi:
@gabrieljackson I figured it made sense to just add this as an additional feature under the `cloud security installation deletion-lock/unlock` umbrella, since it's fairly similar to the API Lock. lmk if you'd like a more dedicated command for this

internal/api/security.go
```diff
@@ -159,6 +161,68 @@ func handleInstallationUnlockAPI(c *Context, w http.ResponseWriter, r *http.Requ
...
+		c.Logger.WithError(err).Error("failed to query installation")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if installation == nil {
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	// TODO: Should this respect the APISecurityLock?
```
nickmisasi, line 182: Discussed in 1:1 - we can skip the API Security Lock

gabrieljackson (approved):
Looks good!

PR #951, opened 2023-07-28
Implement NGINX IP Filtering

stafot:
#### Summary
Implement NGINX IP Filtering

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-5916

#### Release Note

```release-note
Implement NGINX IP Filtering
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 ## Tool Versions
 GOLANG_VERSION := $(shell cat go.mod | grep "^go " | cut -d " " -f 2)
 ALPINE_VERSION = 3.18.2
-TERRAFORM_VERSION=1.0.7
+TERRAFORM_VERSION=1.5.5
 KOPS_VERSION=v1.24.5
 HELM_VERSION=v3.11.2
 KUBECTL_VERSION=v1.24.4
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The following is required to properly run the cloud server.
 ##### Note: when versions are specified, it is extremely important to follow the requirement. Newer versions will often not work as expected
 
 1. Install [Go](https://golang.org/doc/install)
-2. Install [Terraform](https://learn.hashicorp.com/terraform/getting-started/install.html) version v1.0.7
+2. Install [Terraform](https://learn.hashicorp.com/terraform/getting-started/install.html) version v1.5.5
    1. Try using [tfswitch](https://warrensbox.github.io/terraform-switcher/) for switching easily between versions
 3. Install [kops](https://github.com/kubernetes/kops/blob/master/docs/install.md) version 1.24.X
 4. Install [Helm](https://helm.sh/docs/intro/install/) version 3.11.X


stafot:
@nickmisasi Did your proposed changes please review again when you have some time.

internal/store/installation.go.orig
```diff
@@ -28,9 +28,9 @@ func init() {
 			"MattermostEnvRaw", "PriorityEnvRaw", "SingleTenantDatabaseConfigRaw", "ExternalDatabaseConfigRaw",
 			"Installation.CreateAt", "Installation.DeleteAt", "Installation.DeletionPendingExpiry",
 <<<<<<< HEAD
-			"APISecurityLock", "LockAcquiredBy", "LockAcquiredAt", "CRVersion", "Installation.DeletionLocked", "AllowedRanges",
+			"APISecurityLock", "LockAcquiredBy", "LockAcquiredAt", "CRVersion", "Installation.DeletionLocked", "AllowedIPRanges",
```
nickmisasi, line 31: Accidentally added this I think?

nickmisasi (approved):
Changes LGTM. One file that I think needs to be removed. 

I think it's worth still considering moving to an array of strings rather than parsing out a string csv everywhere, but what's here is definitely functional enough as it is

internal/store/installation.go.orig
stafot: @nickmisasi for the array of strings, needed an external library dependancy and in my humble opinion the benefits were low to add this extra complexity, but sure I am not strongly opinionated on this.

model/installation_request.go
```diff
@@ -433,6 +458,25 @@ func (p *PatchInstallationDeletionRequest) Apply(installation *Installation) boo
...
 
+func (p *PatchInstallationRequest) addMissingIngressSourceRanges(installation *Installation) string {
+	var ips []string
+	if p.AllowedIPRanges != nil {
+		ips = strings.Split(*p.AllowedIPRanges, ",")
+	}
+
+	allowedIPRanges := strings.Split(installation.AllowedIPRanges, ",")
+	for _, ip := range ips {
+		if !contains(allowedIPRanges, ip) {
```
stafot, line 469: Need to fix this too

internal/supervisor/cluster_installation.go
```diff
@@ -226,6 +226,7 @@ func (s *ClusterInstallationSupervisor) createClusterInstallation(clusterInstall
 	}
 
 	err = s.provisioner.CreateClusterInstallation(cluster, installation, dnsRecords, clusterInstallation)
+	logger.Debug(installation.AllowedIPRanges)
```
stafot, line 229: and this

stafot:
gentle reminder to @mattermost/cloud-sre and @gabrieljackson

model/installation_request.go
```diff
@@ -369,6 +380,20 @@ func (p *PatchInstallationRequest) Apply(installation *Installation) bool {
 		applied = true
 		installation.License = *p.License
 	}
+
+	if p.OverrideIPRanges != nil && *p.OverrideIPRanges != installation.OverrideIPRanges {
+		applied = true
+		installation.OverrideIPRanges = *p.OverrideIPRanges
```
gabrieljackson, line 386: I am wondering if we need to store the override bool on the installation itself. It makes sense that we need to be able to pass in an override bool on the patch, but then it seems like patch would just force the new IP range value on the installation and then it could be disregarded.

Example:
1. Set allowed IP range to `1.1.1.1/24`, store on installation and apply to ingress
2. Sometime later, patch installation to IP range `2.2.2.2/24` and use force override. At that time, store the new installation range and apply to ingress.

I see that we use the override in a few other places, but unless I am missing something this would be a simpler architecture. There are similar patching mechanisms like the `ClearOrPatch` logic below.

gabrieljackson (commented):
Looks good and thanks for the tests!

I do have one primary question regarding the override logic below. Feel free to reach out to chat if you have questions and that is easier.

stafot:
@gabrieljackson I believe we are good to go. Tested on my local with all scenarios and simplified a lot one function that I understood that was double-checking things for no reason.

gabrieljackson (approved):
Looks good! Thank you for working though my feedback.

I think there is some extra validation we can do on the IP range strings in the future, but this is fine for now. Nice work.

stafot:
@gabrieljackson I will come with an IP validation addtion PR as soon as possible.

PR #955, opened 2023-08-01
[CLD-6035] Fix potential for installation (and therefore cluster) to leak after E2E failure

nickmisasi:
#### Summary
The recent refactor introduced re-using a cluster for multiple installation test suites. We ran into an issue where the first suite failed while running, but there was no logic to explicitly clean up the installation created from that test suite. This meant that an installation was still alive on the cluster when its deletion was later attempted after all tests had finished. Cluster couldn't be deleted because there was an installation on it. 

This change makes it so that after each run the installation is cleaned up, regardless of the result, so that there are no "orphaned" installations from previous suites that would prevent cluster deletion in the future. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6035
#### Release Note

```release-note
None
```

### DIFF
--- a/e2e/tests/installation/installation_test.go
+++ b/e2e/tests/installation/installation_test.go
@@ -8,6 +8,7 @@
 package installation
 
 import (
+	"context"
 	"github.com/pkg/errors"
 	"math/rand"
 	"testing"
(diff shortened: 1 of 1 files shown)


gabrieljackson (approved):
👍

PR #964, opened 2023-08-14
Add new database report tooling

gabrieljackson:
This change adds the following:
 - New flag for printing schema counts when running a multitenant database report.
 - New validation report command to run a check that database and installation records match.

Fixes https://mattermost.atlassian.net/browse/CLD-6130

```release-note
Add new database report tooling
```

### DIFF
--- a/cmd/cloud/databases.go
+++ b/cmd/cloud/databases.go
@@ -7,6 +7,7 @@ package main
 import (
 	"fmt"
 
+	"github.com/mattermost/mattermost-cloud/internal/common"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


cmd/cloud/databases.go
```diff
@@ -419,6 +420,94 @@ func newCmdDatabaseSchemaGet() *cobra.Command {
...
+
+func contains(val string, checkAgainst []string) bool {
+	for _, check := range checkAgainst {
+		if val == check {
+			return true
+		}
+	}
+
+	return false
+}
```
nickmisasi, line 509: Can re-use https://github.com/mattermost/mattermost-cloud/blob/master/internal/common/util.go#L8 instead

nickmisasi (approved):
Small nit otherwise LGTM!

cmd/cloud/databases.go
gabrieljackson: Ah good call. Forgot we had that.

PR #966, opened 2023-08-22
Add IP validator to IP filtering feature. 

stafot:
Add IP validator to IP filtering feature. 
Fix a minor bug in override mechanism.

#### Summary
Add IP validator to IP filtering feature. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6167

#### Release Note

```release-note
Add IP validator to IP filtering feature. 
Fix a minor bug in override mechanism.
```


model/installation_request.go
```diff
@@ -472,6 +477,25 @@ func (p *PatchInstallationRequest) addMissingIngressSourceRanges(installation *I
...
+func (p *PatchInstallationRequest) filterInvalidIngressSourceRanges(ips string) string {
+	var validIPs []string
+
+	allowedIPRanges := strings.Split(ips, ",")
+
+	for _, ip := range allowedIPRanges {
+		if IsIPRangeValid(ip) {
+			validIPs = append(validIPs, ip)
+		} else if !IsIPRangeValid(ip) {
+			logrus.Info("Invalid IP range will not be appended: ", ip)
```
gabrieljackson, line 489: I believe it would be better to return an error here. Currently, our patching logic enforces a completely valid request or errors so we should probably stick to that and avoid silent failures.

model/installation_request.go
```diff
@@ -460,9 +462,12 @@ func (p *PatchInstallationRequest) addMissingIngressSourceRanges(installation *I
 	}
 
 	allowedIPRanges := strings.Split(installation.AllowedIPRanges, ",")
+
 	for _, ip := range ips {
-		if !contains(allowedIPRanges, ip) {
+		if IsIPRangeValid(ip) && !contains(allowedIPRanges, ip) {
 			allowedIPRanges = append(allowedIPRanges, ip)
+		} else if !IsIPRangeValid(ip) {
+			logrus.Info("Invalid IP range will not be appended: ", ip)
```
gabrieljackson, line 470: I think the same thing applies here. For patching we should enforce the request to be valid.

gabrieljackson (changes requested):
Looks good, just requesting stricter validation handling.

model/installation_request.go
nickmisasi: I'm +1 on an error, and think we should not just throw out invalid ranges. From a user perspective (both myself the consuming dev and the end users) it'll be better to let people know there's an issue with their config than to silently throw out anything that's invalid. To the end user, disappearing config would look buggy

nickmisasi (changes requested):
Nothing to really add other than Gabe's mention on the error. LGTM otherwise.

model/installation_request.go
```diff
@@ -460,9 +462,12 @@ func (p *PatchInstallationRequest) addMissingIngressSourceRanges(installation *I
 	}
 
 	allowedIPRanges := strings.Split(installation.AllowedIPRanges, ",")
+
 	for _, ip := range ips {
-		if !contains(allowedIPRanges, ip) {
+		if IsIPRangeValid(ip) && !contains(allowedIPRanges, ip) {
 			allowedIPRanges = append(allowedIPRanges, ip)
+		} else if !IsIPRangeValid(ip) {
+			logrus.Errorf("Invalid IP range will not be appended: %s", ip)
```
nickmisasi, line 470: I think we still need a return here. Something that will cause a 400 to be returned to the caller

nickmisasi (changes requested):

gabrieljackson:
Thanks for your work on this @stafot! We will merge https://github.com/mattermost/mattermost-cloud/pull/970 shortly which will contain the IP validation logic.

PR #970, opened 2023-09-11
[CLD-6304] IP Validator for IP Filtering + Storage Changes for UX

nickmisasi:
#### Summary
Our UX designs for the IP Filtering feature require a few more fields to be tracked by the IP filters, including:
- Rule Name/Description - something the user can enter in to remember why a rule is set
- Enabled/Disabled - a boolean value so users can toggle on/off certain rules. This allows us to keep these disabled rules stored, rather than simply removing certain ones from the list

I've converted the AllowedIPRanges column of the installation table in the database to support JSON - it will now store data in a list of objects of type AllowedIPRange - an example can be seen below:
```
    "AllowedIPRanges": [
        {
            "CIDRBlock": "192.168.0.0/32",
            "Description": "Rule 1",
            "Enabled": false,
            "OwnerID": ""
        },
        {
            "CIDRBlock": "192.168.0.101/32",
            "Description": "Rule 2",
            "Enabled": false,
            "OwnerID": ""
        }
    ],
```

This PR also builds on top of Stavros' here: https://github.com/mattermost/mattermost-cloud/pull/966 as he introduced validation mechanisms for entered IP Ranges. 

Additions of these rules via the Cloud CLI will look like:
To _append_ to the existing list of AllowedIPRanges
`cloud installation update --installation hrouquejhfyzpxzm83s3jti5ey --allowed-ip-ranges="[{\"CIDRBlock\": \"192.168.0.0/32\"}]"`

To _replace_ the existing list of AllowedIPRanges
`cloud installation update --installation hrouquejhfyzpxzm83s3jti5ey --allowed-ip-ranges="[{\"CIDRBlock\": \"192.168.0.0/32\"}]" --override-ip-ranges=true"`

To _remove_ all existing AllowedIPRanges
`cloud installation update --installation hrouquejhfyzpxzm83s3jti5ey --allowed-ip-ranges="[]" --override-ip-ranges=true`

Note: instances of AllowedIPRange in a PATCH will be merged with existing instances of AllowedIPRange based on the CIDRBlock as a key.  That means that doing:

`cloud installation update --installation hrouquejhfyzpxzm83s3jti5ey --allowed-ip-ranges="[{\"CIDRBlock\": \"192.168.0.0/32\"}]"`

Followed by 

`cloud installation update --installation hrouquejhfyzpxzm83s3jti5ey --allowed-ip-ranges="[{\"CIDRBlock\": \"192.168.0.0/32\", Enabled: false}]"`

Will apply the `Enabled: false` to the AllowedIPRange without requiring the use of OverrideIPRanges.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6304

#### Release Note

```release-note
None
```


model/installation.go
```diff
@@ -81,6 +85,102 @@ type InstallationFilter struct {
...
+
+func (a *AllowedIPRanges) ToAnnotationString() string {
+	if a == nil {
+		return ""
+	}
+
+	var IPs []string
+	for _, allowedIPRange := range *a {
+		IPs = append(IPs, allowedIPRange.CIDRBlock)
+	}
```
nickmisasi, line 163: This needs to check the `Enabled` field before appending

model/installation.go
```diff
@@ -81,6 +85,102 @@ type InstallationFilter struct {
 	Name            string
 }
 
+type AllowedIPRanges []AllowedIPRange
+
+type AllowedIPRange struct {
+	CIDRBlock   string
+	Description string
+	Enabled     bool
+	// TODO - necessary?
```
stafot, line 94: @nickmisasi leftover?

nickmisasi: We still haven't decided :) I think it would be useful to store the MM User ID

stafot (approved):
LGTM. I need to test though locally

gabrieljackson (approved):
LGTM!

stafot:
Let's update the design doc with the new commands!

PR #974, opened 2023-10-16
Add `--deletion-locked` installation list filter flag

gabrieljackson:
This allows for listing installations that are either deletion locked or not. This change also adds the installation lock values to the default installation table output.

Fixes https://mattermost.atlassian.net/browse/CLD-6449

```release-note
Add `--deletion-locked` installation list filter flag
```

### DIFF
--- a/internal/api/helpers.go
+++ b/internal/api/helpers.go
@@ -91,3 +91,17 @@ func parseGroupConfig(u *url.URL) (bool, bool, error) {
 
 	return includeGroupConfig, includeGroupConfigOverrides, nil
 }
+
+func parseDeletionLocked(u *url.URL) (*bool, error) {
+	valueStr := u.Query().Get("deletion_locked")
+	if valueStr == "" {
+		return nil, nil
+	}
+
+	locked, err := strconv.ParseBool(valueStr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to deletion_locked boolean")
+	}
+
+	return &locked, nil
+}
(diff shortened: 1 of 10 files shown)


internal/api/helpers.go
```diff
@@ -91,3 +91,17 @@ func parseGroupConfig(u *url.URL) (bool, bool, error) {
...
+
+func parseDeletionLocked(u *url.URL) (*bool, error) {
+	valueStr := u.Query().Get("deletion_locked")
+	if valueStr == "" {
+		return nil, nil
+	}
+
+	locked, err := strconv.ParseBool(valueStr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to deletion_locked boolean")
```
nickmisasi, line 103:
```suggestion
		return nil, errors.Wrap(err, "failed to convert deletion_locked to boolean")
```

nickmisasi: 0/5 on what this changes to but it's just a bit awkward as is

internal/store/installation_test.go
```diff
@@ -1292,3 +1309,7 @@ func fixDNSRecords(num int) []*model.InstallationDNS {
 		{DomainName: fmt.Sprintf("dns-%d.example.com", num)},
 	}
 }
+
+func bToP(b bool) *bool {
+	return &b
+}
```
nickmisasi, line 1315: Can you use the one you added [here](https://github.com/mattermost/mattermost-cloud/pull/974/files#diff-8c3c91c8947fcc1b9fbfc09a56bb021b291aea7d6432a2ba751c8d93cec0c690R76)?

nickmisasi (approved):
2 small nits otherwise LGTM!

internal/store/installation_test.go
gabrieljackson: I actually can't cleanly because that one belongs to the `api_test` package. I am going to open a ticket to clean all this stuff up. We have a lot more of these hanging around simply because we didn't want to import large packages for simple test tooling.

PR #975, opened 2023-10-18
[CLD-6421] Add support for --master-instance-type on cluster resize for KOPS

nickmisasi:
#### Summary
Adds support for `--size-master-instance-type` when resizing a cluster. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6421

#### Release Note

```release-note
None
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -389,6 +389,10 @@ func executeClusterResizeCmd(flags clusterResizeFlags) error {
 		request.NodeMaxCount = &flags.nodeMaxCount
 	}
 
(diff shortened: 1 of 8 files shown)


gabrieljackson (commented):
Hey, this is a good start, but I believe the master instance type change will not lead to any action being taken on the cluster.

We will need to update the kops logic here to interpret the change request into kops actions for when the update occurs.

https://github.com/mattermost/mattermost-cloud/blob/master/internal/provisioner/kops_provisioner.go#L709

gabrieljackson (approved):
LGTM! Nice work

stafot (approved):
LGTM

PR #981, opened 2023-11-15
Add group nil check for SLI checks

gabrieljackson:
This prevents a panic when an installation is removed from a group.

Fixes https://mattermost.atlassian.net/browse/CLD-6656

```release-note
Add group nil check for SLI checks
```


nickmisasi (approved):

stylianosrigas (approved):
LGTM

PR #982, opened 2023-11-15
Add webhook header counts to table output

gabrieljackson:
The number of headers for each webhook is now displayed in the standard table when listing webhooks.

```release-note
Add webhook header counts to table output
```

### DIFF
--- a/cmd/cloud/webhook.go
+++ b/cmd/cloud/webhook.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/mattermost/mattermost-cloud/model"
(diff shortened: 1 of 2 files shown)


cmd/cloud/webhook.go
```diff
@@ -146,10 +147,10 @@ func executeWebhookListCmd(flags webhookListFlag) error {
 	if flags.outputToTable {
 		table := tablewriter.NewWriter(os.Stdout)
 		table.SetAlignment(tablewriter.ALIGN_LEFT)
-		table.SetHeader([]string{"ID", "OWNER", "URL"})
+		table.SetHeader([]string{"ID", "OWNER", "URL", "HEADERS"})
```
nickmisasi, line 150:
```suggestion
		table.SetHeader([]string{"ID", "OWNER", "URL", "HTTP HEADERS"})
```

Nit suggestion: "Headers" is a bit vague, especially considering this is in the "header" of a table :P

nickmisasi (approved):
One suggestion not blocking. LGTM

PR #983, opened 2023-11-16
Support arbitrary domain creation for installations

gabrieljackson:
We were previously blocking all domain additions for installations that didn't have a subdomain that matched the `name` value. There were many reasons for this, but it was inflexible and needed to be improved. There is more work to do in supporting custom domains, but this allows for completely switching an installation's domain after creation.

 - Allows for creating any new installation domain assuming the provisioner has access to the zone.
 - Allows for deleting domains for an installation after they are no longer marked as primary.
 - Fixes a bug where certain domains couldn't be selected as primary.
 - Improves domain management logging.

Fixes https://mattermost.atlassian.net/browse/CLD-6663

```release-note
Support arbitrary domain creation for installations
```

### DIFF
--- a/internal/store/installation_dns.go
+++ b/internal/store/installation_dns.go
@@ -91,14 +91,41 @@ func (sqlStore *SQLStore) getDNSRecordsForInstallations(db queryer, installation
 
 // SwitchPrimaryInstallationDomain sets given domain as primary reducing all others to non-primary.
 func (sqlStore *SQLStore) SwitchPrimaryInstallationDomain(installationID string, installationDNSID string) error {
-	query := sq.Update(installationDNSTable).
-		Set("IsPrimary",
-			sq.Case().When(sq.Expr("ID = ?", installationDNSID), "true").Else("false"),
-		).Where("InstallationID = ?", installationID)
+	tx, err := sqlStore.beginTransaction(sqlStore.db)
+	if err != nil {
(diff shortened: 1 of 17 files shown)


internal/store/installation_dns.go
```diff
@@ -91,15 +91,42 @@ func (sqlStore *SQLStore) getDNSRecordsForInstallations(db queryer, installation
...
 		return errors.Wrap(err, "failed to switch primary DNS")
 	}
+	// Perform validation that exactly one row was updated just in case.
+	count, err := result.RowsAffected()
+	if err != nil {
+		return errors.Wrap(err, "failed to count rows affected on IsPrimary=true update")
+	}
+	if count != 1 {
+		return errors.Errorf("expected exactly 1 row update, but got %d", count)
+	}
```
nickmisasi, line 124: The installationdns tables' ID column is a unique primary key so I think this check of one row edited will be redundant given you are using the ID column in the where clause

nickmisasi (approved):
One nit otherwise LGTM

internal/store/installation_dns.go
gabrieljackson: I was mainly worried about 0 rows being edited. Starting to think this level of concern isn't necessary though.

stylianosrigas (approved):
LGTM 🚀

PR #985, opened 2023-11-20
Ip filtering refactor

stafot:
#### Summary
IP filtering refactor

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6683

#### Release Note

```release-note
Refactor IP Filtering to use 419 status code instead of 403
```

### DIFF
--- a/internal/provisioner/cluster_installation_provisioner.go
+++ b/internal/provisioner/cluster_installation_provisioner.go
@@ -1086,7 +1086,7 @@ func getHibernatingIngressAnnotations() *model.IngressAnnotations {
 //}
 
 func addSourceRangeWhitelistToAnnotations(annotations *model.IngressAnnotations, allowedIPRanges *model.AllowedIPRanges, internalIPRanges []string) {
-	// If all allowedIPRanges are disabled, we don't continue so we don't unintentionally lock the workspace to internal traffic only
+	// If all allowedIPRanges are disabled, we don't continue, so we don't unintentionally lock the workspace to internal traffic only
 	if allowedIPRanges.AllRulesAreDisabled() {
 		return
 	}
--- a/internal/provisioner/cluster_installation_provisioner.go
+++ b/internal/provisioner/cluster_installation_provisioner.go
@@ -1062,29 +1062,6 @@ func getHibernatingIngressAnnotations() *model.IngressAnnotations {
 	return annotations
 }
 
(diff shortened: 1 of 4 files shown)


model/ingress.go
```diff
@@ -95,3 +98,24 @@ func (ia *IngressAnnotations) SetDefaults() {
...
+	configSnippet := existingSnippet + "\n" + allowDirectivesBuilder.String() + `
+       deny all;
+       error_page 403 =419 /custom_419_page;
+   `
+
+	// Setting up the IngressAnnotations struct
+	ia := &IngressAnnotations{
+		ConfigurationSnippet: configSnippet,
+	}
+	return ia
```
nickmisasi, line 120: I think you could continue to use strings.Builder to make it a little easier on the eyes. Something like:

```suggestion
    // Build the configSnippet using strings.Builder
    configSnippetBuilder := strings.Builder{}
    configSnippetBuilder.WriteString(existingSnippet)
    configSnippetBuilder.WriteString("\n")
    configSnippetBuilder.WriteString(allowDirectivesBuilder.String())
    configSnippetBuilder.WriteString(`
       deny all;
       error_page 403 =419 /custom_419_page;
   `)
    configSnippet := configSnippetBuilder.String()

    // Setting up the IngressAnnotations struct
    ia := &IngressAnnotations{
        ConfigurationSnippet: configSnippet,
    }
    return ia
```

internal/provisioner/cluster_installation_provisioner.go
```diff
@@ -1083,6 +1083,10 @@ func addSourceRangeWhitelistToAnnotations(annotations *model.IngressAnnotations,
 	}
 
 	annotations.WhitelistSourceRange = allIPRanges
+
+	//update the ConfigurationSnippet
+	updatedAnnotations := model.ConfigureIngressAnnotations(allIPRanges, annotations.ConfigurationSnippet)
+	annotations.ConfigurationSnippet = updatedAnnotations.ConfigurationSnippet
```
nickmisasi, line 1089: Do we still need whitelist source ranges to be set now? Will that directive not conflict with our configuration snippets?

stafot: @nickmisasi Not sure. Looks like they don't conflict, during my testing. But your point seems valid.

PR #987, opened 2023-11-27
Temporary solution to deploy custom error configmaps in clusters

stafot:
#### Summary
Temporary solution to deploy custom error configmaps in clusters

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-6718

#### Release Note

```release-note
Temporary solution to deploy custom error configmaps in clusters
```

### DIFF
--- a/internal/provisioner/provisioner_utils.go
+++ b/internal/provisioner/provisioner_utils.go
@@ -245,3 +245,15 @@ func ensureNamespaceExists(k8sClient *k8s.KubeClient, namespace string) error {
 	}
 	return nil
 }
+
+// checkCustomErrorPagesConfigMapExists checks if a ConfigMap with custom error pages exists.
+func checkCustomErrorPagesConfigMapExists(k8sClient *k8s.KubeClient, namespace string) (bool, error) {
+	_, err := k8sClient.Clientset.CoreV1().ConfigMaps(namespace).Get(context.TODO(), "custom-error-pages", metav1.GetOptions{})
+	if k8sErrors.IsNotFound(err) {
+		return false, nil
+	} else if err != nil {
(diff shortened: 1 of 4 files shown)


andrleite (approved):

internal/provisioner/provisioner_utils.go
```diff
@@ -174,3 +175,73 @@ func waitForNamespacesDeleted(ctx context.Context, namespaces []string, k8sClien
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta http-equiv="refresh" content="0; url=%s/cloud/inaccessible" />
+    <title>HTML 5 Boilerplate</title>
+</head>
+<body>
+</body>
+</html>`, portalURL),
+	}
```
nickmisasi, line 211: Should we use template files embedded into the provisioner binary that can be loaded and populated here using Go's template library? It might be a bit cleaner than having HTML in a map with Sprintf?

nickmisasi (commented):
One comment about potential refactor. 

I think we should probably also have security take a look at this to make sure there's no potential issues around the HTML we added here

internal/provisioner/provisioner_utils.go
esarafianou: @stafot the portalURL configured by us internally, right?

stafot: Right. But this PR is now obsolete as fortunately ingress-nginx approved my PR to support this functionality on their helm chart. So I will close this one and we will avoid this hackie solution. 
Thanks for taking the time and review though.

PR #1004, opened 2024-01-15
Migrate Mattermost Operator deployment to Helm

gabrieljackson:
This change includes the necessary logic to migrate the Mattermost Operator resources to the helm chart. The migration happens via the `cluster provision` process.

Future changes will be made to remove the migration process when it is no longer needed as well as the existing manifests.

https://mattermost.atlassian.net/browse/CLD-6942

```release-note
Migrate Mattermost Operator deployment to Helm
```


internal/tools/kops/run.go
```diff
@@ -68,7 +68,6 @@ func (c *Cmd) run(arg ...string) (stdout []byte, stderr []byte, err error) {
 	cmd := exec.Command(c.kopsPath, arg...)
 	cmd.Env = append(
 		os.Environ(),
-		"KOPS_FEATURE_FLAGS=SpecOverrideFlag", // Required in kops 1.14.0 for set command.
```
gabrieljackson, line 71: This was deprecated and is no longer needed.

stylianosrigas:
@andrleite Added for review, as this will be an extra utility to manage via Argo ;)

cmd/cloud/server.go
```diff
@@ -513,16 +514,15 @@ func checkRequirements(logger logrus.FieldLogger) error {
 	}
 	logger.Infof("[startup-check] Using helm: %s", version)
 
-	// Check for extra tools that don't have a wrapper, but are still required.
-	extraTools := []string{
-		"kubectl",
+	kubectlClient, err := kubectl.New(silentLogger)
+	if err != nil {
+		return errors.Wrap(err, "failed kubectl client health check")
```
nickmisasi, line 519: nit: this could be more descriptive like `client does not exist` since it's really just doing a lookup that the binary exists in the PATH?

nickmisasi: I guess the error that you're wrapping states that... so maybe you don't even need to wrap it? Or just leave it.. 0/5

nickmisasi (approved):
LGTM

stafot (approved):

andrleite (approved):

andrleite:
LGTM,  Just a comment, we're removing the logic to deploy add-ons with helm from provisioner and migrating it to argocd. @gabrieljackson we can have a quick chat and I can show you what I'm doing.

gabrieljackson:
@andrleite yep, we definitely can chat. This work was primarily meant to handle the transition to helm as safely as possible. Afterwards we can migrate to argocd.

PR #1009, opened 2024-01-22
Helm client refactor and local chart install support

gabrieljackson:
Even with the migration of utilities away from being managed by the provisioner, there is still a need to be able to test charts locally. This change attempts to support that use-case while also refactoring the helm client. As a result, provisioning speed is now also improved. This change does the following:

 - Adds server flag for deploying the mattermost operator helm chart locally from 'helm-charts/mattermost-operator'
 - Removes a redundant helm debug flag and instead relies on the debug log flag.
 - Cleans up helm repo add and repo update and moves them inside the helm package.
 - Removes previous logic where every helm repo added also caused a complete update of all configured repos. Instead, repos are now only updated once per cluster provision which reduces log noise and speeds up the process.
 - Adds helm client logic to install local charts.
 - Refactors the helm and kubectl clients to take in a kubecfg path on creation which is used for all commands. Passing in `--kubeconfig` flags is no longer needed for each command.

Fixes https://mattermost.atlassian.net/browse/CLD-7009

```release-note
Helm client refactor and local chart install support
```

### DIFF
--- a/cmd/cloud/server.go
+++ b/cmd/cloud/server.go
@@ -124,10 +124,10 @@ func executeServerCmd(flags serverFlags) error {
 	}
 	model.SetUtilityDefaults(flags.utilitiesGitURL)
 
-	if flags.deployLocalMattermostOperator {
-		_, err := os.Stat(helm.LocalMattermostOperatorHelmDir)
+	if flags.mattermostOperatorHelmDir != "" {
+		_, err := os.Stat(flags.mattermostOperatorHelmDir)
 		if err != nil {
-			return errors.Wrapf(err, "failed to ensure directory %s exists; disable server flag deploy-local-mattermost-operator or ensure directory exists", helm.LocalMattermostOperatorHelmDir)
+			return errors.Wrap(err, "failed to ensure local operator helm directory exists; check your mattermost-operator-helm-dir flag value")
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


gabrieljackson:
Just leaving a separate note that I tested this by symlinking my WIP helm chart changes in my local `mattermost-helm` repo checkout to `helm-charts/mattermost-operator` and it works great.

cmd/cloud/server.go
```diff
@@ -130,6 +124,13 @@ func executeServerCmd(flags serverFlags) error {
 	}
 	model.SetUtilityDefaults(flags.utilitiesGitURL)
 
+	if flags.deployLocalMattermostOperator {
+		_, err := os.Stat(helm.LocalMattermostOperatorHelmDir)
+		if err != nil {
+			return errors.Wrapf(err, "failed to ensure directory %s exists; disable server flag deploy-local-mattermost-operator or ensure directory exists", helm.LocalMattermostOperatorHelmDir)
+		}
+	}
```
nickmisasi, line 132: Rather than hard coding the local directory, could you just accept a folder as the argument rather than a boolean? Would allow you to not have to simlink/copy/paste into the helm-charts/mattermost-operator folder

nickmisasi (approved):
One small suggestion that I'm 0/5 on, otherwise LGTM!

andrleite (approved):

PR #1017, opened 2024-02-08
[CLD-5863] Enabling provisioner to deploy utilities using Argocd.

andrleite:
#### Summary
Enabling the provisioner to deploy utilities using Argocd. This change doesn't cover Utilities deployed by the provisioner through manifests files (mattermost-operator, bifrost and external-snapshotter). To deploy a utility using Argocd we need to set the version as "unmanaged" in the creation or provision process. For example: `--nginx-version "unmanaged"`

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/CLD-5863

#### Release Note

```release-note
Enable provisioner to deploy utilities through Argocd
```


internal/provisioner/utility/utility_group.go
```diff
@@ -214,19 +240,22 @@ func (group utilityGroup) ProvisionUtilityGroup() error {
 	}
 
 	for _, utility := range group.utilities {
+		fmt.Printf("Provisioning utility %s\n", utility.Name())
```
gabrieljackson, line 243: User logger instead?

internal/provisioner/utility/unmanaged.go
```diff
@@ -29,14 +56,218 @@ func (u *unmanaged) ValuesPath() string {
...
+	switch u.Name() {
+	case model.PgbouncerCanonicalName:
+		err = deployManifests(k8sClient, u.logger)
+		if err != nil {
+			return err
+		}
+		if err = u.utiliyArgocdDeploy(u.Name()); err != nil {
+			return errors.Wrapf(err, "failed to provision %s utility", u.Name())
+		}
+
```
gabrieljackson, line 80: So we don't really want to overload the `unmanaged` object like this. Ideally, you use the `unmanaged` object directly for utilities that don't have any special creation or deletion logic. If we need that logic like what is going on here then it probably makes more sense to have things like `unmanagedPgbouncer` and `unamangedNginx` just like we have for the normal managed handles.

I know that we want to get this is so I won't make this a blocker, but it should be cleaned up later.

gabrieljackson (commented):
I did a quick pass on this and wanted to fire off these questions before proceeding.

go.mod
```diff
@@ -86,90 +97,188 @@ require (
...
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.26.11
+	k8s.io/kubectl => k8s.io/kubectl v0.26.11
+	k8s.io/kubelet => k8s.io/kubelet v0.26.11
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.26.11
+	k8s.io/metrics => k8s.io/metrics v0.26.11
+	k8s.io/mount-utils => k8s.io/mount-utils v0.26.11
+	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.26.11
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.26.11
+	// https://github.com/golang/go/issues/33546#issuecomment-519656923
+	sigs.k8s.io/kustomize/api => sigs.k8s.io/kustomize/api v0.12.1
```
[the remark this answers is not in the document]
andrleite: We need this to be able to integrate with Argocd API, https://argo-cd.readthedocs.io/en/stable/user-guide/import/
I had a few incompatibility issues with other libraries and with this config, everything worked. 
I can combine the replace block.

internal/provisioner/cluster_register.go
```diff
@@ -86,7 +68,7 @@ func (cr *ClusterRegister) clusterRegister(s3StateStore string) error {
 		return errors.Wrap(err, "failed to commit to repo")
 	}
 
-	if err = cr.gitClient.Push(logger); err != nil {
+	if err = cr.gitClient.Push("feat-CLD-5708", logger); err != nil {
```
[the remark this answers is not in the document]
andrleite: It should be `main`.  I changed it back for testing purposes and forgot I'll fix it. We need to change it to test only I don't think it needs to be dynamic.

internal/provisioner/utility/utility_group.go
andrleite: definitely yes, I'm going to fix it

internal/provisioner/cluster_register.go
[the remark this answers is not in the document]
andrleite: tbh, it'll be better to have it configurable to avoid changing t in different places of the code in cases of test in a specific branch. I'll review it.

internal/provisioner/utility/unmanaged.go
andrleite: Understood, I'll try to change it quickly but we can add a task to make it better.

internal/provisioner/cluster_register.go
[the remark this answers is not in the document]
andrleite: fixed it

internal/provisioner/cluster_register.go
```diff
@@ -169,7 +145,7 @@ func (cr *ClusterRegister) deregisterClusterFromArgocd() error {
...
```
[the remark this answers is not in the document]
andrleite: fixed it

internal/provisioner/utility/utility_group.go
andrleite: fixed it

internal/provisioner/cluster_provisioning.go
```diff
@@ -26,11 +28,38 @@ import (
...
+
+			cluster.UtilityMetadata.ArgocdClusterRegister.Registered = true
+			logger.Infof("Cluster: %s successfully registered in argocd", cluster.ID)
+
+		default:
+			logger.Info("Cluster already registered, skipping argocd cluster register")
+		}
+	default:
+		logger.Info("ManagedByArgocd is false, skipping argocd cluster registration")
+	}
```
nickmisasi, line 61: Any particular reason to use a switch here, rather than just an if/else? You could probably narrow down to one level as well - adding the `ManagedByArgocd` value to the logger fields will allow you to communicate that information via logs, then you can just check `cluster.UtilityMetdata.ManagedByArgocd && !cluster.UtilityMetadata.ArgocdClusterRegister.Registered`

internal/provisioner/utility/argo_utility.go
```diff
@@ -0,0 +1,190 @@
...
+
+func ProvisionUtilityArgocd(utilityName, tempDir, clusterID string, allowCIDRRangeList []string, awsClient aws.AWS,
+	gitClient git.Client, argocdClient argocd.Client, logger log.FieldLogger) error {
+	// Pull the latest changes from the repo
+	err := gitClient.Pull(logger)
+	if err != nil {
+		return errors.Wrap(err, "failed to pull from repo")
+	}
+
+	//TODO: Skip provision utility if it is already provisioned
```
nickmisasi, line 31: Should we open a ticket for this?

internal/provisioner/utility/argo_utility.go
```diff
@@ -0,0 +1,190 @@
...
+			return err
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
```
nickmisasi, line 190: While this works.. Consider using the built in Go template engine

nickmisasi (approved):
Couple small nits but not much else to add on top of Gabe's review. Nice work!

internal/provisioner/cluster_provisioning.go
andrleite: No particular reason I was trying to keep it simple but your suggestion is better. I changed it. Thank you @nickmisasi

internal/provisioner/utility/argo_utility.go
andrleite: Makes sense, I'm going to open a task for it. Thanks

gabrieljackson (approved):
Thanks for working though our comments. LGTM

PR #1025, opened 2024-03-05
Add logic to skip AMI updates for instance groups

gabrieljackson:
There are situations where it is ideal to not apply the standard AMI to all kops instance groups in a cluster. This change allows for setting an instance group metadata label with a key of `mattermost/cloud-provisioner-ami-skip` and any value to instruct the provisioner to skip any AMI updates on that instance group.

Fixes https://mattermost.atlassian.net/browse/CLD-7252

Example instance group snippet:

```
apiVersion: kops.k8s.io/v1alpha2
kind: InstanceGroup
metadata:
  creationTimestamp: "2024-03-04T14:44:27Z"
  generation: 3
  labels:
    kops.k8s.io/cluster: etuajy5q93ymdqj9pcb1oa4bzw-kops.k8s.local
    mattermost/cloud-provisioner-ami-skip: "true"
  name: nodes-us-east-1a
spec:
  ...
```

```release-note
Add logic to skip AMI updates for instance groups
```

### DIFF
--- a/internal/provisioner/kops_utils.go
+++ b/internal/provisioner/kops_utils.go
@@ -15,6 +15,11 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// SkipAMIUpdateLabelKey is the label key that can be applied to a kops instance
+// group to indicate to the provisioner that it should skip updating the AMI.
+// The label value can be any string.
+const SkipAMIUpdateLabelKey string = "mattermost/cloud-provisioner-ami-skip"
+
 // verifyTerraformAndKopsMatch looks at terraform output and verifies that the
 // given kops name matches. This should only catch errors where terraform output
 // was incorrectly created from kops or if the terraform client is targeting the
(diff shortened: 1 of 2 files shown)


internal/provisioner/kops_utils.go
```diff
@@ -50,10 +55,14 @@ func updateKopsInstanceGroupAMIs(kops *kops.Cmd, kopsMetadata *model.KopsMetadat
 	var ami string
 	for _, ig := range instanceGroups {
 		if ig.Spec.Image != kopsMetadata.ChangeRequest.AMI {
+			if val, ok := ig.Metadata.Labels[SkipAMIUpdateLabelKey]; ok {
```
nickmisasi, line 58: Should you check `val` for truthy here?

nickmisasi (approved):
One small thing otherwise LGTM

internal/provisioner/kops_utils.go
gabrieljackson: We could, but the value really doesn't matter so I decided to keep it simple for now.

stafot:
thanks for this work @gabrieljackson

stafot (approved):

PR #1029, opened 2024-03-11
feat: Introduce --header global variable to access provisioner with auth

toninis:
#### Summary

- Introduce `--header` cli input to optionally provide the auth token
- Introduce new `createClient` function to create the respective client if token is provided
- Rerouted all client initializations through the new function 
- Added clusterFlags in all base flag types 

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7291

#### Release Note

```release-note
Introduce --header variable in order to add extra headers for provisioner server api calls
```

### DIFF
--- a/cmd/cloud/cluster_flag.go
+++ b/cmd/cloud/cluster_flag.go
@@ -6,16 +6,19 @@ import (
 
 func setClusterFlags(command *cobra.Command) {
 	command.PersistentFlags().String("server", defaultLocalServerAPI, "The provisioning server whose API will be queried.")
+	command.PersistentFlags().String("token", "", "The auth token to access provisioning server.")
 	command.PersistentFlags().Bool("dry-run", false, "When set to true, only print the API request without sending it.")
 }
 
 type clusterFlags struct {
 	serverAddress string
+	token         string
 	dryRun        bool
 }
 
 func (flags *clusterFlags) addFlags(command *cobra.Command) {
 	flags.serverAddress, _ = command.Flags().GetString("server")
+	flags.token, _ = command.Flags().GetString("token")
 	flags.dryRun, _ = command.Flags().GetBool("dry-run")
 }
 
--- a/cmd/cloud/cluster_flag.go
+++ b/cmd/cloud/cluster_flag.go
@@ -6,19 +6,19 @@ import (
 
 func setClusterFlags(command *cobra.Command) {
 	command.PersistentFlags().String("server", defaultLocalServerAPI, "The provisioning server whose API will be queried.")
-	command.PersistentFlags().String("token", "", "The auth token to access provisioning server.")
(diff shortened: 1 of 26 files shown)


cmd/cloud/cluster_flag.go
```diff
@@ -6,16 +6,19 @@ import (
 
 func setClusterFlags(command *cobra.Command) {
 	command.PersistentFlags().String("server", defaultLocalServerAPI, "The provisioning server whose API will be queried.")
+	command.PersistentFlags().String("token", "", "The auth token to access provisioning server.")
```
gabrieljackson, line 9: Let's change this to accept any number of header keys+values. I believe `Flags().StringToStringVar` may be what we want here.

This way we can specify new headers or change the token key if we want.

gabrieljackson (changes requested):

cmd/cloud/cluster_flag.go
toninis: Yep seems more valid since we provide the option . 
I will rename the flag to `--extra-headers` WDYT ?

cmd/cloud/cluster_flag.go
```diff
@@ -6,16 +6,19 @@ import (
 
 func setClusterFlags(command *cobra.Command) {
 	command.PersistentFlags().String("server", defaultLocalServerAPI, "The provisioning server whose API will be queried.")
+	command.PersistentFlags().StringToString("header", nil, "The extra headers to sent in every API call towards the provisioning server. Accepts format: HEADER_KEY=HEADER_VALUE. Use the flag multiple times to set multiple headers.")
```
gabrieljackson, line 9:
```suggestion
	command.PersistentFlags().StringToString("header", nil, "The extra headers to send in every API call towards the provisioning server. Accepts format: HEADER_KEY=HEADER_VALUE. Use the flag multiple times to set multiple headers.")
```

cmd/cloud/utils.go
```diff
@@ -57,3 +57,15 @@ func runDryRun(request interface{}) error {
...
 	return nil
 }
+
+func createClient(flags clusterFlags) *model.Client {
+	if len(flags.headers) > 0 {
+		authHeaders := make(map[string]string, len(flags.headers))
+		for key, value := range flags.headers {
+			authHeaders[key] = value
+		}
+		return model.NewClientWithHeaders(flags.serverAddress, authHeaders)
```
gabrieljackson, line 67: Technically the headers can be anything, so I would just use `headers` for the var name.
```suggestion
		headers := make(map[string]string, len(flags.headers))
		for key, value := range flags.headers {
			headers[key] = value
		}
		return model.NewClientWithHeaders(flags.serverAddress, headers)
```

gabrieljackson (approved):
Just two small non-blocker suggestions below. Looks good!

cmd/cloud/cluster_flag.go
gabrieljackson: The name you selected (`headers`) LGTM

spirosoik (approved):
LGTM! great job

PR #1032, opened 2024-03-30
Change AMI handling for kops metadata

gabrieljackson:
The recent addition of AMI overrides can lead to warnings being generated when rebuilding the kops metadata that is stored in the cluster object. This change alters the AMI warning logic and now stores the AMI value for each instance group separately.

Fixes https://mattermost.atlassian.net/browse/CLD-7414

```release-note
Change AMI handling for kops metadata
```


stafot (approved):
LGTM

nickmisasi (approved):
LGTM

PR #1033, opened 2024-04-04
Support arm AMI IG rollouts. Support AMI names and not only AMI IDs

stafot:
#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-7403
https://mattermost.atlassian.net/browse/CLD-7411

#### Release Note

```release-note
Support arm AMI IG rollouts. Support AMI names and not only AMI IDs
```

### DIFF
--- a/cmd/cloud/cluster.go
+++ b/cmd/cloud/cluster.go
@@ -546,7 +546,7 @@ func executeClusterListCmd(flags clusterListFlags) error {
 }
 
 func defaultClustersTableData(clusters []*model.ClusterDTO) ([]string, [][]string) {
-	keys := []string{"ID", "STATE", "VERSION", "MASTER NODES", "WORKER NODES", "AMI ID", "NETWORKING", "VPC", "STATUS"}
+	keys := []string{"ID", "STATE", "VERSION", "MASTER NODES", "WORKER NODES", "AMI ID/NAME", "NETWORKING", "VPC", "STATUS"}
 	values := make([][]string, 0, len(clusters))
 
 	for _, cluster := range clusters {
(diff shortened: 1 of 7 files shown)


stylianosrigas (approved):

cmd/cloud/cluster.go
```diff
@@ -574,7 +574,7 @@ func defaultClustersTableData(clusters []*model.ClusterDTO) ([]string, [][]strin
 			provisionerMetadata.Version,
 			fmt.Sprintf("%d x %s", masterCount, masterInstanceType),
 			fmt.Sprintf("%d x %s (max %d)", provisionerMetadata.NodeMinCount, provisionerMetadata.NodeInstanceType, provisionerMetadata.NodeMaxCount),
-			provisionerMetadata.AMI,
+			NormalizeAMISuffix(provisionerMetadata.AMI),
```
gabrieljackson, line 577: So instead of normalizing the AMI value in the cluster table, shouldn't we be storing the normalized value directly? That seems less confusing overall and eliminates the need for redoing it when showing cluster config.

internal/provisioner/kops_utils.go
```diff
@@ -130,3 +132,35 @@ func updateWorkersKopsInstanceGroupValue(kops *kops.Cmd, kopsMetadata *model.Kop
...
+	// Regex to find existing architecture suffix in the AMI value.
+	re := regexp.MustCompile(`-(amd64|arm64)$`)
+
+	if !re.MatchString(ami) {
+		// Case 2: No architecture suffix, append default "amd64" or the provided label.
+		defaultSuffix := "amd64"
+		if archLabel == "arm64" {
+			defaultSuffix = archLabel
+		}
+		return ami + "-" + defaultSuffix
```
gabrieljackson, line 155: This is a little confusing to read as the default part my no longer be accurate. I would just call this variable `suffix`.

cmd/cloud/cluster.go
stafot: Not sure how this plays with your recent changes having AMIs per instance group. Let's discuss on planning to find a common decision

stafot: Removed for now. If the representation is not ok for us let's improve it on another iteration. This part was totally minor for the purpose of this PR.

internal/provisioner/kops_utils.go
stafot: changed upon suggestion

stafot:
gentle ping to re-review @gabrieljackson

gabrieljackson (approved):

PR #1041, opened 2024-04-22
Feat/cld 7451

andrleite:
#### Summary
 - This PR is going to skip the commit utility to argocd if it's already registered.
 - Fix file check for argo utility registration.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/CLD-7451

#### Release Note

```release-note
None
```

### DIFF
--- a/internal/provisioner/utility/utility_group.go
+++ b/internal/provisioner/utility/utility_group.go
@@ -241,6 +241,14 @@ func (group utilityGroup) ProvisionUtilityGroup() error {
 	for _, utility := range group.utilities {
 		logger.Infof("Provisioning utility %s\n", utility.Name())
 
+		if group.cluster.ActualUtilityVersion(utility.Name()).Chart == "unmanaged" {
+			logger.WithFields(log.Fields{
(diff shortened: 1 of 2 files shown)


internal/provisioner/utility/utility_group.go
```diff
@@ -241,6 +241,14 @@ func (group utilityGroup) ProvisionUtilityGroup() error {
 	for _, utility := range group.utilities {
 		logger.Infof("Provisioning utility %s\n", utility.Name())
 
+		if group.cluster.ActualUtilityVersion(utility.Name()).Chart == "unmanaged" {
```
gabrieljackson, line 244: I think we probably want to check that the actual version and the desired version are unmanaged right? I know it's unlikely that we would want to go back the provisioner-managed, but this check wouldn't allow it as is.

andrleite: Thanks, Gabe. I changed it to consider the `desiredVersion` in the if statement.

gabrieljackson (approved):

PR #1043, opened 2024-04-25
Create cluster dir for argocd output custom values

andrleite:
#### Summary
Create cluster dir for argocd output custom values

#### Release Note

```release-note
None
```

### DIFF
--- a/internal/provisioner/utility/argo_utility.go
+++ b/internal/provisioner/utility/argo_utility.go
@@ -47,8 +47,15 @@ func ProvisionUtilityArgocd(utilityName, tempDir, clusterID string, allowCIDRRan
 		return errors.Wrap(err, "failed to write argo application file")
 	}
 
+	// Create the cluster output directory for the utility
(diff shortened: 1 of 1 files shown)


internal/provisioner/utility/argo_utility.go
```diff
@@ -47,8 +47,15 @@ func ProvisionUtilityArgocd(utilityName, tempDir, clusterID string, allowCIDRRan
 		return errors.Wrap(err, "failed to write argo application file")
 	}
 
+	// Create the cluster output directory for the utility
+	clusterOutputDir := tempDir + "/apps/" + awsClient.GetCloudEnvironmentName() + "/helm-values/" + clusterID
```
gabrieljackson, line 51: Just a heads-up that you can use `filepath.Join()` to do this sort of thing and reduce risk of building a bad path: https://pkg.go.dev/path/filepath#Join

gabrieljackson (approved):
LGTM, just one thing for the future

PR #1056, opened 2024-06-06
Use new bulk import flags

gabrieljackson:
The flag --extract-content=false is used to increase import performance by skipping attachment parsing. The --json flag replaces the old format flag, but behaves the same.

Fixes https://mattermost.atlassian.net/browse/CLD-7856

```release-note
Use new bulk import flags
```


stylianosrigas (approved):
LGTM

nickmisasi (approved):

PR #1059, opened 2024-07-02
Add external cluster importing

gabrieljackson:
This allows for importing k8s clusters that are managed externally. An AWS secret manager secret containing the kubecfg must be provided so that the provisioner can securely connect to create and manage Mattermost installations. Other functions such as cluster resizing and provisioning are no-ops for external clusters.

Fixes https://mattermost.atlassian.net/browse/CLD-7703

```release-note
Add external cluster importing
```


nickmisasi (approved):
LGTM! I thought this would have taken quite a bit more code than it did - awesome! 

Probably something to address in a subsequent PR, but this one feels like it would be a good candidate for an E2E test

gabrieljackson:
> LGTM! I thought this would have taken quite a bit more code than it did - awesome!
> 
> Probably something to address in a subsequent PR, but this one feels like it would be a good candidate for an E2E test

Yep! I have a follow-up for better handling of clusters in our AWS accounts which should help with E2E testing as well.

andrleite (approved):

PR #1061, opened 2024-07-11
Use argocd to manage utilities for e2e-cluster

andrleite:
#### Summary
Use argocd to manage utilities for e2e-cluster

#### Ticket Link
 Fixes https://mattermost.atlassian.net/browse/CLD-7783

#### Release Note

```release-note
None
```

### DIFF
--- a/e2e/tests/shared/shared.go
+++ b/e2e/tests/shared/shared.go
@@ -98,6 +98,47 @@ func SetupTestWithDefaults(testName string) (*Test, error) {
 		AMI:                config.KopsAMI,
 		VPC:                config.VPC,
 		Provisioner:        config.Provisioner,
+		ArgocdClusterRegister: map[string]string{
+			"cluster-type": "customer",
+		},
+		DesiredUtilityVersions: map[string]*model.HelmUtilityVersion{
+			"nginx": {
+				Chart: "unmanaged",
+			},
+			"nginx-internal": {
+				Chart: "unmanaged",
+			},
+			"thanos": {
+				Chart: "unmanaged",
+			},
+			"prometheus-operator": {
+				Chart: "unmanaged",
+			},
+			"node-problem-detector": {
+				Chart: "unmanaged",
+			},
+			"cloudprober": {
+				Chart: "unmanaged",
+			},
+			"fluent-bit": {
+				Chart: "unmanaged",
+			},
+			"teleport": {
+				Chart: "unmanaged",
+			},
+			"pgbouncer": {
+				Chart: "unmanaged",
+			},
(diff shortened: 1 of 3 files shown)


e2e/tests/shared/shared.go
```diff
@@ -100,6 +100,57 @@ func SetupTestWithDefaults(testName string) (*Test, error) {
...
+			AMI:                config.KopsAMI,
+			VPC:                config.VPC,
+			Provisioner:        config.Provisioner,
+			ArgocdClusterRegister: map[string]string{
+				"cluster-type": "customer",
+			},
+			DesiredUtilityVersions: map[string]*model.HelmUtilityVersion{
+				"nginx": {
+					Chart: "unmanaged",
+				},
```
gabrieljackson, line 116: We should use the constants for all these. Example:
```suggestion
				model.NginxCanonicalName: {
					Chart: model.UnmanagedUtilityVersion,
				},
```

e2e/tests/shared/shared.go
```diff
@@ -73,7 +73,7 @@ type Test struct {
 	Cleanup           bool
 }
 
-func SetupTestWithDefaults(testName string) (*Test, error) {
+func SetupTestWithDefaults(testName, utiliyManaged string) (*Test, error) {
```
gabrieljackson, line 76: I think passing in `utiliyManaged` as a bool makes a bit more sense than as a string.

```suggestion
func SetupTestWithDefaults(testName string, argoManagedUtilities bool) (*Test, error) {
```

e2e/tests/shared/shared.go
```diff
@@ -100,6 +100,57 @@ func SetupTestWithDefaults(testName string) (*Test, error) {
 		Provisioner:        config.Provisioner,
 	}
 
+	if utiliyManaged == "argocd" {
+		createClusterReq = &model.CreateClusterRequest{
+			AllowInstallations: true,
+			Annotations:        testAnnotations(testID),
+			AMI:                config.KopsAMI,
+			VPC:                config.VPC,
+			Provisioner:        config.Provisioner,
```
gabrieljackson, line 109: Instead of recreating the request we can just alter the existing one to change or add the values we want.

gabrieljackson (changes requested):
I think we can make a few improvements which I outlined below

nickmisasi (commented):
Nothing to add from Gabe's suggestions, just commenting so you can re-request review when you've addressed them!

gabrieljackson (approved):
LGTM. Thanks for addressing the feedback.

PR #1062, opened 2024-07-16
Log when webhook header values are empty

gabrieljackson:
This should help catch issues where env vars are not properly set.

Fixes https://mattermost.atlassian.net/browse/CLD-8065

```release-note
Log when webhook header values are empty
```


nickmisasi (approved):

fmartingr (approved):
👍

PR #1070, opened 2024-08-26
Handle missing VPC ID on initial provision

gabrieljackson:
Kops clusters don't have their VPC ID stored in metadata the first time the cluster is provisioned. This change adds a direct lookup in those cases.

Fixes https://mattermost.atlassian.net/browse/CLD-8293

```release-note
Handle missing VPC ID on initial provision
```


andrleite (approved):

stafot (approved):
LGTM

PR #1080, opened 2024-10-01
fix aws-cli path permission

andrleite:
#### Summary
This will fix the aws-cli path permission.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/CLD-8390

#### Release Note

```release-note
None
```

### DIFF
--- a/build/bin/user_setup
+++ b/build/bin/user_setup
@@ -4,6 +4,8 @@ set -euo pipefail
 # ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
 mkdir -p ${HOME}
 chown ${USER_UID}:0 ${HOME}
+chown ${USER_UID}:0 /mattermost-cloud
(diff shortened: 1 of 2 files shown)


build/bin/user_setup
```diff
@@ -4,6 +4,10 @@ set -euo pipefail
 # ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
 mkdir -p ${HOME}
 chown ${USER_UID}:0 ${HOME}
+chown ${USER_UID}:0 /mattermost-cloud
+# ensure /.aws exists and is accessible by group USER_UID
+mkdir /.aws
```
nickmisasi, line 9: You should use the `-p` flag, otherwise if the directory already exists this will throw an error and the script will fail because of `set -e` line at the top

nickmisasi (approved):

PR #1083, opened 2024-10-16
'Upgrade' version of k8s.io/client-go to proper format

nickmisasi:
#### Summary
In https://github.com/mattermost/mattermost-cloud/pull/1017 the version of this package was changed from `v0.27.2` to `v1.5.2` while on the surface this appears to be an upgrade, it was actually a downgrade. The package itself seems to have changed up it's versioning scheme with the introduction of Go modules in go1.16 - see the README [here](https://github.com/kubernetes/client-go). This PR returns the package version to the proper way, and upgrades it to the latest minor version. This enables other projects to use go get against the Provisioner.
#### Ticket Link
N/a

#### Release Note

```release-note
None
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	k8s.io/api v0.30.3
 	k8s.io/apiextensions-apiserver v0.30.3
 	k8s.io/apimachinery v0.30.3
-	k8s.io/client-go v1.5.2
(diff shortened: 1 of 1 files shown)


andrleite (approved):

gabrieljackson (approved):
Great find!

PR #1086, opened 2024-12-02
Add support for --context flag

nickmisasi:
#### Summary
Adds support for a global --context flag, allowing you to override the context used from your ~/.cloud/config.json file for a single run. This value can also be set via the `CP_CONTEXT` env variable if that workflow suits you better. The value in currentcontext of your ~/.cloud/config.json folder is used if either of the above are not specified. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-8611

#### Release Note

```release-note
None
```


gabrieljackson (approved):
Awesome! Thanks for adding this.

PR #1089, opened 2024-12-10
Add option to always schedule on external clusters

gabrieljackson:
External clusters may have scaling behavior that will accommodate any number of new installations. This change allows for ignoring cluster resource thresholds on external clusters so that installations are always scheduled on this cluster type.

Fixes https://mattermost.atlassian.net/browse/CLD-8686

```release-note
Add option to always schedule on external clusters
```


nickmisasi (approved):
LGTM

PR #1091, opened 2024-12-10
[CLD-8628 #2] Add custom status code for installation name/dns conflicts + bonus

nickmisasi:
#### Summary
This PR adds the ability for the provisioner to return a custom `409` (StatusConflict) status code when the requested installation name, or DNS is already in use on creation. This will allow us to be more communicative to users of the Cloud plugin when they rapidly attempt to delete and recreate an installation within a few minutes of each other, after we turn on the 24h deletion pending window. 

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-8628

#### Release Note

```release-note
None
```

### DIFF
--- a/internal/api/installation.go
+++ b/internal/api/installation.go
@@ -275,7 +275,7 @@ func handleCreateInstallation(c *Context, w http.ResponseWriter, r *http.Request
 
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusAccepted)
-	outputJSON(c, w, installation.ToDTO(annotations, dnsRecords))
+	outputJSON(c, w, installation.ToDTO(annotations, dnsRecords, nil))
 }
 
 func selectGroupForAnnotation(c *Context, annotations []string) (string, error) {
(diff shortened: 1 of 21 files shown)


internal/api/installation.go
```diff
@@ -661,7 +669,15 @@ func handleDeleteInstallation(c *Context, w http.ResponseWriter, r *http.Request
...
-	err = c.EventProducer.ProduceInstallationStateChangeEvent(installationDTO.Installation, oldState)
+	// Try to parse the user ID from the request context, so it can be passed along with the webhook event
+	actorID := ""
+	if value := r.Context().Value(ContextKeyUserID{}); value != nil {
+		if str, ok := value.(string); ok && str != "" {
+			actorID = str
+		}
+	}
+
+	err = c.EventProducer.ProduceInstallationStateChangeEvent(installationDTO.Installation, oldState, events.DataField{Key: "actor_id", Value: actorID})
```
nickmisasi, line 680: I snuck this in while I was here. This passes the ID of the actor that initiated the deletion request, so that webhook listeners can easily detect who did the deletion. The main use case for this will be the Cloud Plugin. Currently, whenever an installation is deleted for any reason, a message "Cloud isntallation <> has been manually deleted and is now removed from the cloud plugin", which confuses users, as sometimes an automatic cleanup script deletes them. With this change, the Cloud Plugin will be able to tell if it itself made the deletion request, or if it was another system, and adjust the message sent accordingly.

stafot (approved):

PR #1092, opened 2024-12-17
Add custom installation volume support

gabrieljackson:
Kubernetes volumes can now be added to installation deployments. Volumes backed by secrets can be created, updated, and deleted after an installation has been created. The volume secret data is first stored in AWS Secrets Manager and then appended as a k8s secret in an installation update. Volumes are mounted in the specified mount path of the container which should support many cases where sensitive information is required at the filesystem level of Mattermost.

Fixes https://mattermost.atlassian.net/browse/CLD-8711

```release-note
Add custom installation volume support
```

### DIFF
--- a/cmd/cloud/installation_flag.go
+++ b/cmd/cloud/installation_flag.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost-cloud/model"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
@@ -79,7 +80,7 @@ func (flags *installationCreateFlags) addFlags(command *cobra.Command) {
 
 type installationPatchRequestChanges struct {
 	ownerIDChanged          bool
-	versionCHanged          bool
+	versionChanged          bool
 	imageChanged            bool
 	sizeChanged             bool
 	licenseChanged          bool
@@ -89,7 +90,7 @@ type installationPatchRequestChanges struct {
 
 func (flags *installationPatchRequestChanges) addFlags(command *cobra.Command) {
 	flags.ownerIDChanged = command.Flags().Changed("owner")
-	flags.versionCHanged = command.Flags().Changed("version")
+	flags.versionChanged = command.Flags().Changed("version")
 	flags.imageChanged = command.Flags().Changed("image")
 	flags.sizeChanged = command.Flags().Changed("size")
 	flags.licenseChanged = command.Flags().Changed("license")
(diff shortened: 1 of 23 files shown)


internal/api/installation.go
```diff
@@ -390,6 +395,216 @@ func handleUpdateInstallation(c *Context, w http.ResponseWriter, r *http.Request
...
+	installationDTO, status, unlockOnce := getInstallationForTransition(c, installationID, newState)
+	if status != 0 {
+		w.WriteHeader(status)
+		return
+	}
+	defer unlockOnce()
+
+	oldState := installationDTO.State
+
+	err = createVolumeRequest.Apply(installationDTO.Installation)
```
nickmisasi, line 430: Do we need to validate anything with this? Even if as simple as checking for conflicts with current volume names?

nickmisasi (approved):
One small callout but overall LGTM. Nicely done!

internal/api/installation.go
gabrieljackson: Yep, we definitely want to do one final validation to look for issues. If you trace out the `Apply` you will end up here were we run some final checks: https://github.com/mattermost/mattermost-cloud/pull/1092/files#diff-888e481ee6df2e28a4dbed85a127b1d24d0a6d7cbefbdda42792d926dc1ec70bR63

cmd/cloud/installation_flag.go
```diff
@@ -209,6 +210,126 @@ func (flags *installationDeleteFlags) addFlags(command *cobra.Command) {
...
+	baseVolumeFlags
+	installationID string
+	volumeName     string
+}
+
+func (flags *installationCreateVolumeFlags) addFlags(command *cobra.Command) {
+	command.Flags().StringVar(&flags.installationID, "installation", "", "The id of the installation to create a volume for.")
+	command.Flags().StringVar(&flags.volumeName, "volume-name", "", "The name of the volume to create.")
+	command.Flags().StringVar(&flags.mountPath, "mount-path", "", "The container path to mount the volume in.")
+	command.Flags().BoolVar(&flags.readOnly, "read-only", true, "Whether the volume should be read only or not.")
```
andrleite, line 230: Do we need a read-only flag as a mount of type Secret is always read-only https://kubernetes.io/docs/concepts/storage/volumes/#secret

andrleite (approved):

cmd/cloud/installation_flag.go
gabrieljackson: Good catch, but this is just prepping for the future when we may have other volume types.

PR #1101, opened 2025-03-20
Add labels to cluster installation for external-dns ingress handling

stafot:
#### Summary
CLD-9019: Add labels to cluster installation for external-dns ingress handling

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-9019

#### Release Note

```release-note
Add labels to cluster installation for external-dns ingress handling
```

### DIFF
--- a/internal/provisioner/cluster_installation_provisioner.go
+++ b/internal/provisioner/cluster_installation_provisioner.go
@@ -51,7 +51,7 @@ func (provisioner Provisioner) CreateClusterInstallation(cluster *model.Cluster,
 		return errors.Wrap(err, "failed to get kube config path")
 	}
 
-	return provisioner.createClusterInstallation(clusterInstallation, installation, installationDNS, configLocation, logger)
+	return provisioner.createClusterInstallation(clusterInstallation, installation, installationDNS, configLocation, cluster, logger)
 }
 
 func (provisioner Provisioner) EnsureCRMigrated(cluster *model.Cluster, clusterInstallation *model.ClusterInstallation) (bool, error) {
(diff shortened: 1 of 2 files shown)


toninis (approved):

internal/provisioner/cluster_installation_provisioner.go
```diff
@@ -1081,6 +1081,7 @@ func clusterInstallationBaseLabels(installation *model.Installation, clusterInst
 	return map[string]string{
 		"installation-id":         installation.ID,
 		"cluster-installation-id": clusterInstallation.ID,
+		"dns":                     strings.TrimSuffix(clusterInstallation.ClusterID, "-public"),
```
gabrieljackson, line 1084: ClusterIDs should only ever be our standard 26 char IDs. Is there a reason we expect this suffix to exist?

stafot: @gabrieljackson Added this suffix mostly for sanity reasons to know that these ingreses will be handled by our public external DNS we will have a different one for the private DNSes we are using but for this we don't need any provisioner code changes as it affects only argo utilities

gabrieljackson (approved):

internal/provisioner/cluster_installation_provisioner.go
```diff
@@ -1077,21 +1078,27 @@ func generateClusterInstallationResourceLabels(installation *model.Installation,
...
-	return map[string]string{
+func clusterInstallationBaseLabels(installation *model.Installation, clusterInstallation *model.ClusterInstallation, cluster *model.Cluster) map[string]string {
+	labels := map[string]string{
 		"installation-id":         installation.ID,
 		"cluster-installation-id": clusterInstallation.ID,
 	}
+
+	if cluster != nil && cluster.Name != "" {
+		labels["dns"] = cluster.Name + "-public"
+	}
```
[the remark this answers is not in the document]
stafot: @gabrieljackson the problem that leads me to adopt this solution is that argoCD doesn't have the provisioner cluster id but has provisioner cluster name as info that can use in labelling @andrleite can provide more context if needed to why I was forced to this solution which was not preferred by me too

gabrieljackson (approved):

PR #1110, opened 2025-05-12
Add installation deletion scheduling

gabrieljackson:
This change adds optional installation deletion scheduling logic to the provisioner. Installations can now be created with a configurable TTL which can be useful for quick testing and other use-cases.

Installations with scheduled deletion will only be deleted if they are not deletion-locked and the current time is past their scheduled deletion time. Deletion-locked installations will be deleted once their lock is removed if the scheduled deletion time has passed.

This is completely separate from the existing deleting pending logic which allows for specifying how long an installation remains in a deletion-pending state after it was scheduled for deletion.

Fixes https://mattermost.atlassian.net/browse/CLD-9188

```release-note
Add installation deletion scheduling
```

### DIFF
--- a/cmd/cloud/installation.go
+++ b/cmd/cloud/installation.go
@@ -37,6 +37,7 @@ func newCmdInstallation() *cobra.Command {
 	cmd.AddCommand(newCmdInstallationVolumeDelete())
 	cmd.AddCommand(newCmdInstallationUpdateDeletion())
 	cmd.AddCommand(newCmdInstallationCancelDeletion())
+	cmd.AddCommand(newCmdInstallationScheduleDeletion())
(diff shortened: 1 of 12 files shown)


nickmisasi (approved):
LGTM! We'll probably want to discuss whether we should support skipping the deletion pending time for these short lived instances to save on cost

stafot (approved):
LGTM. Could you share some reasoning around this functionality? I mean was there a request for this? Obviously looks very useful to me.

gabrieljackson:
> LGTM. Could you share some reasoning around this functionality? I mean was there a request for this? Obviously looks very useful to me.

Yep, we have a few existing use-cases, but we will also be introducing 1-hour trials soon and this will directly support that experience.

PR #1115, opened 2025-06-12
Add probe overrides to installation level

nickmisasi:
#### Summary
Extends https://github.com/mattermost/mattermost-cloud/pull/1112 to allow for Pod Probe Overrides at the installation level, rather than only working server-wide. This is primarily in an effort to support the new Cloud Preview environments (which take longer to spin up because of database load) but could be useful elsewhere.

#### Release Note

```release-note
Support for installation level overrides of PodProbeOverrides
```

### DIFF
--- a/model/installation.go
+++ b/model/installation.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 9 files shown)


model/installation.go
```diff
@@ -71,6 +73,12 @@ type Installation struct {
 	configMergeGroupSequence int64
 }
 
+// PodProbeOverrides contains configuration for overriding pod probe settings.
+type PodProbeOverrides struct {
```
nickmisasi, line 77: Initially attempted to use the provisioner declaration of this type, but bringing that to the model package caused a circular dependency issue. Open to suggestions on centralizing, but 0/5 on this small duplication

gabrieljackson (approved):
LGTM, I think you just need to retry the build workflow.

andrleite (approved):

gabrieljackson (approved):

PR #1119, opened 2025-07-03
Update password creation to be scram-sha-256 instead of md5

angeloskyratzakos:
#### Summary
Update password creation to be scram-sha-256 instead of md5

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-9325
https://mattermost.atlassian.net/browse/CLD-9335

#### Release Note

```release-note
Update password creation to be scram-sha-256 instead of md5

```


internal/tools/aws/database_multitenant_pgbouncer_test.go
```diff
@@ -0,0 +1,329 @@
...
+
+		assert.Contains(t, expectedQuery, "INSERT INTO pgbouncer.pgbouncer_users")
+		assert.Contains(t, expectedQuery, username)
+		assert.Contains(t, expectedQuery, scramHash)
+		assert.Contains(t, expectedQuery, "usename")
+		assert.Contains(t, expectedQuery, "passwd")
+
+		// Ensure proper SQL formatting
+		assert.True(t, strings.Count(expectedQuery, "'") >= 4) // At least 4 single quotes for string values
+	})
```
gabrieljackson, line 127: Quite a few of these tests really aren't covering anything in the codebase. This one for example just creates a string and then checks if certain substrings exist in it. This could be changed into something more useful if it imported constants or functions from the code, but as is won't really help prevent bugs or regressions. There are others like it below.

docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: "3.7"
```
gabrieljackson, line 1: Is it a good idea to unpin the version?

gabrieljackson (approved):
Looks solid! I will approve, but calling out two things below. They aren't blockers, but if you feel like addressing them feel free to retag me for review.

internal/tools/aws/database_multitenant_pgbouncer_test.go
angeloskyratzakos: will remove them in another PR to avoid building the image again

PR #1152, opened 2026-03-10
fix: count all deletion states in GetGroupStatus pending deletion bucket

andrleite:
GetGroupStatus only matched the exact "deletion-pending" state when computing InstallationsPendingDeletion. Installations in other deletion sub-states (deletion-pending-requested, deletion-pending-in-progress, deletion-failed, etc.) fell into the InstallationsUpdating catch-all bucket, causing elrond's waitForGroupRelease to block indefinitely on installations that will never complete an update because they are being deleted.

Use AllInstallationDeletionStates instead of the single InstallationStateDeletionPending to properly categorize all deletion-related installations.

Made-with: Cursor

More details in the Jira ticket.

#### Ticket Link
https://mattermost.atlassian.net/browse/SEC-9899

#### Release Note

```release-note
Fix GetGroupStatus to count all deletion states as pending deletion instead of only "deletion-pending"
```


rahb3rt (approved):

nickmisasi (approved):

christopherfickess (approved):

fmartingr (approved):
👏

PR #1157, opened 2026-05-28
fix: tighten GitLab host matching and request handling in helm values fetch

edgarbellot:
## Summary

Small cleanup around `fetchFromGitlabIfNecessary` so it lines up with the way `--utilities-git-url` is meant to be used:

- Match the configured GitLab host by exact hostname (case-insensitive) instead of a `"git"` prefix.
- Require `https://` when reaching out to the configured host.
- Send the token via the standard `PRIVATE-TOKEN` header instead of appending it to the query string.
- Use a dedicated `http.Client` with redirects disabled and a request timeout, so a misbehaving upstream cannot stall the supervisor goroutine.
- Reject any non-2xx response explicitly.

Adds a unit test file for the function (none existed before), covering the host match, scheme requirement, header propagation, redirect behavior, and the no-token-configured path.

JIRA ticket: https://mattermost.atlassian.net/browse/MM-69024

Thank you to Samy Ghannad for letting us know about this issue.

```release-note
NONE
```

### DIFF
--- a/internal/provisioner/utility/helm_utils.go
+++ b/internal/provisioner/utility/helm_utils.go
@@ -13,12 +13,23 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/mattermost/mattermost-cloud/internal/tools/helm"
 	"github.com/mattermost/mattermost-cloud/model"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
+
+// gitlabFetchClient is the HTTP client used to download Helm values files
+// from the configured GitLab instance. Redirects are not followed and a
+// timeout is applied to keep the supervisor goroutine responsive.
+var gitlabFetchClient = &http.Client{
+	Timeout: 30 * time.Second,
+	CheckRedirect: func(req *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
+	},
+}
 
 const (
 	defaultKubeConfigPath            = ""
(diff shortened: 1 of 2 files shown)


internal/provisioner/utility/helm_utils.go
angeloskyratzakos, line 1: **Temp file not cleaned up on decode/write failure** (`internal/provisioner/utility/helm_utils.go:338`)
If `base64.StdEncoding.DecodeString` (line 351) or `os.WriteFile` (line 356) fails, the function returns an error but the temporary file created at line 346 is left on disk with no cleanup function returned. Over time, repeated failures could accumulate orphan temp files.

Suggested fix — remove the temp file on error paths:
```go
temporaryValuesFile, err := os.CreateTemp(os.TempDir(), "helm-values-file-")
if err != nil {
    return "", nil, errors.Wrap(err, "failed to create temporary file for Helm values file")
}
tmpPath := temporaryValuesFile.Name()
temporaryValuesFile.Close()

content, err := base64.StdEncoding.DecodeString(valuesFileBytes.Content)
if err != nil {
    os.Remove(tmpPath)
    return "", nil, errors.Wrap(err, "failed to decode base64-encoded YAML file")
}

err = os.WriteFile(tmpPath, content, 0600)
if err != nil {
    os.Remove(tmpPath)
    return "", nil, errors.Wrap(err, "failed to write values file to disk for Helm to read")
}
```

This also addresses the file descriptor leak noted by CodeRabbit by closing the handle immediately.

internal/provisioner/utility/helm_utils.go
```diff
@@ -283,18 +309,26 @@ func fetchFromGitlabIfNecessary(path string) (string, func(string), error) {
...
-	resp, err := http.Get(path)
+	resp, err := gitlabFetchClient.Do(req)
 	if err != nil {
 		return "", nil, errors.Wrap(err, "failed to request the values file from Gitlab")
 	}
-	if resp.StatusCode >= 400 {
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return "", nil, errors.Errorf("request to Gitlab failed with status: %s", resp.Status)
 	}
```
angeloskyratzakos, line 338: **Unbounded io.ReadAll on HTTP response body** (`internal/provisioner/utility/helm_utils.go:335`)
The `io.ReadAll(resp.Body)` call reads the entire response into memory without any size limit. If the upstream GitLab instance (or a compromised endpoint) returns an unexpectedly large response, this could exhaust memory and crash the supervisor goroutine.

Suggested fix — wrap the body with `io.LimitReader`:
```go
const maxGitlabResponseSize = 10 << 20 // 10 MiB
body, err := io.ReadAll(io.LimitReader(resp.Body, maxGitlabResponseSize))
```

This is behind HTTPS + host validation + auth, so exploitation requires a compromised GitLab, but a defensive limit is cheap insurance.

edgarbellot:
@angeloskyratzakos Thanks for the review! Addressed both points in 0902732. Please take a look when you have a moment.

andrleite (approved):

PR #1158, opened 2026-07-14
SEC-9253: opt-in cross-installation node separation

angeloskyratzakos:
#### Summary

Community (`mm-rxoc`) and hub (`mm-c4ja`) Mattermost Cloud installations share the internal cluster and can currently be scheduled onto the **same node**. This adds an **opt-in** mechanism so annotated installations avoid co-locating with any other opted-in installation.

The provisioner owns and **overwrites** the CR's `Spec.Scheduling.Affinity` and `Spec.ResourceLabels` on every reconcile, so the durable fix lives in `internal/provisioner/cluster_installation_provisioner.go`:

- A new Cloud annotation `separate-nodes` opts an installation in.
- When opted in, `clusterInstallationStableLabels` stamps a `node-separation-group=separate-nodes` pod label (reaches pods via `Spec.ResourceLabels`).
- `generateAffinityConfig` appends a **soft** (`PreferredDuringSchedulingIgnoredDuringExecution`, weight 100) `podAntiAffinity` term selecting that label, with `NamespaceSelector: {}` (cross-namespace — peers live in their own namespaces) and `topologyKey: kubernetes.io/hostname`.
- `model.Installation` doesn't carry annotations (separate join), so `resolveNodeSeparation` fetches them via `store.GetAnnotationsForInstallation` at the create/update reconcile call sites.

Separation is **soft on purpose** (captain's decision): it never blocks scheduling — under node pressure the scheduler may still co-locate rather than leave a pod `Pending`. Flipping to a hard guarantee is a one-line change (move the term to `RequiredDuringSchedulingIgnoredDuringExecution`) once guaranteed spare-node capacity is confirmed.

**Blast radius:** installations **without** the `separate-nodes` annotation get byte-identical generated affinity and labels to before (covered by tests). All changes are in mattermost-cloud; no operator changes.

Rollout after merge: add the `separate-nodes` annotation to `community` and `hub`, then reconcile.

#### Ticket Link

https://mattermost.atlassian.net/browse/SEC-9253

#### Release Note

```release-note
NONE
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,22 @@
+# Project agent memory
+
+This file is the project's committed home for project-intrinsic agent knowledge: build, test, release, architecture, and sharp-edge notes that should travel with the code.
+
+- Add durable project-specific notes here as they are discovered through real work.
+
+## Pod scheduling / affinity
+
+- The provisioner OWNS the Mattermost CR's scheduling: `ensureScheduling` in
+  `internal/provisioner/cluster_installation_provisioner.go` overwrites
+  `mattermost.Spec.Scheduling.Affinity` (via `generateAffinityConfig`) and
+  `mattermost.Spec.ResourceLabels` (via `clusterInstallationStableLabels`) on EVERY
+  create/update reconcile. Hand-patching the CR or the Deployment is reverted on the next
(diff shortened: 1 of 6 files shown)


angeloskyratzakos:
@coderabbitai resolve

PR #1160, opened 2026-09-03
Align operator v2

andrleite:
#### Summary

Align the provisioner with `mattermost-operator` v2.0.0:

- Replace all `v1alpha1` imports with `v1beta1` across model, k8s, provisioner, supervisor, and tooling packages
- Drop the `ClusterInstallation` Kubernetes client (`MattermostClientsetV1Alpha`, `createOrUpdateClusterInstallation`, scheme registration) — the CRD was removed in operator v2
- Update `Size` types: `ClusterInstallationSize` is now `v1beta1.Size` (app replicas/resources only — no Minio or Database components)
- Remove `includeDatabase`/`includeMinio` bool args from `CalculateCPUMilliRequirement` and `CalculateMemoryMilliRequirement`
- Rename `GetClusterSize` → `GetMattermostSize` following the operator rename
- Bump Docker build image default to `golang:1.26.6` to match `go.mod`
- Suppress new QF/ST staticcheck categories introduced by golangci-lint v2 to avoid unrelated churn in pre-existing code

#### Release Note

```release-note
NONE
```


internal/provisioner/cluster_installation_provisioner.go
```diff
@@ -1084,13 +1083,11 @@ func setMMInstanceSize(installation *model.Installation, mattermost *mmv1beta1.M
 
 // This function is adapted from Mattermost Operator, we can make it public
 // there to avoid copying.
-func overrideReplicasAndResourcesFromSize(size mmv1alpha1.ClusterInstallationSize, mm *mmv1beta1.Mattermost) {
+func overrideReplicasAndResourcesFromSize(size mmv1beta1.Size, mm *mmv1beta1.Mattermost) {
```
angeloskyratzakos, line 1086: Andre do you think that might anyone still uses these ones?

**FileStore and Database resource overrides removed from Mattermost CR**

The old `overrideReplicasAndResourcesFromSize` called `mm.Spec.FileStore.OverrideReplicasAndResourcesFromSize(size)` and `mm.Spec.Database.OverrideReplicasAndResourcesFromSize(size)`. The new version only sets App replicas and scheduling resources.

If any existing Mattermost CRs in production still have FileStore or Database specs that previously relied on these size-driven overrides, those components will now get the operator's defaults instead of the size-specific values. Verify that operator v2 handles this correctly for all in-flight installations.

cmd/tools/cloudburst/cloudburst.go
```diff
@@ -396,7 +396,7 @@ func (b *Blaster) createInstallation() (*cloud.Installation, error) {
 			Filestore:       b.filestore,
```
angeloskyratzakos, line 398: Minor one:
**cloudburst relies on SetDefaults to derive Name from DNSNames**

The `createInstallation` call provides `DNSNames` but not `Name`. This works because `SetDefaults()` now derives `Name` from `DNSNames[0]`. However, the load-test tool generates DNS names with random IDs (`b.testID + NewID()[:6]`), producing installation names like `testid-a3f2bc` which could collide in edge cases with the 6-char random suffix.

This is low-risk for a load-test tool, but worth noting that `Name` derivation is an implicit contract. Setting `Name` explicitly would make the intent clearer and avoid surprises if `SetDefaults` logic changes.

internal/provisioner/cluster_installation_provisioner.go
andrleite: Yes, confirmed — all production installations use external database and filestore (verified against the cluster). The FileStore/Database size overrides only applied to in-cluster operator-managed components, which aren't used. Safe to remove.

cmd/tools/cloudburst/cloudburst.go
andrleite: Fixed. The change generates installationName once (e.g. "mytestid-a3f2bc") and uses it for both Name and the DNS prefix, so validation passes, and the intent is explicit — no implicit SetDefaults reliance.

.golangci.yml
```diff
@@ -1,31 +1,40 @@
...
     - unused
+  exclusions:
+    paths:
+      - internal/mocks
+  settings:
+    govet:
+      enable-all: true
+      disable:
+        - fieldalignment
+    staticcheck:
```
fmartingr, line 29: Any particular reason for ignoring this?

Makefile
```diff
@@ -69,7 +69,7 @@ GOIMPORTS_VER := v0.31.0
 GOIMPORTS_BIN := goimports
 GOIMPORTS := $(TOOLS_BIN_DIR)/$(GOIMPORTS_BIN)
 
-GOLANGCILINT_VER := v1.64.8
+GOLANGCILINT_VER := v2.13.1
```
fmartingr: Mattermost core uses v2.12.2, unsure if we want to be in sync with that (defaults usually change with releases)

fmartingr (approved):

nickmisasi (approved):

PR #1161, opened 2026-09-10
feat: add HTTPRoute (Gateway API) support to installation provisioner

andrleite:
#### Summary

Adds support for [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) HTTPRoute as an alternative to the traditional Ingress resource for routing traffic to Mattermost installations.

**New fields on the Installation model:**
- `IngressType` — selects the routing resource: `"ingress"` (default, no behavior change for existing installations) or `"httproute"`
- `GatewayConfig` — required when `IngressType` is `"httproute"`, holds the Gateway name, namespace, and optional section name

Both fields are exposed on `CreateInstallationRequest` and `PatchInstallationRequest`, persisted to the store, and covered by a DB schema migration (`0.54 → 0.55`).

**Provisioner behavior:**
- HTTPRoute installations set `Spec.HTTPRoute` and disable `Spec.Ingress` on the `Mattermost` CR
- Ingress installations (the default) clear `Spec.HTTPRoute` — same behavior as today
- Wired through create, update, and hibernation flows

Depends on [mattermost/mattermost-operator#473](https://github.com/mattermost/mattermost-operator/pull/473) for the `HTTPRouteSpec` and `GatewayReference` types on the operator side. The operator dependency is pinned to `b03fd868` which includes those types.

#### Release Note

```release-note
Add HTTPRoute (Gateway API) support for installation traffic routing. Installations can now be created or updated with --ingress-type httproute and a gateway-config to use a Gateway API HTTPRoute resource instead of a Kubernetes Ingress. Existing installations default to ingress
```


nickmisasi (commented):
Will this become the new default in the Provisioner? We'll either need to maintain or disable/remove the IP Filtering feature, because it's available in enterprise, and won't work

andrleite:
> Will this become the new default in the Provisioner? We'll either need to maintain or disable/remove the IP Filtering feature, because it's available in enterprise, and won't work

HTTPRoute is not the new default — it's opt-in and "ingress" remains the default. Existing installations are completely unaffected.

Your concern is valid for the longer term. The options would be:

- Gateway API equivalent — implement IP filtering via a gateway-specific mechanism (e.g. NetworkPolicy, or gateway extensions like Envoy's RateLimitPolicy/SecurityPolicy). This is a non-trivial effort and varies by gateway implementation.
- Deprecate AllowedIPRanges — if the feature has low adoption, remove it before making HTTPRoute the default.
For this PR, I've added a hard error in the provisioner if AllowedIPRanges + HTTPRoute somehow coexist, so enterprise customers are protected. A future migration plan would need to address this explicitly before changing the default.

The idea is to start migrating the test env to the HTTPRoute; at some point it'll become default. Before flipping the default, we'll need to resolve AllowedIPRanges — either through a Gateway API equivalent or by deprecating it. We'll track that as a prerequisite for the default switch.
@DSchalla , @angeloskyratzakos Thoughts?

fmartingr (approved):

andrleite:
> > Will this become the new default in the Provisioner? We'll either need to maintain or disable/remove the IP Filtering feature, because it's available in enterprise, and won't work
> 
> HTTPRoute is not the new default — it's opt-in and "ingress" remains the default. Existing installations are completely unaffected.
> 
> Your concern is valid for the longer term. The options would be:
> 
> * Gateway API equivalent — implement IP filtering via a gateway-specific mechanism (e.g. NetworkPolicy, or gateway extensions like Envoy's RateLimitPolicy/SecurityPolicy). This is a non-trivial effort and varies by gateway implementation.
> * Deprecate AllowedIPRanges — if the feature has low adoption, remove it before making HTTPRoute the default.
>   For this PR, I've added a hard error in the provisioner if AllowedIPRanges + HTTPRoute somehow coexist, so enterprise customers are protected. A future migration plan would need to address this explicitly before changing the default.
> 
> The idea is to start migrating the test env to the HTTPRoute; at some point it'll become default. Before flipping the default, we'll need to resolve AllowedIPRanges — either through a Gateway API equivalent or by deprecating it. We'll track that as a prerequisite for the default switch. @DSchalla , @angeloskyratzakos Thoughts?

@nickmisasi I checked production, and we don't have any customers using this feature even so I'll backport this to HTTPRoute before going to production.

